PR #15, opened 2016-05-02
Append number to target device

johnavp1989:
This successfully mounts /dev/sda0 when a single disk is defined but fails when multiple disks are defined.

> libvirt_domain.domain: diffs didn't match during apply. This is a bug with Terraform and should be reported as a GitHub Issue.

> Mismatch reason: attribute mismatch: disk.1.volume_id

I'm stuck on this issue and would appreciate any help in resolving this.

### DIFF
--- a/libvirt/disk_def.go
+++ b/libvirt/disk_def.go
@@ -2,6 +2,7 @@ package libvirt
 
 import (
 	"encoding/xml"
+        "strconv"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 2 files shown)


dmacvicar:
Thanks a lot for your patch, it was much easier to understand the problem thanks to it!. This time I am using a different solution as we need to generate disk names, not partition names and I prefer to setup the name outside of the disk definition structs.

johnavp1989:
Glad it was helpful. I cheated and used partition names because I didn't know how to create and manipulate arrays. You solution is much better. Thanks for addressing this so quickly!

PR #27, opened 2016-06-04
Vendor all the dependencies

flavio:
Use [govend](https://github.com/govend/govend) to vendor all the dependencies of this project. This will make the whole build process more reliable and reproducible.

What do you think about that? I think I should also change the README of the project to mention we need go 1.5+


dmacvicar:
After reading a bit about `govend` and `Godeps` I am unsure I want to go this direction.

You can see that the travis for this PR still fails, because something is missing, so it is easy for the vendor meta-data and the checked in sources to go out of sync.

I don't think the sources belong in git, only the metadata that specifies which versions we depend on. Just like a Gemfile/gemspec.

I would be willing to checkin a `vendor.yml` or a `Godeps/Godeps.json` into git with the metadata about the "frozen" versions we depend. But the actual process of incorporating the sources into the tarball should be done when updating the tarball in OBS (a simple script that unpacks, executes godep and packs everything) and in `travis.yml` before building.

Having the metadata *and* the sources in git feels redundant.

flavio:
I update the PR a lot since the last time you checked it out, as you can see now all the tests are passing.

This what I changed:
  * The `vendor.yml` file has to include **all** the project dependencies to ensure builds are consistent and reproducible
  * Travis configuration file has been changed to
    * Use latest stable release of Go (1.6, which takes advantage of vendored code out of the box)
    * Do not pull any dependency, they are already there inside of the `vendor` directory
    * Do not test the code twice: the old code ran the tests inside of the `install` section and then Travis executed them another time as part of the way "default" golang project support are handled.
    * Be in charge of how our tests are run: the old code (and travis' default one) executed all the tests inside of the git checkout, that triggered the execution of the tests of our vendored code.

I know that vendoring the code in that way is not nice however:

  * That makes things easier at packaging time: there's no need to checkout the source code of the project and then, with a separate step, download all the vendored code. This has to be done on the packager laptop because OBS has no network connection to perform the 2nd step.
  * Checking out the vendored code takes time and sometimes might not work at all. A repository might be offline or be even removed (nodejs anyone?). Having everything bundled with the main repo ensures building is always going to work.
  * This setup will make it easier to spot usage of unvendored code. The build will fail on Travis if someone starts requiring a new library which is not part of `vendor.yml`. Referencing all the used libraries is really important to ensure a project will keep working over the time.

If you really don't want to add all the dependencies to the project please at least take the new `vendor.yml` file and allow me to change how Travis will download the deps (via `govend` instead of `go get`).

moio:
@dmacvicar: I went over this again with @flavio and changed my mind.

When a (new) contributor wants to build this project she needs all dependencies and build dependencies, obviously.

That means it's either up to:
 1. you (via vendoring) or
 2. the community of developers of all dependencies and build dependencies
to make sure those are available for download. **Obviously option 2. is worse in terms of availability.**

In the Java/mvn ecosystem we can rely on the Central Repo being available almost 24/7 - thanks to Sonatype this has been a reality for years, so developers rarely vendor dependencies. In the Ruby world we have rubygems.org.

Here we don't have such choice, at least not yet, so between you and the whole world I would still trust you more for availability.

:+1:

flavio:
Ok, after having discussed with @dmacvicar we agreed to track all the dependencies using a `vendor.yml` file but to not add them to this git repo (at least for now).

I changed the PR to leave all the good travis changes.

flavio:
Travis is green, now it's rebuilding because I just reverted a change I made to the `.gitignore` file. You can safely merge it

PR #30, opened 2016-06-12
Allow to specify the base volume by using a name

flavio:
It's now possible to specify a base volume by using the
`base_volume_name`. This is an approach done also by other Terraform
providers: either use an "internal" ID or a human-friendly name.

This makes easier to share the same base volume between different
terraform projects.

### DIFF
--- a/docs/providers/libvirt/r/volume.html.markdown
+++ b/docs/providers/libvirt/r/volume.html.markdown
@@ -34,4 +34,5 @@ The following arguments are supported:
 * `size` - (Optional) The size of the volume in bytes (if you don't like this, help fix [this issue](https://github.com/hashicorp/terraform/issues/3287).
    If `source` is specified, `size` will be set to the source image file size.
 * `base_volume_id` - (Optional) The backing volume (CoW) to use for this volume.
-
+* `base_volume_name` - (Optional) The name of the backing volume (CoW) to use for this volume. Note well: when `base_volume_pool` is not specified the volume is going to be searched inside of `pool`.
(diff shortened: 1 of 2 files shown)


dmacvicar:
I think it is a bit more complicated than that.

If you look at the schema for volume, it already uses `name`. Why? Because there is `pool`. `(name, pool)` identifies a volume.

When you specify a base volume, we use `base_volume_id`, which maps to `key` in libvirt. `key` is unique at the libvirt connection, so we can ask libvirt for a volume without having to specify the pool (the pool is part of the key).
For example, for a normal qcow2 volume on a `dir://` storage pool `/var/lib/libvirt/images`, `name` is `sles12sp1_vagrant_box_image_3.2.0.img`, `key` is `/var/lib/libvirt/images/sles12sp1_vagrant_box_image_3.2.0.img`.

So, if you specify only `name`, you are assuming a pool. In this case you are assuming the pool of the base volume is the same as the volume definition.

You need either to document that assumption or to add `base_volume_pool` too.

flavio:
You are right, this is already documented here:

> +* `base_volume_name` - (Optional) The name of the backing volume (CoW) to use for this volume. Note well: this is going to be searched inside of the chosen `pool`.

Maybe it's not clear enough?

However I like the idea of using a `base_volume_pool`. I would go that way:
  * no `base_volume_pool` specified: use the same `pool` as the new volume
  * `base_volume_pool` is specified: use it to look for the base volume

What do you tihnk?

dmacvicar:
I think that way it would make more sense.

flavio:
I changed the code to reflect the recommended change

PR #31, opened 2016-06-12
Domain: allow to set title and description

flavio:
Allow to set title and description of a domain. This feature might seem
superfluous at first sight, but can be used to quickly integrate
terraform with other tools.

For example, this can be used to ship custom grains to salt-ssh via a
custom salt roster module.

If you are fine with this PR I'll extend the docs to reflect these changes.

### DIFF
--- a/libvirt/domain_def.go
+++ b/libvirt/domain_def.go
@@ -5,13 +5,15 @@ import (
 )
 
 type defDomain struct {
-	XMLName  xml.Name  `xml:"domain"`
-	Name     string    `xml:"name"`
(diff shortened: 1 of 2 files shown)


dmacvicar:
Why would you use the description field for integration with other tools instead of the `metadata` section of the domain? (https://libvirt.org/formatdomain.html#elementsMetadata )

flavio:
That's easy to answer: I didn't know about it :smiley_cat: 

I'll rework the PR to use this libvirt feature.

flavio:
I'm going to file a new PR

PR #32, opened 2016-06-15
Domain: allow to set custom metadata

flavio:
Allow to set custom metadata of a domain. This feature can be used to quickly
integrate terraform with other tools.

For example, this can be used to ship custom grains to salt-ssh via a
custom salt roster module.

This supersedes PR https://github.com/dmacvicar/terraform-provider-libvirt/pull/31

### DIFF
--- a/docs/providers/libvirt/r/domain.html.markdown
+++ b/docs/providers/libvirt/r/domain.html.markdown
@@ -31,6 +31,10 @@ The following arguments are supported:
 * `vcpu` - (Optional) The amount of virtual CPUs. If not specified, a single CPU will be created.
 * `disk` - (Optional) An array of one or more disks to attach to the domain. The `disk` object structure is documented below.
 * `network_interface` - (Optional) An array of one or more network interfaces to attach to the domain. The `network_interface` object structure is documented below.
+* `metadata` - (Optional) A string containing valid XML data. This is going to be
+  added to the final domain inside of the [metadata tag](https://libvirt.org/formatdomain.html#elementsMetadata).
+  This can be used to integrate terraform into other tools by inspecting the
+  the contents of the `terraform.tf` file.
 
 The `disk` block supports:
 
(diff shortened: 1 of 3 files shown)


flavio:
The PR has been changed to use the `metadata` attribute.

dmacvicar:
I like it more now. What do you think about the fact of allowing custom XML there in the terraform string? Could this be used to "corrupt" the final generated xml?

Ideally metadata should be an arbitrary map, which generates:

```xml
<metadata>
    <terraform-libvirt:foo xmlns:terraform-libvirt="http://github.com/dmacvicar/terraform-provider-libvirt/">..</terraform-libvirt:foo>
    <terraform-libvirt:bar xmlns:terraform-libvirt="http://github.com/dmacvicar/terraform-provider-libvirt/">..</terraform-libvirt:bar>
  </metadata>
```

However, arbitrary Map type is (AFAIL) not supported by terraform.

All other providers `user_data` attribute is a plain String. Why not generating something like

```xml
<metadata>
    <terraform-libvirt:user_data xmlns:terraform-libvirt="http://github.com/dmacvicar/terraform-provider-libvirt/">
     <![CDATA[ that string ]]>
    </terraform-libvirt:user_data>
  </metadata>
```

or even simpler:

```xml
<metadata>
     <![CDATA[ that string ]]>
  </metadata>
```

What do you think?

flavio:
I updated the code to reflect the suggested changes:
```xml
<metadata>
    <terraform-libvirt:user_data xmlns:terraform-libvirt="http://github.com/dmacvicar/terraform-provider-libvirt/">
     <![CDATA[ that string ]]>
    </terraform-libvirt:user_data>
  </metadata>
```

Note well: https://github.com/dmacvicar/libvirt-go/pull/3 has to be merged to build this PR.

georgecodes:
Whatever happened to this? I appear to be able to configure metadata in the Terraform code, but it never makes it to a domain, and any code which does so isn't in any releases that I can see.

dmacvicar:
@georgecodes I don't use the feature. Do you have a simple example I can try?

georgecodes:
metadata = "<user_data:valid>XML</user_data:valid>"

This on a libvirt resource *should* populate the metadata element of the guest xml definition.

PR #35, opened 2016-06-27
Build against released terraform 0.6.16 instead of master

moio:
Current build instructions are broken, when running `terraform` you get the following error:

> Incompatible API version with plugin. Plugin version: 1, Ours: 2

This happens because `go get`, by default, [will always use the `master` branch for dependencies](stackoverflow.com/questions/24855081/how-do-i-import-a-specific-version-of-a-package-using-go-get?answertab=votes#tab-top) and at the same time [terraform developers changed the plugin API to 2](https://groups.google.com/forum/#!topic/terraform-tool/a1agp4CAEDk) in `master`.

This patch informs the user to build against the released `terraform` version.

This patch should be reverted as soon as 0.7 is actually released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,13 @@ $ terraform destroy
 2.  [Set up Gopath](https://golang.org/doc/code.html)
 3.  `git clone` this repository into `$GOPATH/src/github.com/dmacvicar/terraform-provider-libvirt`
 4.  Run `go get` to get dependencies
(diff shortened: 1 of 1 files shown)


dmacvicar:
Yeah, I need to merge a govend file so that one could just vendor fixed dependency versions (without merging the actual dependencies into my git)

Issue #4, opened 2016-03-06
ssh key resources


dmacvicar:
The main idea is to allow ssh key resources, and then locally assemble the cloud init metadata, wrap it into an iso and create a metadata volume for the instance resource uploading this created image.

cloud-init can use this disk as a metadata/userdata source to configure the ssh key.

flavio:
I'm working on it

flavio:
I think this can now be closed.

PR #39, opened 2016-06-27
Cloudinit

flavio:
This is PR introduces support for cloudinit. This allows to set a custom hostname and a custom authorized key for the root user as requested [here](https://github.com/dmacvicar/terraform-provider-libvirt/issues/4).

## How it works

The user has to define a `libvirt_cloudinit` resource and associate it with a domain. Behind the scene the code will:
  * create the cloud-init related files
  * create a new ISO file containing these files
  * upload the ISO file to the desired pool
  * add a CDROM drive to the domain

## Requirements

The `genisoimage` command is required at runtime to generate the cloud-init ISO. I tried to understand how to use `libisofs` but I honestly gave up, it seemed an incredible waste of time...

I also had to include new external libraries. Once we are done with this PR I'll go back the the other one about govend, the situation is about to escape our control...

## Future work

Cloud-init is a can of worms, right now I decided to support only 2 types of attributes (local host and ssh keys). I was thinking to further expend the cloud-init support by allowing the user to enter custom values for the `user-data` and `meta-data` files via custom strings. What do you think about that?

### DIFF
--- a/docs/providers/libvirt/r/domain.html.markdown
+++ b/docs/providers/libvirt/r/domain.html.markdown
@@ -35,6 +35,10 @@ The following arguments are supported:
   added to the final domain inside of the [metadata tag](https://libvirt.org/formatdomain.html#elementsMetadata).
   This can be used to integrate terraform into other tools by inspecting the
   the contents of the `terraform.tf` file.
+* `cloudinit` - (Optional) The `libvirt_cloudinit` disk that has to be used by
+  the domain. This is going to be attached as a CDROM ISO. Changing the
+  cloud-init won't cause the domain to be recreated, however the change will
+  have effect on the next reboot.
 
 The `disk` block supports:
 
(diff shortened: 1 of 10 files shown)


flavio:
Note well: this PR requires these code changes:
https://github.com/dmacvicar/libvirt-go/pull/4

flavio:
Note well: this PR requires these code changes:
https://github.com/dmacvicar/libvirt-go/pull/4

docs/providers/libvirt/r/domain.html.markdown
```diff
@@ -31,6 +31,14 @@ The following arguments are supported:
 * `vcpu` - (Optional) The amount of virtual CPUs. If not specified, a single CPU will be created.
 * `disk` - (Optional) An array of one or more disks to attach to the domain. The `disk` object structure is documented below.
 * `network_interface` - (Optional) An array of one or more network interfaces to attach to the domain. The `network_interface` object structure is documented below.
+* `metadata` - (Optional) A string containing valid XML data. This is going to be
+  added to the final domain inside of the [metadata tag](https://libvirt.org/formatdomain.html#elementsMetadata).
```
dmacvicar: after your PR, does it really need to be valid xml data?

flavio: You are right, I think this PR is somehow screwed up, there are some changes inside of it that have already been marged by you. I'm going to rebase it.

dmacvicar:
too late!

PR #48, opened 2016-07-04
Fix simple networking

flavio:
Some of the new attributes have to be omitted when not being used,
otherwise the simple networking with the default libvirt network won't
work


flavio:
Travis was happy about this PR but then I accidentally broke it with a force... I don't think I have the rights to trigger another build

inercia:
LGTM

PR #49, opened 2016-07-04
Get address of lan interfaces

flavio:
When a network card is attached to a LAN the `domain.ListAllInterfaceAddresses` API cannot be used to get the IP address associated with the interfaces.

The only solution to this problem is to use the [Qemu guest agent](http://wiki.libvirt.org/page/Qemu_guest_agent) to fetch this information.

This PR will take advantage of the qemu guest ageint (when installed inside of the domain) to workaround the limitations of the `domain.ListAllInterfaceAddresses`.

In theory this PR could also allow to relax the required version of libvirt.

The PR includes already the fix shipped with https://github.com/dmacvicar/terraform-provider-libvirt/pull/48, I'll do a rebase once the code is merged into master.

### DIFF
--- a/libvirt/qemu_agent.go
+++ b/libvirt/qemu_agent.go
@@ -1,0 +1,90 @@
+package libvirt
+
+import (
+	"encoding/json"
+	"log"
+	"strings"
+
+	libvirt "github.com/dmacvicar/libvirt-go"
+)
+
+type QemuAgentInterfacesResponse struct {
+	Interfaces []QemuAgentInterface `json:"return"`
+}
+
+type QemuAgentInterface struct {
+	Name        string `json:"name"`
+	Hwaddr      string `json:"hardware-address"`
+	IpAddresses []struct {
+		Type    string `json:"ip-address-type"`
+		Address string `json:"ip-address"`
+		Prefix  uint   `json:"prefix"`
+	} `json:"ip-addresses"`
+}
+
+// Retrieve all the interfaces attached to a domain and their addresses. Only
+// the interfaces with at least an IP address are returned.
+// When wait4ipv4 is turned on the code will not report interfaces that don't
+// have a ipv4 address set. This is useful when a domain gets the ipv6 address
+// before the ipv4 one.
+func getDomainInterfacesViaQemuAgent(domain *libvirt.VirDomain, wait4ipv4 bool) []libvirt.VirDomainInterface {
+	log.Print("[DEBUG] get network interfaces using qemu agent")
+
+	var interfaces []libvirt.VirDomainInterface
+
+	result := domain.QemuAgentCommand(
+		"{\"execute\":\"guest-network-get-interfaces\"}",
+		libvirt.VIR_DOMAIN_QEMU_AGENT_COMMAND_DEFAULT,
+		0)
+
+	log.Printf("[DEBUG] qemu-agent response: %s", result)
+
+	response := QemuAgentInterfacesResponse{}
+	if err := json.Unmarshal([]byte(result), &response); err != nil {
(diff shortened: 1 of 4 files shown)


libvirt/qemu_agent.go
```diff
@@ -0,0 +1,90 @@
...
+		ipv4Assigned := false
+		for _, addr := range iface.IpAddresses {
+			if addr.Address == "" {
+				// ignore interfaces without an address (eg. waiting for dhcp lease)
+				continue
+			}
+
+			libVirtAddr := libvirt.VirDomainIPAddress{}
+			libVirtAddr.Addr = addr.Address
+			libVirtAddr.Prefix = addr.Prefix
```
inercia: what about this?

```language:Go
libVirtAddr := libvirt.VirDomainIPAddress{
    Addr: addr.Address,
    Prefix = addr.Prefix,
}
```

flavio: Good catch, I'm going to fix it.

libvirt/resource_libvirt_domain.go
```diff
@@ -748,3 +759,36 @@ func newDiskForCloudInit(virConn *libvirt.VirConnection, volumeKey string) (defD
...
+		case libvirt.VirError:
+			virErr := err.(libvirt.VirError)
+			if virErr.Code != libvirt.VIR_ERR_OPERATION_INVALID || virErr.Domain != libvirt.VIR_FROM_QEMU {
+				return interfaces, fmt.Errorf("Error retrieving interface addresses: %s", err)
+			}
+		}
+	}
+	log.Printf("[DEBUG] Interfaces %+v", interfaces)
+
+	return interfaces, nil
```
inercia: What would be the best alternative: to invoke `ListAllInterfaceAddresses`  as a fallback when the `qemu-agent` cannot obtain a list of IP addresses or the opposite?

libvirt/resource_libvirt_domain.go
```diff
@@ -748,3 +759,36 @@ func newDiskForCloudInit(virConn *libvirt.VirConnection, volumeKey string) (defD
...
+	// get all the interfaces using the qemu-agent, this includes also
+	// interfaces that are not attached to networks managed by libvirt
+	// (eg. bridges, macvtap,...)
+	interfaces := getDomainInterfacesViaQemuAgent(domain, true)
+	if len(interfaces) > 0 {
+		// the agent will always return all the interfaces, both the
+		// ones managed by libvirt and the ones attached to bridge interfaces
+		// or macvtap. Hence it has the highest priority
+		return interfaces, nil
+	}
```
inercia: Isn't `qemu-agent` authoritative about the IP addresses? I mean that, if `qemu-agent` is installed and we do not get IPs, maybe we could just return here...

flavio: Yes, `qemu-agent` is authoritative about the IP addresses. Unfortunately I didn't find a way to figure out if  `qemu-agent` is installed into the VM. No error is raised when there's no agent, an empty list of interfaces is returned. That's why the code has still to use the `ListAllInterfaceAddresses` API in certain cases.

libvirt/resource_libvirt_domain.go
flavio: `qemu-agent` is authoritative and returns **all** the network interfaces, even the ones not backed by a libvirt network. That's why the code first tries to get them with the agent, and then with the `ListAllInterfaceAddresses` API.

flavio:
@inercia thanks for the feedback, do you think we can merge the PR. This code is going to be needed by our CI server

inercia:
LGTM

PR #50, opened 2016-07-05
Support UEFI images by adding firmware/nvram options for domains

inercia:
This PR adds support for UEFI images by adding firmware/nvram options for domains.
Minor fixes


dmacvicar:
Can you explain what is the use case for supporting this?

inercia:
> Can you explain what is the use case for supporting this?

Some images need a [UEFI](https://www.wikiwand.com/en/Unified_Extensible_Firmware_Interface)-capable domain in order to be booted from the [EFI partition](https://www.wikiwand.com/en/EFI_system_partition). This seems to be the preferred booting mechanism for KIWI-generated images by the way. So _libvirt_ needs a `firmware` to load as well as a `nvram` working memory in order to load this kind of images...

docs/providers/libvirt/r/domain.html.markdown
```diff
@@ -40,6 +40,42 @@ The following arguments are supported:
...
   have effect on the next reboot.
 
+Some extra arguments are also provided for using UEFI images:
+
+* `firmware` - (Optional) The UEFI rom images for exercising UEFI secure boot in a qemu
+environment. Users should usually specify one of the standard _Open Virtual Machine
+Firmware_ (_OVMF_) images available for their distributions. The file will be opened
+read-only.
+* `nvram` - (Optional) the _nvram_ variables file corresponding to the firmware. When provided,
+this file must be writable and specifi to this domain, as it will be updated when running
```
flavio: typo: "specifi" -> "specific"

libvirt/resource_libvirt_domain.go
```diff
@@ -106,6 +118,34 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		domainDef.Metadata.TerraformLibvirt.Xml = metadata.(string)
 	}
 
+	if firmware, ok := d.GetOk("firmware"); ok {
+		firmwareFile, err := homedir.Expand(firmware.(string))
```
flavio: I don't understand why we need `homedir` here. Is it because a user could set `firmware = "~/foo"`? If that's the case I would just remove this dependency and force the user to write the full path (like with all the configuration files used by other programs)

inercia: Yes, it is a convenience function for paths that are relative to the home directory. I think it makes things easier, but I will remove it if you think it is superfluous...

flavio:
The code LGMT, however it looks like your editor is not formatting the code properly (or at least it disagrees with go-fmt). Can you fix it so that it doesn't change the `switch` cases formatting?

libvirt/resource_libvirt_domain.go
flavio: I think we can remove it, this is something that is not done by any program I know. Also the user can place these files into the project directory and then terraform will automatically find them (`firmware = "foo"` -> `<pwd>/foo`)

inercia:
@flavio done

flavio:
Everything looks good to me. @dmacvicar do you have something against this PR, can we merge it?

mssola:
What's the status of this PR ?

flavio:
I think we can merge it. @inercia: feel free to do that after checking everything looks still good

PR #51, opened 2016-07-06
Activate stopped networks

flavio:
Make sure the network is active after the apply command is issued.

This is really handy when running apply on a host that has just been rebooted.


inercia:
LGTM

PR #52, opened 2016-07-12
Start storage pool if volume was not found

moio:
libvirt storage pools are not necessarily started by the time you call `terraform plan`.

In case they are not, you get this slightly misleading error:

```
Error refreshing state: 1 error(s) occurred:

* libvirt_volume.data_disk: Can't retrieve volume <STORAGE_PATH>/<VOLUME_NAME>
```

Fact that volume can't be retrieved is just because the pool has not been started yet.

Currently you would have to connect to the libvirt host via `virt-manager` or `virsh`, lookup the correct pool name and start it.

With this patch the provider will attempt to do that automatically, which improves UI/UX.

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -349,6 +349,14 @@ func resourceLibvirtVolumeRead(d *schema.ResourceData, meta interface{}) error {
 			}
 			defer volPool.Free()
 
+			active, err := volPool.IsActive()
+			if err != nil {
+				return fmt.Errorf("Error retrieving status of pool %s for volume %s: %s", volPoolName, volId, err)
+			}
+			if active {
+				return fmt.Errorf("Can't retrieve volume %s", d.Id())
+			}
+
 			err = volPool.Create(0)
 			if err != nil {
 				return fmt.Errorf("Error starting pool %s: %s", volPoolName, err)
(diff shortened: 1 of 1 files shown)


libvirt/resource_libvirt_volume.go
```diff
@@ -337,7 +337,31 @@ func resourceLibvirtVolumeRead(d *schema.ResourceData, meta interface{}) error {
...
+			volPoolName := d.Get("pool").(string)
+			log.Printf("[INFO] Volume %s not found, attempting to start pool %s", volId, volPoolName)
+
+			volPool, err := virConn.LookupStoragePoolByName(volPoolName)
+			if err != nil {
+				return fmt.Errorf("Error retrieving pool %s for volume %s: %s", volPoolName, volId, err)
+			}
+			defer volPool.Free()
+
+			err = volPool.Create(0)
```
flavio: Can you replace `0` with `libvirt.VIR_STORAGE_POOL_CREATE_NORMAL`?

flavio: Also can you make sure the pool is not active (see [this](http://libvirt.org/html/libvirt-libvirt-storage.html#virStoragePoolIsActive)) before invoking `Create`?

flavio:
The code LGTM. I would have preferred to have the new code wrapped up into a dedicated function, but that's something we can do later one once we do a refactoring of the whole project.

libvirt/resource_libvirt_volume.go
moio: > Also can you make sure the pool is not active (see this) before invoking Create?

It's not really needed (at least for directory-backed pools, which is the case I tested), still I added it.

> Can you replace 0 with libvirt.VIR_STORAGE_POOL_CREATE_NORMAL?

No, that constant does not exist in `libvirt-go`.

Issue #67, opened 2016-09-23
Failed apply doesn't clean up domains


cyphar:
On my machine `terraform apply` fails pretty often, and when `terraform apply` fails the domains that failed to set up are "stopped" but are not "undefined". So subsequent `apply`s will fail because the domains already exist. Is there a way for us to do cleanup when apply fails?

MalloZup:
encountered same issue :(

PR #69, opened 2016-10-03
Better handling of errors while creating a domain

flavio:
Sometimes if an error occurs at domain creation time, the resulting domain is **unknown** to terraform. That makes impossible to remove it, causing later runs of `terraform apply` to fail.

This commit marks the domain ID as a partial attribute that has to be saved to the `.tfstate` as soon as it's created.

That should fix issues like #67


mssola:
Makes sense. LGTM

Issue #24, opened 2016-06-02
cpu-model selection


fisabelle:
I just realized that my terraform VM are now spawned with the emulator CPU type.
I tried to figure out how to make qemu/libvirt use 'host-model' as default to workaround that.
My VM are very slow at performing key computations right now.
Do you have any workaround for that or plan to support specifying it ?

fisabelle:
I found out that with libvirt 1.3 I can use: 
 -device virtio-rng-pci
to supply a hardware random number generator to the VM.
Anyway to specify some devices within the terraform resources would be appreciated.

dmacvicar:
@fisabelle What do you mean with emulator CPU type? Can you paste the `ps` output that shows the qemu command line libvirt is using?

fisabelle:
Hi.
This is what is being used:

 /usr/libexec/qemu-kvm -name centos7_vm_domain_78,debug-threads=on -S -machine pc-i440fx-rhel7.2.0,accel=kvm,usb=off -m 4096 -realtime mlock=off -smp 2,sockets=2,cores=1,threads=1 -uuid 692e3a40-54e3-4921-a27e-77e943ea58e4 -no-user-config -nodefaults -chardev socket,id=charmonitor,path=/var/lib/libvirt/qemu/domain-54-centos7_vm_domain_78/monitor.sock,server,nowait -mon chardev=charmonitor,id=monitor,mode=control -rtc base=utc -no-shutdown -no-acpi -boot strict=on -device piix3-usb-uhci,id=usb,bus=pci.0,addr=0x1.0x2 -drive file=/home/vmstore/ramdisk/centos7_vm_disk_78,format=qcow2,if=none,id=drive-virtio-disk0 -device virtio-blk-pci,scsi=off,bus=pci.0,addr=0x4,drive=drive-virtio-disk0,id=virtio-disk0,bootindex=1 -netdev tap,fd=45,id=hostnet0,vhost=on,vhostfd=47 -device virtio-net-pci,netdev=hostnet0,id=net0,mac=52:54:00:a1:a3:78,bus=pci.0,addr=0x3 -spice port=5909,addr=127.0.0.1,disable-ticketing,seamless-migration=on -device cirrus-vga,id=video0,bus=pci.0,addr=0x2 -device virtio-balloon-pci,id=balloon0,bus=pci.0,addr=0x5 -msg timestamp=on


[...]

PR #70, opened 2016-10-04
Add a random number generator device to all the domains

flavio:
This should make the 1st boot of some domains faster and solve issue #24


mssola:
LGTM

PR #71, opened 2016-10-08
Fix random issues when handling multiple cloudinit volumes

flavio:
Sometimes the libvirt provider fails with cryptic messages when multiple cloudinit volumes are being used.

libvirtd contains several messages like `internal error: pool has asynchronous jobs running`. This seems to be caused by invoking `pool.Refresh()` while a volume is being uploaded.

This commit introduces a new structure that can be used to synchronize operations involving storage pools.

From my testing this solves the issues we have experiences so far.


mssola:
LGTM

inercia:
LGTM

Issue #79, opened 2016-10-24
Add volume_attachment resources or equivalent


moio:
I want to create a libvirt domain with a variable number of attached volumes (case in point is one main disk always present, and one optional data disk).

Currently this is not possible in any way I tried. Although the `disk` attribute looks like a map it cannot be passed parameters (the following does not work):

``` terraform
resource "libvirt_domain" "domain" {
disk = ["${var.disks}"]
```

Where `var.disks` is a map with a `volume_id` key and a correct id value. In fact not even the following works:

``` terraform
disk = ["${map("volume_id", "123")}"]
```

We get the following error on 0.7.7:

`Error reading config for libvirt_domain[domain]: At 31:11: root.disk[0]: not an object type for map (*ast.LiteralType)`

It would be great to have a way around this, eg:
- having a `disk_ids` plain list of ids instead of a list of maps/objects
- having a separate resource to associate disks to instances, similar to [aws_volume_attachment](https://www.terraform.io/docs/providers/aws/r/volume_attachment.html)
- being able to use a plain map for the existing `disk` variable.

Thanks in advance!!!

inercia:
Maybe we can find a solution in the context proposed in #46 ...

moio:
@inercia I do not think this is related.

Besides the fact [I would agree with closing #46](https://github.com/dmacvicar/terraform-provider-libvirt/issues/46#issuecomment-229858075) the RFC is about embedding parameters for the ephemeral, "must-have" OS basic volume in the domain resource, while this is about extra "data" disks which might or might be there - hence in variable quantity.

moio:
My proposal would be to add a `disk_ids` list in the domain resource, which could be populated via variables or interpolation.

This could even supersede the current `disk` variable, unless there is a reason I cannot see for keeping one map object instead of a plain id.

@flavio, @dmacvicar?

moio:
Note this comment about `disk_ids` not being the right way to address this: https://github.com/dmacvicar/terraform-provider-libvirt/pull/81#issuecomment-257312643

moio:
Asked [HashiCorp](https://github.com/hashicorp/terraform/issues/9747) if this is a core issue or by design.

tboerger:
I would take other providers as a template, Scaleway got a comparable approach.

moio:
[...]

PR #81, opened 2016-10-27
Simplify volume attachment

moio:
This changes the syntax to attach a volume to a domain from:

``` hcl
  ...
  disk = {
    volume_id = "${libvirt_volume.volume1.id}"
  }
  disk = {
    volume_id = "${libvirt_volume.volume2.id}"
  }
}
```

to:

``` hcl
  ...
  volume_ids = ["${libvirt_volume.volume1.id}, ${libvirt_volume.volume2.id}"]
}
```

Besides simplification, this has the advantage of allowing to use interpolation to create variable-volume resources. Example:

``` hcl

resource "libvirt_volume" "base_volume" {
  name = "base"
  base_volume_id = "${libvirt_volume.opensuse_leap.id}"
}

variable "data_volume_ids" {
  description = "array of ids of volumes additional to the base one"
  default = []
}

  ...
  volume_ids = ["${concat(list(libvirt_volume.base_volume.id), var.data_volume_ids)}"]
}
```

Original motivation for this was to DRY-up the package-mirror and minion-swarm modules in [sumaform](https://github.com/moio/sumaform).

Please see the discussion in issue #79, I needed to get this fixed so after some days I attempted myself. I am a total newbie in Go and in this project so please bear with me :wink:!

Also note this is on top of PR #80.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -73,13 +73,13 @@ func resourceLibvirtDomain() *schema.Resource {
 				Optional: true,
 				ForceNew: false,
 			},
-			"disk": &schema.Schema{
+			"volume_ids": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
 				Required: false,
 				ForceNew: true,
-				Elem: &schema.Resource{
-					Schema: diskCommonSchema(),
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
 				},
 			},
 			"network_interface": &schema.Schema{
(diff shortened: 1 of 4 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -73,13 +73,13 @@ func resourceLibvirtDomain() *schema.Resource {
...
-			"disk": &schema.Schema{
+			"volume_ids": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
 				Required: false,
 				ForceNew: true,
-				Elem: &schema.Resource{
-					Schema: diskCommonSchema(),
+				Elem: &schema.Schema{
+          Type:     schema.TypeString,
```
MaximilianMeister: you forgot to use `gofmt` on the file :)

dmacvicar:
The original idea of using volume_id inside disk is because not all disks are volumes, and not all volumes are disks.

So if you want to add a cdrom explicitly, you will need a `disk { }` with different parameters. Same for iSCSI disks, rbd and others. (Right now CDROMs are created/supported only implicitly by the cloud-init feature).

This PR assumes that all disks are volumes, that is the case right now, but closes the door to support other types to solve a completely unrelated problem of variable expansion.

dmacvicar:
After discussing on irc for a while I am:
* Convinced that this is not the way to go (because it mixes volumes and disks, when not all disks are volumes)
* Not sure if https://github.com/dmacvicar/terraform-provider-libvirt/issues/79 proposal about volume attachment is the way to go. I don't have the knowledge of the importance of the problem being solved (DRY and parametrization) with the amount of questions this raises.

In any case, while volume_attachment is also wrong in the sense that you don't attach volumes to domains, but disks to domains, and disks can be volumes:

* Allowing disks resources outside of domains, with same feature set as the inline disk (eg, to reference a volume).
* Adding a disk <-> domain attachment resource

May satisfy the conceptual libvirt model and solve @moio 's problem ?

So you would create a volume resource, a disk resource referencing it, and attachment resource.
Alternatively it could be:

disk resource
volume resource (disk references it)
and then allowing in domain `disk_ids` as an alternative to a list of `disk` blocks.

@moio  ?

moio:
@dmacvicar understand and agree. Closing this PR now, adding your comment to #79 as to the preferred ways to solve this issue.

PR #93, opened 2016-12-31
Partially working support for creating bridge connections

deasmi:
This is not the correct solution I'm sure, and terraform crashes
however correct changes are applied to libvirt host


deasmi:
Please DO NOT MERGE THIS as it causes terraform to crash and will not be a final solution.

PR #96, opened 2017-02-08
Add support for CoreOS ignition and console blocks

eamonnotoole:
A few changes here:

1.  CoreOS Ignition support:  A CoreOS Ignition file can be
    specified for a domain using the "coreos_ignition"
    parameter.  This requires the emission of additional
    XML and also the setting of the XML name-space to
    "http://libvirt.org/schemas/domain/qemu/1.0"
2.  "graphics" block:  A "graphics" block can be specified
    in a domain definition.  We've added this because we've
    found that for some builds of qemu the default "spice"
    emulator doesn't work and results in failure to boot the
    VMs.
3.  "console" block: One or more "console" blocks can be
    specified in a domain definition.  Note the description
    in domain.html.markdown, and the description in
    https://libvirt.org/formatdomain.html#elementsConsole

### DIFF
--- a/docs/providers/libvirt/r/domain.html.markdown
+++ b/docs/providers/libvirt/r/domain.html.markdown
@@ -41,6 +41,13 @@ The following arguments are supported:
   cloud-init won't cause the domain to be recreated, however the change will
   have effect on the next reboot.
 
+The following extra argument is provided for CoreOS images:
+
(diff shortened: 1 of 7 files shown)


flavio:
LGTM, thanks for the contributions

PR #98, opened 2017-02-17
CoreOS Ignition support updates

eamonnotoole:
Hopefully this patch is acceptable.  It changes the "coreos_ignition"
parameter introduced by a previous patch to a block.  I can redo the
patch if this change is deemed unacceptable.

This patch updates the CoreOS ignition support. "coreos_ignition"
is now a block, that looks as follows:

   coreos_ignition {
     content = <string> *Required*
     generated_ignition_file_dir = <string> *Optional*
   }

If the "coreos_ignition" block is present, then content must be
defined and can point to either the location of an existing
ignition file or the contents of a Terraform Ignition provider
object.  If the latter, you can also opitionally specify the
directory in which the ignition file will be written using
"generated_ignition_file_dir".  If this directory isn't specified,
"/tmp" will be used.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 	"time"
+	"path"
 
 	"crypto/sha256"
 	"encoding/hex"
@@ -82,11 +83,10 @@ func resourceLibvirtDomain() *schema.Resource {
 				ForceNew: false,
 			},
 			"coreos_ignition": &schema.Schema{
-				Type:     schema.TypeString,
+				Type:     schema.TypeMap,
 				Required: false,
 				Optional: true,
-				ForceNew: false,
-				Default:  "",
+				ForceNew: true,
 			},
 			"disk": &schema.Schema{
 				Type:     schema.TypeList,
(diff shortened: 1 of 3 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -155,22 +155,29 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
...
 		var ign_str string
 		if !file {
+			var ignitionFileLoc string
+			ignitionFileLoc, ok := ignitionMap["generated_ignition_file_dir"].(string)
+			if !ok {
+				ignitionFileLoc = "/tmp"
+			}
 			ignitionHash := hash(ignitionString)
-			tempFileName := fmt.Sprint("/tmp/", ignitionHash, ".ign")
+			tempFileName := path.Join(ignitionFileLoc, fmt.Sprint(ignitionHash, ".ign"))
```
inercia: What about using [`TempFile()`](https://golang.org/pkg/io/ioutil/#TempFile).

dmacvicar:
Now that I was trying to implement generic CmdLine support for the kernel and build the ignition code on top of that, I raised some questions in https://github.com/dmacvicar/terraform-provider-libvirt/pull/96, so I would prefer to delay merging this a bit until I get a better picture.

eamonnotoole:
I'm going to close this PR - I have a different one that I'd like to submit which adds a "libvirt_ignition" resource.

PR #103, opened 2017-02-22
create domain without graphics device

monstermunchkin:
When a graphics device is specified, a video device is automatically
added as well. This can cause problems on some exotic machines.

In order to prevent the issues, it should be possible to specify domains
without these devices.

**Note:** there will be no graphics per default, instead this setting has to be specified explicitly if one wants graphics support.

### DIFF
--- a/docs/providers/libvirt/r/domain.html.markdown
+++ b/docs/providers/libvirt/r/domain.html.markdown
@@ -248,6 +248,10 @@ resource "libvirt_domain" "my_machine" {
 }
 ```
 
+Note the following:
+* The `graphics` block is ignored for the architectures `s390x` and `ppc64`.
+
+
 The optional `console` block allows you to define a console for the domain.  The block
 looks as follows:
 ```
--- a/libvirt/domain_def.go
+++ b/libvirt/domain_def.go
@@ -25,13 +25,7 @@ type defDomain struct {
 		Disks             []defDisk             `xml:"disk"`
 		NetworkInterfaces []defNetworkInterface `xml:"interface"`
 		Console           []defConsole          `xml:"console"`
-		Graphics          struct {
-			Type     string `xml:"type,attr"`
-			Autoport string `xml:"autoport,attr"`
-			Listen   struct {
-				Type string `xml:"type,attr"`
-			} `xml:"listen"`
-		} `xml:"graphics"`
+		Graphics          *defGraphics          `xml:"graphics,omitempty"`
 		// QEMU guest agent channel
 		QemuGAChannel struct {
 			Type   string `xml:"type,attr"`
@@ -54,6 +48,14 @@ type defDomain struct {
 		XMLName xml.Name `xml:"qemu:commandline"`
 		Cmd     []defCmd `xml:"qemu:arg"`
 	}
+}
+
+type defGraphics struct {
+	Type     string `xml:"type,attr"`
+	Autoport string `xml:"autoport,attr"`
+	Listen   struct {
+		Type string `xml:"type,attr"`
+	} `xml:"listen"`
 }
 
 type defMetadata struct {
@@ -130,6 +132,7 @@ func newDomainDef() defDomain {
 	domainDef.VCpu.Placement = "static"
 	domainDef.VCpu.Amount = 1
 
+	domainDef.Devices.Graphics = &defGraphics{}
 	domainDef.Devices.Graphics.Type = "spice"
 	domainDef.Devices.Graphics.Autoport = "yes"
 	domainDef.Devices.Graphics.Listen.Type = "none"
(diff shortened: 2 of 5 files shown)


flavio:
The code LGTM, however I'm a bit concerned about the usability change.

What happens if you declare a node with no graphic card and then you try to access it with virt-manager?
Are you going to be able to see the screen and interact with it?

monstermunchkin:
If you don't declare graphics, you will need to specify a console to see something, at least that's the case for our exotic machines s390x and ppc64le. You can then interact with it normally.
If you specify neither graphics nor console, you won't see anything.

flavio:
> If you specify neither graphics nor console, you won't see anything.

I think we could have a UX problem there. The `graphics` section has always been optional but in fact we added by default a `spice` device that made everything work out of the box.

With this PR the usual behaviour will change and people won't have a way to use tools like `virt-manager` to interact with a domain.

I would be fine merging this PR as long as it also updates the documentation stating how to have a "screen" inside of tools like `virt-manager` (basically how to write a `.tf` domain that uses `spice`.

What is the opinion of other people like @dmacvicar and @inercia about this PR?

monstermunchkin:
> I think we could have a UX problem there. The graphics section has always been optional but in fact we added by default a spice device that made everything work out of the box.

We could also keep the default behaviour, and simply have a special value for the graphics type, e.g. `none`, which disables graphics:
```
graphics {
  type = "none"
}
```

> I would be fine merging this PR as long as it also updates the documentation stating how to have a "screen" inside of tools like virt-manager (basically how to write a .tf domain that uses spice.

In addition, we could print a warning stating that they won't be able to interact with the VM if neither graphics nor console are enabled,

dmacvicar:
Can you explain a bit more about "This _can_ cause problems on some _exotic_ machines.".

I don't like the idea of taking more and more of the "cloud" feeling of the provider into the direction of a `libvirt.xml` to `terraform.hcl` mapper just because of some corner case.

monstermunchkin:
> Can you explain a bit more about "This can cause problems on some exotic machines.".

Qemu (in IBS) has `spice` disabled for architectures like s390x and ppc64le so it will not work unless we start supporting it for these architectures. Using `vnc` as graphics caused problems as well on both machines.

The problem is not always the `graphics` device itself, but the `video` device which is [automatically added]( https://libvirt.org/formatdomain.html#elementsVideo) when specifying the first. This can lead to issues such as a missing `vgabios-stdvga.bin` file.

Omitting the `graphics` and `video` devices, and just using a `console` however works like a charm and there is no need to extend support for other packages.

dmacvicar:
@flavio Do you think it would make more sense to disable graphics in those platforms explicitly (eg. terraform-provider-libvirt not adding the xml block on those architectures) and documenting this?.

flavio:
> @flavio Do you think it would make more sense to disable graphics in those platforms explicitly (eg. terraform-provider-libvirt not adding the xml block on those architectures) and documenting this?.

I like this approach.

@monstermunchkin I guess there's a way to figure out the architecture of the domain being created

monstermunchkin:
>> @flavio Do you think it would make more sense to disable graphics in those platforms explicitly (eg. terraform-provider-libvirt not adding the xml block on those architectures) and documenting this?.

> I like this approach.

I don't like this approach. This only makes sense if you run `terraform` on the same host the VM is being created, which we won't be doing. The reason for that is Terraform currently doesn't build on s390x or ppc64le (on IBS). We will be running `terraform` on an amd64 host which then uses SSH to connect to "the other side".

flavio:
The [virConnectGetCapabilities](http://libvirt.org/html/libvirt-libvirt-host.html#virConnectGetCapabilities) API returns also the architecture of the libvirt host you are connected to.

For example:
```go
flavio@melange ~/go/src/_/virtinfo » vim main.go 
flavio@melange ~/go/src/_/virtinfo » cat main.go
package main

import (
	"encoding/xml"
	"fmt"
	"os"

	libvirt "github.com/dmacvicar/libvirt-go"
)

type HostCapabilities struct {
	XMLName xml.Name `xml:"capabilities"`
	Host    struct {
		XMLName xml.Name `xml:"host"`
		CPU     struct {
			XMLName xml.Name `xml:"cpu"`
			Arch    string   `xml:"arch"`
		}
	}
}

func main() {
	conn, err := libvirt.NewVirConnection("")

	if err != nil {
		fmt.Errorf("Something went wrong: %v", err)
		os.Exit(1)
	}

	info, err := conn.GetCapabilities()
	if err != nil {
		fmt.Errorf("Something went wrong: %v", err)
		os.Exit(1)
	}
	capabilities := HostCapabilities{}
	xml.Unmarshal([]byte(info), &capabilities)
	fmt.Printf("%s\n", capabilities.Host.CPU.Arch)
}
```

This will print the architecture of localhost (in my case it would be `x86_64`)

flavio:
👏

PR #104, opened 2017-02-23
cloud-init user_data

### LINKED ISSUE #66: Limited options for cloud-init
It would be great to have cloud-init option to inject custom scripts in the cluster. Right now that is not supported and it gets difficult to predict the IP addresses for the VM's in case of a dynamic cluster.

inercia:
Provide a `user_data` argument for specifying raw user data.

The `user_data`  provided is merged with the other parameters we've been using so far (like the `ssh_authorized_keys`), so this allow us to still use some of those conveniences and, at the same time, it should be backwards compatible...

Closes #66

### DIFF
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -29,7 +29,8 @@ type defCloudInit struct {
 		LocalHostname string `yaml:"local-hostname,omitempty"`
 		InstanceID    string `yaml:"instance-id"`
 	}
-	UserData struct {
+	UserDataRaw string `yaml:"user_data"`
+	UserData    struct {
 		SSHAuthorizedKeys []string `yaml:"ssh_authorized_keys"`
 	}
 }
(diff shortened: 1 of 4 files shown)


flavio:
Some of the changes of this PR are related with `gofmt`. Can you get rid of them by pushing the gofmt changes into a dedicated commit on master and then rebasing this PR?

I think you can submit the `gofmt` changes straight to master, without going through a PR.

libvirt/cloudinit_def.go
```diff
@@ -22,6 +23,10 @@ import (
 const USERDATA string = "user-data"
 const METADATA string = "meta-data"
 
+type UserDataStruct struct {
```
inercia: `SomethingStruct` is probably the worst name you could choose for a `struct`... :-)

flavio: Any hint? :)

inercia: Hmmm... is just `UserData` taken? Or maybe `CloudInitUserData`? :-)

flavio: Done

flavio: done

PR #105, opened 2017-02-28
Add network option to disable DHCP services

rawlinp:
When creating a VM with multiple network interfaces on different
networks, you typically don't want a DHCP server running on every
network. For example, a CoreOS VM will send out a DHCP request on every
network interface by default when it boots up (given no other config
from say, Ignition). This is problematic for the VM's networking
because it will end up with multiple default routes.

An example use case of this is a VM with DHCP enabled on one NIC (to be
used as the "management" network), from which the rest of the VM's
interfaces will be configured.

### DIFF
--- a/docs/providers/libvirt/r/domain.html.markdown
+++ b/docs/providers/libvirt/r/domain.html.markdown
@@ -186,7 +186,9 @@ The network will _NOT_ be managed by the Terraform/libvirt provider.
 * `network_id` - (Optional) The ID of a network resource to attach this interface to.
 The network will be under the control of the Terraform/libvirt provider.
 * `mac` - (Optional) The specific MAC address to use for this interface.
-* `addresses` - (Optional) An IP address for this domain in this network
+* `addresses` - (Optional) An IP address for this domain in this network.
+Note: if this network has DHCP disabled (`dhcp = false`), specifying an address before creation
+will override that setting and enable DHCP services on this network.
 * `hostname` - (Optional) A hostname that will be assigned to this domain resource in this network.
 * `wait_for_lease`- (Optional) When creating the domain resource, wait until the network
 interface gets a DHCP lease from libvirt, so that the computed ip addresses will be available
--- a/docs/providers/libvirt/r/network.markdown
+++ b/docs/providers/libvirt/r/network.markdown
@@ -50,6 +50,7 @@ The following arguments are supported:
 * `domain` - The domain used by the DNS server.
 * `addresses` - A list of (0 or 1) ipv4 and (0 or 1) ipv6 subnets in CIDR notation
    format for being served by the DHCP server. Address of subnet should be used.
+* `dhcp` - (Optional) Enable DHCP services on this network (default is `true`).
 * `mode` -  One of:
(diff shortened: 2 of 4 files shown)


inercia:
I'm not sure if it makes sense to have `addresses = [something]` and `dhcp = false`, so I was wondering if maybe we could skip this new `dhcp` argument and have the same effect when `addresses` is not defined (or defined to be an empty set).

rawlinp:
@inercia - thanks for the quick feedback. I like that idea, but the 'nat' forwarding mode requires an `<ip family='ipv4' address='x.x.x.x' prefix='x'/>` element with an address that can be used on the bridge associated with the virtual network. That IP is then used as a gateway on the host and is necessary to perform NAT, so it seems that regardless of whether or not DHCP services are enabled, the `addresses` range will need to be supplied. Does that seem right?

Another thought I had was maybe forgetting the `dchp` option, specifying the `addresses` option on the network, and determining whether or not to start DHCP services based upon whether or not `addresses` is present on a domain's `network_interface`. That is, the libvirt network's xml definition would leave out the `<dhcp>` element by default, then as domains are created, if they have `addresses` for their `network_interfaces`, this libvirt provider would dynamically add the `<dhcp>` element along with the `<host>` entries, similar to what it does today (although I don't think it currently adds the `<range>` element). However, I'm not sure this is a better approach.

rawlinp:
I tried out that 2nd possible approach [1] that dynamically updates the network to start DHCP services once domains are created with requested addresses/hostnames. I'm still unsure if that's the way to go. It would also require a small change to one line [2] in libvirt-go.

[1] https://github.com/rawlinp/terraform-provider-libvirt/commit/cc41dd69ffe2a12e8cbdee7196524f00625f6d23
[2] https://github.com/dmacvicar/libvirt-go/blob/master/network.go#L173

inercia:
I find more elegant this _dynamic_ solution (to add DHCP services when `addresses` is specified) even if that requires some changes to _libvirt-go_, but I don't see if there could be side effects in some other scenarios. I guess it would require some testing...

rawlinp:
I've created a new PR for the dynamic solution: #111. I've tested each of the various network types (`nat`, `none`, `route`, `bridge`), and the new DHCP behavior doesn't seem to have unintended consequences.

I found that #111 doesn't actually require a change to libvirt-go like I originally thought. It turned out to be a race condition rather than a persistence issue, but the race condition manifested itself much more often when the network definition wasn't persisted. With some synced access to the network definition, the race condition goes away altogether.

As a side note, I do think I found an unrelated bug with the `none` and `bridge` network types, however, because interfaces of those network types aren't added to a domain's xml definition. See [1] and [2].

[1] https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_domain.go#L348
[2] https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_domain.go#L410

inercia:
So can we close this PR and focus on #111 ?

rawlinp:
Yes, we can focus on #111.

PR #106, opened 2017-03-02
simplify the install documentation

MalloZup:
Setting $GOPATH and installing go should be a prereq of go-devs.

Additionally go get -u does already the clone in src


MalloZup:
For me now it's ok. I just simplified the README, and removed the .terraform dot file that is not needed.

📻

PR #107, opened 2017-03-02
Write Ignition file as a volume in a libvirt storage pool

eamonnotoole:
This avoids the problem where the Ignition file is remote to
the host on which the libvirt domain is being created.  We've
added a "libvirt_ignition" resource which manages the Ignition
file in the libvirt volume - creates it, and destroys it.  The
"coreos_ignition" field in the libvirt_domain definition must
point to the Id of a "libvirt_ignition" resource.

The code is modelled on that used for CloudInit.


libvirt/resource_libvirt_domain.go
```diff
@@ -12,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"crypto/sha256"
+	"encoding/hex"
```
inercia: Please reformat these imports.

libvirt/resource_libvirt_coreos_ignition.go
```diff
@@ -0,0 +1,95 @@
+package libvirt
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform/helper/schema"
+	"log"
+)
```
eamonnotoole: Done

libvirt/resource_libvirt_domain.go
eamonnotoole: Done

eamonnotoole:
@inercia @dmacvicar @flavio What's the prognosis with this PR?  There haven't been any comments on it in two months.

flavio:
LGTM too

PR #110, opened 2017-03-06
Attempt to fix test run and code coverage issues

eamonnotoole:
With travis and with coveralls.  First, fixes applied to .travis.yml
to install qemu-kvm and to set TF_ACC.  Also we run "go test" with
"-v"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,18 +5,15 @@ go:
   - 1.7
 
 before_install:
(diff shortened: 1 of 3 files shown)


flavio:
I'm in favor of merging this PR. I'm not 100% sure about using this external ppa, but I don't see anything dangerous in doing that.
I wish Travis had a more recent ubuntu image...

What do the others think about it?

PR #111, opened 2017-03-07
Dynamically start DHCP services on networks

rawlinp:
Rather than starting DHCP services on a network when it is created, they
should be started dynamically when domains are created with interfaces
on the network with requested addresses (or hostnames).

Because domains can be created in parallel, some read/write access to
the libvirt network definition must be synced to prevent an error that
occurs when a DHCP range is added twice to a network.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -519,35 +519,6 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 	return nil
 }
 
-// enableDHCP enables DHCP services on a given libvirt network
-func enableDHCP(network *libvirt.VirNetwork) error {
-	// mutex required due to concurrent access of the network definition
-	// (adding the same DHCP range twice causes an error while creating the libvirt domain)
-	dhcpMutex.Lock()
-	defer dhcpMutex.Unlock()
-	networkDef, err := newDefNetworkfromLibvirt(network)
-	if err != nil {
-		return err
-	}
-	for _, ip := range networkDef.Ips {
-		if ip.Dhcp == nil {
-			ipNet, err := GetIPNet(ip.Address, ip.Prefix)
-			if err != nil {
-				return err
-			}
-			start, end := NetworkRange(ipNet)
-
-			// skip the network, gateway, and broadcast addresses
-			start[len(start)-1] = start[len(start)-1] + 2
-			end[len(end)-1]--
-			if err := addDHCPRange(network, start.String(), end.String()); err != nil {
-				return err
-			}
-		}
-	}
-	return nil
-}
-
 func resourceLibvirtDomainUpdate(d *schema.ResourceData, meta interface{}) error {
 	log.Printf("[DEBUG] Update resource libvirt_domain")
 
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
@@ -18,6 +19,7 @@ import (
 )
 
 var PoolSync = NewLibVirtPoolSync()
+var dhcpMutex = &sync.Mutex{}
 
 func init() {
 	spew.Config.Indent = "\t"
@@ -359,10 +361,23 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 			}
 
 			hostname := domainDef.Name
+			hostnameSet := false
 			if hostnameI, ok := d.GetOk(prefix + ".hostname"); ok {
 				hostname = hostnameI.(string)
+				hostnameSet = true
 			}
-			if addresses, ok := d.GetOk(prefix + ".addresses"); ok {
+			var addresses interface{}
+			addressesSet := false
+			if a, ok := d.GetOk(prefix + ".addresses"); ok {
+				addresses = a
+				addressesSet = true
+			}
+			if hostnameSet || addressesSet {
+				if err := enableDHCP(&network); err != nil {
+					return err
+				}
+			}
+			if addressesSet {
 				// some IP(s) provided
 				for _, addressI := range addresses.([]interface{}) {
 					address := addressI.(string)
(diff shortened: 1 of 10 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -344,16 +346,31 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 			if err != nil {
 				return fmt.Errorf("Error retrieving network name: %s", err)
 			}
+			dhcpMutex.Lock()
 			networkDef, err := newDefNetworkfromLibvirt(&network)
+			dhcpMutex.Unlock()
```
inercia: Why are we locking around `newDefNetworkfromLibvirt`?

libvirt/resource_libvirt_domain.go
```diff
@@ -344,16 +346,31 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
...
 			if !networkDef.HasDHCP() {
 				continue
 			}
 
 			hostname := domainDef.Name
+			hostnameSet := false
 			if hostnameI, ok := d.GetOk(prefix + ".hostname"); ok {
 				hostname = hostnameI.(string)
+				hostnameSet = true
```
inercia: What about `if hostnameI, hostnameSet := d.GetOk(prefix + ".hostname"); hostnameSet {... ` ?

libvirt/resource_libvirt_domain.go
inercia: Oh, I see, I think it is for avoiding concurrent changes made by `enableDHCP`, right?

libvirt/resource_libvirt_domain.go
```diff
@@ -494,6 +512,31 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 	return nil
 }
 
+// enableDHCP enables DHCP services on a given libvirt network
+func enableDHCP(networkDef defNetwork, network *libvirt.VirNetwork) error {
+	// mutex required due to concurrent reads of the network definition
+	// (adding the same DHCP range twice causes an error while creating the libvirt domain)
+	dhcpMutex.Lock()
+	defer dhcpMutex.Unlock()
```
inercia: What if we move the locks to inside `addDHCPRange`?

libvirt/resource_libvirt_domain.go
rawlinp: Yes, in my testing I had a loop that would create a network of each type and 2 VMs with a NIC on each network, and without this locking it would fail very quickly, whereas with locking I couldn't get it to fail. However, now that I look at it again, I can see the following case:

1. Thread1 acquires lock, reads network1 (no DHCP range), releases lock
2. Thread2 acquires lock, reads network1 (no DHCP range), releases lock
3. Thread 1 advances to enableDHCP, acquires lock, sees no DHCP, enables DHCP, releases lock
4. Thread 2 advances to enableDHCP, acquires lock, sees no DHCP, enabled DHCP -> ERROR

It seems we might be better to just *not* lock here, then read the libvirt network *again* inside the locking of enableDHCP. I think that might be best, because this locking seems to just protect against the case where Thread 1 is currently enabling DHCP while Thread 2 is a bit further behind and is trying to read the network.

libvirt/resource_libvirt_domain.go
rawlinp: I would prefer that, but I remember that not working for some reason. When doing what you suggest, it seems golang doesn't want you to use `hostnameI` or `hostnameSet` outside of the `if` block.

libvirt/resource_libvirt_domain.go
rawlinp: We need to lock around this entire function because multiple threads could both read that `ip.Dhcp == nil` and attempt to add a DHCP range, resulting in error. Also, I believe we need to read the libvirt network again inside the locks like I mentioned in the comment above.

rawlinp:
@inercia, in regards to your earlier comment about making the simplest case harder, how does the user find the IP that would be allocated to their VM via DHCP? I think it might fly by in the output of `terraform apply`, but isn't it easier to just request a static IP from the CIDR they already configured for the network rather than read through the command output? Is there an easier way to find the IP to use in that case?

Would it be too complex to detect the case where the user hasn't specified any `addresses` on any `network_interface` and enable DHCP on the first interface's network just to make the VM usable?

inercia:
@rawlinp So in my view the best behaviour from a user point of view would be:

* if the user creates a network with no `addresses` at all, the network is created with a DHCP serving in a default range.
* if the user creates a network with `addresses = [something]`, the network is created with a DHCP serving in that range.
* if the user creates a network with no `addresses = []`, the network is created with no DHCP.

The first case is the default case I was talking about, the simplest case that most users will use with DHCP and NAT. What do you think? Not sure this is what we currently have...

inercia:
@rawlinp Regarding getting the IP address for a VM in your cluster, all the IPs are saved in the `tfstate` file, but you can also print it from your `.tf` file with something like:

```
output "ip_something" {
  value = "${libvirt_domain.something.network_interface.0.addresses.0}"
}
```

rawlinp:
1. `if !provided { enableDHCP(DEFAULT_RANGES) }`
What would `DEFAULT_RANGES` be? How can we be sure it doesn't clash with any other private CIDR in use on the user's system?
2. `if provided && len(addresses) > 0 { enableDHCP(addresses) }`
This is what I'm trying to avoid by only enabling DHCP when a VM has specified `addresses` on one of its NICs. With multiple NICs on networks with DHCP enabled, a VM can end up with multiple default routes (a default route for every NIC on a DHCP-enabled network). This way, a VM can specify `addresses` on only a single NIC, and the rest can be configured manually if they so choose.
3. `if provided && len(addresses) == 0 { disableDHCP() }`
This kind of ties in with Case 1 above, because without any provided `addresses` on a network, the user cannot use NAT functionality or DHCP because we won't know what IP to use on the virtual bridge (we'd have to access the user's networking stack and compute a non-clashing CIDR to use).

Being able to bring up a VM without specifying its network's `addresses` seems like a nice idea for simplicity's sake, but I don't see it being able to use NAT/DHCP unless it's at least provided its own CIDR for the network. If no CIDR is provided, I don't think it's within the scope of the terraform libvirt provider to fully access the user's networking stack to attempt to compute a non-clashing default CIDR to use.

inercia:
@rawlinp Ok, maybe you are right: to try to compute a non-clashing range would require some work (and some assumptions). So I think we could require `addresses` from users, but please update the documentation clearly stating that, including some examples on how to create a _ready-to-use_ network as well. Most users will start with the DHCP & NAT case (and will probably not go further than that), so that should be well documented.

I also think you have some pending changes regarding where to put the `Lock`s when enabling DHCP and so...

inercia:
@rawlinp This needs to be rebased and resolve some conflicts...

rawlinp:
@inercia I don't think I will have the time to finish working on this patch. If you or someone else still find it valuable, please feel free to take it over.

Bischoff:
What's the problem with starting DHCP when you create the network? It seems to me the right moment.

Also, I need to be able to create networks without DHCP and with fixed IP addresses. If I'm not wrong, this patch would make it impossible. So I would vote very strongly against this patch.

Sorry in advance if I misunderstood something.

rawlinp:
Closing as I can no longer work on this PR, and it's vastly out of date.

PR #112, opened 2017-03-10
Be more verbose/precise by debugging failures.

MalloZup:
We have a race condition here, see issue 88 and printing same string
for two differents error in code is not the best solution for debug.

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -379,7 +379,7 @@ func resourceLibvirtVolumeRead(d *schema.ResourceData, meta interface{}) error {
 		// attempt a new lookup
 		volume, err = virConn.LookupStorageVolByKey(d.Id())
 		if err != nil {
-			return fmt.Errorf("Can't retrieve volume %s", d.Id())
+			return fmt.Errorf("Second attempt: Can't retrieve volume %s", d.Id())
 		}
 	}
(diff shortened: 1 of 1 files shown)


MalloZup:
ok, i found out that we have currently : 

```
libvirt/resource_libvirt_domain.go:261:			return fmt.Errorf("Can't retrieve volume %s", volumeKey)
libvirt/resource_libvirt_domain.go:921:		return disk, fmt.Errorf("Can't retrieve volume %s", volumeKey)
libvirt/resource_libvirt_volume_test.go:93:			return fmt.Errorf("Can't retrieve volume key: %s", err)
libvirt/resource_libvirt_volume.go:256:			return fmt.Errorf("Can't retrieve volume %s", baseVolumeId.(string))
libvirt/resource_libvirt_volume.go:284:			return fmt.Errorf("Can't retrieve volume %s", baseVolumeName.(string))
libvirt/resource_libvirt_volume.go:353:			return fmt.Errorf("Can't retrieve volume %s", d.Id())
libvirt/resource_libvirt_volume.go:371:			return fmt.Errorf("Can't retrieve volume %s", d.Id())
libvirt/resource_libvirt_volume.go:382:			return fmt.Errorf("Second attempt: Can't retrieve volume %s", d.Id())
```
If you like the idea i can try to improve all this common msg, because the don't give the context of the failure.

PR #115, opened 2017-03-20
Add scsi controller option and wwn to virtual disks

eamonnotoole:
We've added the facility to define a disk bus type of scsi to the
specification of a disk in the domain definition.  If the disk is
a scsi disk, a random wwn is generated unless a wwn is provided for
that disk.

The disk stanza now looks as follows:

   disk {
     volume_id = "${libvirt_volume.mydisk.id}"
     scsi = "yes"
     wwn = "05abcd123456789a"
   }

Having "scsi" present with any value will specify a scsi bus for
the disk.  If "wwn" is present for a scsci disk then the value of
"wwn" is used for the disk wwn, otherwise a random value is generated.

### DIFF
--- a/libvirt/disk_def.go
+++ b/libvirt/disk_def.go
@@ -2,12 +2,17 @@ package libvirt
 
 import (
 	"encoding/xml"
+	"math/rand"
+	"time"
 )
+
+const OUI = "05abcd"
 
 type defDisk struct {
 	XMLName xml.Name `xml:"disk"`
 	Type    string   `xml:"type,attr"`
 	Device  string   `xml:"device,attr"`
+	Wwn     string   `xml:"wwn,omitempty"`
 	Format  struct {
 		Type string `xml:"type,attr"`
 	} `xml:"format"`
@@ -49,4 +54,14 @@ func newCDROM() defDisk {
 	disk.Driver.Type = "raw"
 
 	return disk
+}
+
+func randomWWN(strlen int) string {
+	const chars = "abcdef0123456789"
+	rand.Seed(time.Now().UTC().UnixNano())
+	result := make([]byte, strlen)
+	for i := 0; i < strlen; i++ {
+		result[i] = chars[rand.Intn(len(chars))]
+	}
+	return OUI + string(result)
 }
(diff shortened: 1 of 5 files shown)


libvirt/disk_def.go
```diff
@@ -50,3 +53,13 @@ func newCDROM() defDisk {
 
 	return disk
 }
+
+func generateWWN(OUI string, strlen int) string {
```
inercia: Could it be possible to name this function `randomWWN`? We have a couple of `random*` functions so this woiuuld follow the same pattern...

libvirt/resource_libvirt_domain.go
```diff
@@ -256,6 +256,9 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		domainDef.Devices.Console = consoles
 	}
 
+	controller := defController{Type:"scsi", Model:"virtio-scsi"}
+	domainDef.Devices.Controller = append(domainDef.Devices.Controller, controller)
```
inercia: I see you are adding this controller by default, and I was wondering what happens when the disk is not SCSI...

eamonnotoole: The controller is still there but if the disk isn't scsi it won't be used.  As far as I can tell libvirt doesn't remove the controller definition if it isn't used.

libvirt/resource_libvirt_domain.go
```diff
@@ -265,6 +268,14 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		diskKey := fmt.Sprintf("disk.%d", i)
 		diskMap := d.Get(diskKey).(map[string]interface{})
 		volumeKey := diskMap["volume_id"].(string)
+		if _, ok := diskMap["scsi"].(string); ok {
+			disk.Target.Bus = "scsi"
+			if wwn, ok := diskMap["wwn"].(string); ok {
+				disk.Wwn = wwn
+			} else {
+				disk.Wwn = generateWWN("05abcd", 10)
```
eamonnotoole: That's there as a "fake" valid OUI.  I'll refactor to a const.

libvirt/disk_def.go
eamonnotoole: I'll make that change

libvirt/resource_libvirt_domain.go
eamonnotoole: I'll change the code so that the scsi controller is only defined if a scsi disk has been specified.

PR #116, opened 2017-03-21
Use If-Modified-Since for downloading images

### LINKED ISSUE #109: Implement If-Modified-Since when downloading images
Our HTTP client currently downloads the image specified on `source` even when the image in the server is the same as a local copy previously downloaded.

For ths case where a local copy is already in the pool, we could perform the HTTP request with the [`If-Modified-Since`](https://developer.mozilla.org/es/docs/Web/HTTP/Headers/If-Modified-Since) header (using the [modification time](https://golang.org/pkg/os/#FileInfo) of the local image), so it would be downloaded only if the remote file has been updated.

inercia:
The idea is to use `If-Modified-Since` when downloading an image from a `source`, so we can skip the download when _a)_ the image was downloaded in the past and _b)_ the modification time is the same

Closes #109

### DIFF
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -108,14 +108,24 @@ func (ci *defCloudInit) CreateAndUpload(virConn *libvirt.VirConnection) (string,
 	defer volume.Free()
 
 	// upload ISO file
-	stream, err := libvirt.NewVirStream(virConn, 0)
-	if err != nil {
-		return "", err
-	}
-	defer stream.Close()
+	copier := func(src io.Reader) error {
+		stream, err := libvirt.NewVirStream(virConn, 0)
+		if err != nil {
+			return err
+		}
+		defer stream.Close()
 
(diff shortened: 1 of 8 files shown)


inercia:
When using a `source` that points to a local file we also use the modification time. However, maybe we should consider doing a md5 of both files (the existing one and the `source`d one), but I'm concerned about environments where that could be an expensive operation (ie, a NFS mounted volume). Any comments on this?

flavio:
> When using a source that points to a local file we also use the modification time. However, maybe we should consider doing a md5 of both files (the existing one and the sourced one), but I'm concerned about environments where that could be an expensive operation (ie, a NFS mounted volume). Any comments on this?

I'm fine not calculating a checksum of the file. We can always implement that later on if we run into issues. I prefer to keep the check fast.

flavio:
LGTM! 👏

flavio:
I think we simply forgot to merge it 🤦‍♂️ , merging...

PR #117, opened 2017-03-21
Delete domain IP addresses from networks

eamonnotoole:
We've found that, when executing destroy/create cycles on
a domain we get network failures since the old entries for
the domain are still present in the virtual network(s). We've
also found that, on occasion, we don't get complete information
on a domain's interfaces from ListAllInterfaceAddresses. This
patch implements a couple of changes in light of these issues:

1.  We now remove the dhcp entries for a domain when the domain
    is deleted
2.  We check the number of networks returned from getDomainInterfaces
    against the number of networks that we see in the domain XML.
    If the second number is greater than the first then we get
    interface information directly from the networks attached to
    the domain.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -593,6 +593,8 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 				continue
 			}
 
+			networkXMLDesc, err := network.GetXMLDesc(0)
+			log.Printf("[DEBUG] network def xml in create: %s", networkXMLDesc)
 			hostname := domainDef.Name
 			if hostnameI, ok := d.GetOk(prefix + ".hostname"); ok {
 				hostname = hostnameI.(string)
@@ -605,7 +607,6 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 					if ip == nil {
 						return fmt.Errorf("Could not parse addresses '%s'", address)
 					}
-
 					log.Printf("[INFO] Adding IP/MAC/host=%s/%s/%s to %s", ip.String(), mac, hostname, networkName)
 					if err := updateOrAddHost(network, ip.String(), mac, hostname); err != nil {
 						return err
@@ -711,7 +712,7 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 	log.Printf("[INFO] Domain ID: %s", d.Id())
 
 	if len(waitForLeases) > 0 {
-		err = domainWaitForLeases(domain, waitForLeases, d.Timeout(schema.TimeoutCreate))
+		err = domainWaitForLeases(domain, waitForLeases, d.Timeout(schema.TimeoutCreate), domainDef, virConn)
 		if err != nil {
 			return err
 		}
@@ -986,8 +987,8 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	d.Set("filesystems", filesystems)
 
-	// look interfaces with addresses
-	ifacesWithAddr, err := domainGetIfacesInfo(*domain)
+	// lookup interfaces with addresses
+	ifacesWithAddr, err := domainGetIfacesInfo(*domain, domainDef, virConn)
 	if err != nil {
 		return fmt.Errorf("Error retrieving interface addresses: %s", err)
 	}
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -335,6 +335,65 @@ func TestAccLibvirtDomain_NetworkInterface(t *testing.T) {
 	})
 }
 
+func TestAccLibvirtDomain_CheckDHCPEntries(t *testing.T) {
+	var domain libvirt.Domain
+	var network libvirt.Network
+
+	var configWithDomain = fmt.Sprintf(`
+	    resource "libvirt_network" "acceptance-test-network" {
+		    name = "acceptance-test-network"
+		    mode = "nat"
+		    domain = "acceptance-test-network-local"
+		    addresses = ["192.0.0.0/24"]
+	    }
+
+            resource "libvirt_domain" "acceptance-test-domain" {
+                    name = "terraform-test"
+                    network_interface {
+                            network_id = "${libvirt_network.acceptance-test-network.id}"
+                            hostname = "terraform-test"
+                            addresses = ["192.0.0.2"]
+                    }
+            }`)
+
+	var configWithoutDomain = fmt.Sprintf(`
+	    resource "libvirt_network" "acceptance-test-network" {
+		    name = "acceptance-test-network"
+		    mode = "nat"
+		    domain = "acceptance-test-network-local"
+		    addresses = ["192.0.0.0/24"]
+	    }`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtDomainDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: configWithDomain,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLibvirtDomainExists("libvirt_domain.acceptance-test-domain", &domain),
+					testAccCheckLibvirtNetworkExists("libvirt_network.acceptance-test-network", &network),
+				),
+			},
+			resource.TestStep{
+				Config: configWithoutDomain,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLibvirtDestroyLeavesIPs("libvirt_network.acceptance-test-network",
+						"192.0.0.2", &network),
+				),
+			},
+			resource.TestStep{
+				Config:             configWithDomain,
+				ExpectNonEmptyPlan: true,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLibvirtDomainExists("libvirt_domain.acceptance-test-domain", &domain),
+				),
+			},
+		},
+	})
+}
+
 func TestAccLibvirtDomain_Graphics(t *testing.T) {
 	var domain libvirt.Domain
 
@@ -717,6 +776,38 @@ func testAccCheckLibvirtURLDisk(u *url.URL, domain *libvirt.Domain) resource.Tes
 	}
 }
 
+func testAccCheckLibvirtDestroyLeavesIPs(n string, ip string, network *libvirt.Network) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
+			return fmt.Errorf("Not found: %s", n)
+		}
+
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("No libvirt network ID is set")
+		}
+
+		virConn := testAccProvider.Meta().(*Client).libvirt
+
+		retrieveNetwork, err := virConn.LookupNetworkByUUIDString(rs.Primary.ID)
+
+		if err != nil {
+			return err
+		}
+
+		networkDef, err := newDefNetworkfromLibvirt(retrieveNetwork)
+
+		for _, ips := range networkDef.IPs {
+			for _, dhcpHost := range ips.DHCP.Hosts {
+				if dhcpHost.IP == ip {
+					return nil
+				}
+			}
+		}
+		return fmt.Errorf("Hostname with ip '%s' does not have a dhcp entry in network", ip)
+	}
+}
+
 func testAccCheckLibvirtDomainKernelInitrdCmdline(domain *libvirt.Domain, kernel *libvirt.StorageVol, initrd *libvirt.StorageVol) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		xmlDesc, err := domain.GetXMLDesc(0)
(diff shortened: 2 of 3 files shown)


inercia:
>  when executing destroy/create cycles on a domain we get network failures since the old entries for the domain are still present in the virtual network(s). 

@eamonnotoole So you mean that, even when a network is deleted, the DHCP server is still running? Or you mean that some configuration is remaining somewhere, and when a new network is created it takes that configuration somehow? Are you sure the network has been destroyed?

libvirt/resource_libvirt_domain.go
```diff
@@ -996,3 +1048,56 @@ func getDomainInterfaces(domain *libvirt.VirDomain) ([]libvirt.VirDomainInterfac
 
 	return interfaces, nil
 }
+
+func getDomainInterfacesFromNetworks(domain defDomain, meta interface{}) ([]libvirt.VirDomainInterface, error) {
+
+	log.Printf("[DEBUG] getting domain addresses from networks")
+	var virDomain = make([]libvirt.VirDomainInterface, 0)
```
inercia: `virDomain` looks like a _domain_, but it is really a _list of interfaces_. What do you think about `ifacesLst := make([]libvirt.VirDomainInterface, 0)`?

inercia: In fact, I think you wouldn't need the `make` at all as you are `append`ing, right?

libvirt/resource_libvirt_domain.go
```diff
@@ -996,3 +1048,56 @@ func getDomainInterfaces(domain *libvirt.VirDomain) ([]libvirt.VirDomainInterfac
 
 	return interfaces, nil
 }
+
+func getDomainInterfacesFromNetworks(domain defDomain, meta interface{}) ([]libvirt.VirDomainInterface, error) {
```
inercia: This function is really similar to [`getDomainInterfacesViaQemuAgent`](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/qemu_agent.go#L32). It is a pity they do not have the same function signature...

Shouldn't we invoke it from [inside `getDomainInterfaces`](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_domain.go#L979)? So we could try to get the interfaces with the qemu agent, then use `getDomainInterfacesFromNetworks` if it fails, and then try with `ListAllInterfaceAddresses` as a last resource...

libvirt/resource_libvirt_domain.go
```diff
@@ -699,6 +699,25 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
...
+		interfacesIfaces[virDom.Name] = true
+	}
+
+	if len(interfacesXML) > len(interfacesIfaces) {
+		ifacesWithAddr, err = getDomainInterfacesFromNetworks(domainDef, meta)
+		if err != nil {
+			return fmt.Errorf("Error retrieving interface addresses: %s", err)
+		}
+	}
+
```
inercia: Maybe we could move this whole block [inside `getDomainInterfaces`](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_domain.go#L979)...

libvirt/resource_libvirt_domain.go
```diff
@@ -858,6 +877,39 @@ func resourceLibvirtDomainDelete(d *schema.ResourceData, meta interface{}) error
...
+		addresses := d.Get(prefix + ".addresses")
+		for _, addressI := range addresses.([]interface{}) {
+			address := addressI.(string)
+			ip := net.ParseIP(address)
+			if ip == nil {
+				return fmt.Errorf("Could not parse addresses '%s'", address)
+			}
+			log.Printf("[INFO] Removing IP/MAC/host=%s/%s/%s in '%s' network", ip.String(), mac, hostname, networkName)
+			if err := removeHost(&network, ip.String(), mac, hostname); err != nil {
+				return err
```
inercia: So do we have to `removeHost` every single address? This is a pain...

eamonnotoole:
@inercia RE: removing networks.  I wasn't clear enough in my description.  The problem is when you do the following:

- Create a virtual network and a domain that attaches to that virtual network while specifying dhcp addresses
- Remove the domain but keep the virtual network intact.
- Recreate the domain while specifying the same dhcp addresses as before

In this case currently the domain recreation will fail because the old dhcp entries are still present in the virtual network definition.  Note that we haven't destroyed the network; if we destroy the network at the same time as the domain then there is no issue on recreation.

The first part of the patch removed the dhcp entries from the network when a domain is destroyed to avoid the above error.

The second part of the patch attempts to fix an issue whereby on multiple domain destroy/recreate cycles while retaining the virtual network(s) - after perhaps two of those cycles - we no longer get a complete list of network addresses from domain. ListAllInterfaceAddresses.  In fact, the list is often empty.  In this case the first part of the patch fails to remove the existing dhcp entries (since the address list is empty or incomplete) and domain recreation again fails.  We can still get the list of dhcp entries directly from the virtual network xml, which is what we do

libvirt/resource_libvirt_domain.go
eamonnotoole: That sounds good, I'll make the change.

inercia: Thanks

inercia:
> In this case currently the domain recreation will fail because the old dhcp entries are still present in the virtual network definition. Note that we haven't destroyed the network; if we destroy the network at the same time as the domain then there is no issue on recreation.

Ok @eamonnotoole , thanks for the clarification. In this case, is the virtual network managed by Terraform? I think it is not, right? Because to destroy a domain while leaving the network is not what users usually do, right?

And, anyway, what is the wrong with having the same DHCP mappings in the network when the domain is recreated? It will get the same IP and that is all, right? And that is not a so bad behaviour... Maybe we should try to be smarter when querying the IP address...

Regarding the second part of the PR, we know `ListAllInterfaceAddresses` is not reliable at all. That's the reason why we extended it for using the qemu agent when available (and we recommend to use it). Any contribution for improving the reliability of `getDomainInterfaces` is appreciated...

libvirt/resource_libvirt_domain.go
eamonnotoole: This section of code is necessary in the case where a domain is destroyed but the network isn't.  I haven't thought of a way of figuring out if that "apply" has been triggered.

libvirt/resource_libvirt_domain.go
eamonnotoole: Done.

eamonnotoole:
This is something that we do extensively - create virtual networks, and then successively destroy and recreate the domains that are attached to those networks.  We use the same IP addresses each time that we recreate the networks.  The problem is that when you recreate the domain with those same addresses you get a libvirt error that the addresses are already taken.  The only way around this that I've come up with is to remove the dhcp entries from the network definitions (the XML) when you destroy the domain.

Is there an alternative?

It sounds like the method to get interface information from the network definitions is generally useful, which is good to know.

inercia:
> This is something that we do extensively - create virtual networks, and then successively destroy and recreate the domains that are attached to those networks. We use the same IP addresses each time that we recreate the networks. The problem is that when you recreate the domain with those same addresses you get a libvirt error that the addresses are already taken. The only way around this that I've come up with is to remove the dhcp entries from the network definitions (the XML) when you destroy the domain.

So you say that _"when you recreate the domain with those same addresses you get a libvirt error that the addresses are already taken"_. Maybe we should be smarter when we get such an error: in that case we should go further and ask _"ok, and who has that IP?"_ and if the MAC address is the address of the interface I'm working on, then we are fine, right?

> It sounds like the method to get interface information from the network definitions is generally useful, which is good to know.

Yeah, any improvement in the method used for getting addresses is more than welcome.

eamonnotoole:
I've changed the logic for the first part of the patch.  Now before I try to add a domain & IP to the list of dhcp entries for a network I first check to see if the domain is already present in this list, and if so I delete it.  I do things this way since the mac address will change on domain recreation if the mac address hasn't been specified in the .tf, and it seemed easier to just delete the existing entry than to mess around with changing the generated mac back to the the value that I can get from the list of dhcp entries in the network.

eamonnotoole:
@inercia What's the story with this PR?  I've heard nothing in over two months.

eamonnotoole:
@inercia @dmacvicar What's the prognosis with this PR?  I've just rebased, and I think that I've addressed Inercia's issues.

MaximilianMeister:
@inercia I've just built an rpm with this patch, and I can confirm that it fixes the issue. thanks @eamonnotoole

kiall:
When giving machines fixed addresses, this PR somehow causes TF to mix up the IPs:

Two machines, with 10.17.3.0 and 10.17.3.1 as their IPs, but TF always connects to .0.

	libvirt_domain.minion.0 (remote-exec): Connecting to remote host via SSH...
	libvirt_domain.minion.0 (remote-exec):   Host: 10.17.3.0
	libvirt_domain.minion.0 (remote-exec):   User: root
	libvirt_domain.minion.0 (remote-exec):   Password: false
	libvirt_domain.minion.0 (remote-exec):   Private key: true
	libvirt_domain.minion.0 (remote-exec):   SSH Agent: false
	libvirt_domain.minion.1 (remote-exec): Connecting to remote host via SSH...
	libvirt_domain.minion.1 (remote-exec):   Host: 10.17.3.0
	libvirt_domain.minion.1 (remote-exec):   User: root
	libvirt_domain.minion.1 (remote-exec):   Password: false
	libvirt_domain.minion.1 (remote-exec):   Private key: true
	libvirt_domain.minion.1 (remote-exec):   SSH Agent: false

eamonnotoole:
Kiall, I'll have a look at those IP issues next week.

eamonnotoole:
@inercia @kiall @MaximilianMeister I've fixed that bug where the TF doesn't use the correct address for multiple domains.  This is what I see now:

```console
libvirt_domain.domain.1 (remote-exec): Connecting to remote host via SSH...
libvirt_domain.domain.1 (remote-exec):   Host: 192.168.0.3
libvirt_domain.domain.1 (remote-exec):   User: root
libvirt_domain.domain.1 (remote-exec):   Password: false
libvirt_domain.domain.1 (remote-exec):   Private key: false
libvirt_domain.domain.1 (remote-exec):   SSH Agent: true
libvirt_domain.domain.0 (remote-exec): Connecting to remote host via SSH...
libvirt_domain.domain.0 (remote-exec):   Host: 192.168.0.2
libvirt_domain.domain.0 (remote-exec):   User: root
libvirt_domain.domain.0 (remote-exec):   Password: false
libvirt_domain.domain.0 (remote-exec):   Private key: false
libvirt_domain.domain.0 (remote-exec):   SSH Agent: true
libvirt_domain.domain.1 (remote-exec): Connecting to remote host via SSH...
libvirt_domain.domain.1 (remote-exec):   Host: 192.168.0.3
libvirt_domain.domain.1 (remote-exec):   User: root
libvirt_domain.domain.1 (remote-exec):   Password: false
libvirt_domain.domain.1 (remote-exec):   Private key: false
libvirt_domain.domain.1 (remote-exec):   SSH Agent: true
libvirt_domain.domain.0 (remote-exec): Connecting to remote host via SSH...
libvirt_domain.domain.0 (remote-exec):   Host: 192.168.0.2
libvirt_domain.domain.0 (remote-exec):   User: root
libvirt_domain.domain.0 (remote-exec):   Password: false
libvirt_domain.domain.0 (remote-exec):   Private key: false
libvirt_domain.domain.0 (remote-exec):   SSH Agent: true
libvirt_domain.domain.1 (remote-exec): Connecting to remote host via SSH...
libvirt_domain.domain.1 (remote-exec):   Host: 192.168.0.3
libvirt_domain.domain.1 (remote-exec):   User: root
libvirt_domain.domain.1 (remote-exec):   Password: false
libvirt_domain.domain.1 (remote-exec):   Private key: false
libvirt_domain.domain.1 (remote-exec):   SSH Agent: true
libvirt_domain.domain.0 (remote-exec): Connecting to remote host via SSH...
libvirt_domain.domain.0 (remote-exec):   Host: 192.168.0.2
libvirt_domain.domain.0 (remote-exec):   User: root
libvirt_domain.domain.0 (remote-exec):   Password: false
libvirt_domain.domain.0 (remote-exec):   Private key: false
libvirt_domain.domain.0 (remote-exec):   SSH Agent: true
```

kiall:
<aside>It's only seeing the GitHub notification emails a few minutes ago that I made the connection you were the same @eamonnotoole from HPE :)</aside>

I'll happily test this out and verify the issue I found is gone, but can't do much more myself (I know just about zero re this code ;))

eamonnotoole:
@kiall Thanks.  Small world.

dmacvicar:
@kiall did you had the chance to test this?

libvirt/resource_libvirt_domain_test.go
```diff
@@ -540,3 +599,73 @@ func testAccCheckLibvirtScsiDisk(n string, domain *libvirt.Domain) resource.Test
...
+		networkDef, err := newDefNetworkfromLibvirt(retrieveNetwork)
+
+		for _, ips := range networkDef.Ips {
+			for _, dhcpHost := range ips.Dhcp.Hosts {
+				if dhcpHost.Ip == ip {
+					return nil
+				}
+			}
+		}
+		return fmt.Errorf("Hostname with ip '%s' does not have a dhcp entry in network", ip)
```
MalloZup: here, we 2 returns, the 2nd one will never called

libvirt/resource_libvirt_domain.go
```diff
@@ -1061,7 +1073,8 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (defDisk, e
 	return disk, nil
 }
 
-func getDomainInterfaces(domain libvirt.Domain) ([]libvirt.DomainInterface, error) {
+func getDomainInterfaces(domain libvirt.Domain, domainDef defDomain,
```
MalloZup: what is `defDomain`?

MalloZup:
afaik we cannot anymore merge this patch as it is.

Resolving the conflicts, we get then afterwards:

```console
go build
# github.com/dmacvicar/terraform-provider-libvirt/libvirt
libvirt/resource_libvirt_domain.go:1127:12: undefined: defDomain
libvirt/resource_libvirt_domain.go:1163:30: undefined: defNetworkInterface
libvirt/resource_libvirt_domain.go:1164:35: undefined: defDomain
libvirt/resource_libvirt_domain.go:1266:59: undefined: defDomain
libvirt/resource_libvirt_domain.go:1305:45: undefined: defDomain
libvirt/resource_libvirt_domain.go:1327:33: networkDef.Ips undefined (type libvirtxml.Network has no field or method Ips, but does have IPs)
libvirt/utils_libvirt.go:49:52: undefined: defNetwork
Makefile:4: recipe for target 'build' failed
```
From my pov, maybe this defNetworks defDomain where existing at the time

eamonnotoole:
@dmacvicar @MalloZup I've rebased the patch, and removed that redundant return.

MalloZup:
looks good to me

MalloZup:
@inercia feel free to say the lastword review to the rebased one pr ,since you @ begin

eamonnotoole:
@MalloZup @inercia @dmacvicar I rebased, it turns out that a recently merged change a20d2a9 negates the need for removing hosts, so that part of the patch is gone.  The patch now implements an alternative way of getting information on domain interfaces.

dmacvicar:
@MaximilianMeister @inercia is this ok to be merged now?

PR #122, opened 2017-04-28
Add configurable timeouts

monstermunchkin:
Terraform supports configurable timeouts since v0.9.0.

With this commit, it's possible to configure the timeout for the
creation of a libvirt domain. It defaults to 5 minutes, but can be as
short as 10 seconds.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -13,8 +13,14 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	libvirt "github.com/dmacvicar/libvirt-go"
+	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
+
+type DomainMeta struct {
+	domain *libvirt.VirDomain
+	ifaces chan defNetworkInterface
+}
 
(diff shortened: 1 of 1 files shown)


flavio:
LGTM

monstermunchkin:
@inercia it's rebased now. Travis fails because of `apt-get`, perhaps you could restart the test.

libvirt/resource_libvirt_domain.go
```diff
@@ -825,72 +848,102 @@ func resourceLibvirtDomainDelete(d *schema.ResourceData, meta interface{}) error
...
+		stateStr = "paused"
+	case libvirt.VIR_DOMAIN_SHUTDOWN:
+		stateStr = "shutdown"
+	case libvirt.VIR_DOMAIN_CRASHED:
+		stateStr = "crashed"
+	case libvirt.VIR_DOMAIN_PMSUSPENDED:
+		stateStr = "pmsuspended"
+	case libvirt.VIR_DOMAIN_SHUTOFF:
+		stateStr = "shutoff"
+	}
```
flavio: I would prefer to have also a default state to handle the ones that might be added in the future, something like:

```golang
default:
  stateStr = fmt.Sprintf("unknown - %v", state[0])
```

PR #128, opened 2017-05-26
Use file type definition for disks

### LINKED ISSUE #126: Using disk type file instead of type volume would fix apparmor permissions
#  Version Reports:

### Distro  version of host:

```Ubuntu 16.04 ```

### Terraform Version Report
Terraform v0.8.8

### Libvirt version

```1.3.1```

```
commit a125a6cf000c9fb3300d2d80afcfb1b13b2952b3
Merge: 785ae79 339552a
Author: Alvaro <alvaro.saurin@gmail.com>
Date:   Mon Feb 20 11:44:33 2017 +0100

    Merge pull request #101 from rsokolkov/fix-docs-for-network
    
    [Docs] Clearer description for "addresses" param
```
___

Use of disk type volume in the domain XML definition prevents virt-aa-helper from being able to determine the correct file path that should be added to the auto generated apparmor profile. This leads to the requirement to disable apparmor for libvirt which is not always possible in every environment (e.g. no sudo access).

Consequently you'll receive the following message when trying to boot without first setting `security = "none"` in /etc/libvirt/qemu.conf:
```
Error applying plan:

1 error(s) occurred:

* libvirt_domain.bootserver: Error creating libvirt domain: [Code-1] [Domain-10] internal error: process exited while connecting to monitor: 2017-05-18T10:15:33.850493Z qemu-system-x86_64: -drive file=/var/lib/libvirt/images/volume-test,format=qcow2,if=none,id=drive-virtio-disk0: Could not open '/var/lib/libvirt/images/volume-test': Permission denied

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

Looking at the domain XML generated and comparing it to a known working vagrant VM that was working with apparmour enabled highlighted the following differences about the disk definition:

terraform libvirt provider generated disk XML from domain
```
    <disk type='volume' device='disk'>
      <driver name='qemu' type='qcow2'/>
      <source pool='default' volume='volume-test'/>
      <target dev='vda' bus='virtio'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x0c' function='0x0'/>
    </disk>
```

vagrant libvirt provider generated disk XML from domain
```
    <disk type='file' device='disk'>
      <driver name='qemu' type='qcow2'/>
      <source file='/var/lib/libvirt/images/vagrant-libvirt_default.img'/>
      <target dev='vda' bus='sata'/>
      <address type='drive' controller='0' bus='0' target='0' unit='0'/>
    </disk>
```

Using this information then lead to the following issue:
https://bugs.launchpad.net/ubuntu/+source/libvirt/+bug/1343245

While it might be possible for future releases on ubuntu to include a version of virt-aa-helper that will know to look at the volume and pool information in order to resolve the disk image path and add that to the apparmor rules, it seems like it might be useful for this provider to switch the XML generated for the disks used to use a format that is better understood by virt-aa-helper and thus remove the need to disable security on qemu when using.

___
# Additional Infos:

Apparmor was enabled in order to debug why the provider didn't work with it, when the vagrant-libvirt provider works.

### PR DESCRIPTION
**Pull request by electrofelix**
Using file type definition for disks allows virt-aa-helper to identify
the backing file correctly from the generated XML and add the necessary
permissions to permit qemu to be able to access the storage disk
provided it is located within a storage pool.

Ensure that reading of existing tfstate using pool/volume definition
remains working for upgrade compatibility by checking first if `File` is
a non-zero string.

Fixes #126

### DIFF
--- a/libvirt/disk_def.go
+++ b/libvirt/disk_def.go
@@ -17,8 +17,10 @@ type defDisk struct {
 		Type string `xml:"type,attr"`
 	} `xml:"format"`
 	Source struct {
-		Pool   string `xml:"pool,attr"`
-		Volume string `xml:"volume,attr"`
+		File   string `xml:"file,attr,omitempty"`
+		// retain Pool/Volume for compatibility with existing tfstate
+		Pool   string `xml:"pool,attr,omitempty"`
+		Volume string `xml:"volume,attr,omitempty"`
 	} `xml:"source"`
 	Target struct {
 		Dev string `xml:"dev,attr"`
@@ -32,7 +34,7 @@ type defDisk struct {
 
 func newDefDisk() defDisk {
 	disk := defDisk{}
-	disk.Type = "volume"
+	disk.Type = "file"
 	disk.Device = "disk"
 	disk.Format.Type = "qcow2"
 	disk.Target.Bus = "virtio"
@@ -45,7 +47,7 @@ func newDefDisk() defDisk {
 
 func newCDROM() defDisk {
 	disk := defDisk{}
-	disk.Type = "volume"
+	disk.Type = "file"
 	disk.Device = "cdrom"
 	disk.Target.Dev = "hda"
 	disk.Target.Bus = "ide"
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -260,21 +260,12 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		if err != nil {
 			return fmt.Errorf("Can't retrieve volume %s", volumeKey)
 		}
-		diskVolumeName, err := diskVolume.GetName()
+		diskVolumeFile, err := diskVolume.GetPath()
 		if err != nil {
-			return fmt.Errorf("Error retrieving volume name: %s", err)
-		}
-		diskPool, err := diskVolume.LookupPoolByVolume()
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by inercia** [comment]
@electrofelix what would be the consequence of using  `<source file='XXX'/>` instead of `<source pool='default' volume='XXX'/>`? Do they need to be mutually exclusive?

**Comment by inercia** [comment]
@electrofelix I mean that in your PR it seems that you are changing the use of `pool/volume` by just file paths, while some users would prefer to continue using `pool/volume`. Maybe you could modify your code so users could use either `pool/volume` or `file`.

**Comment by electrofelix** [comment]
I think i see where some confusion is occurring, the intention here is not to change how users define the volume resource for the provider, rather just the XML emitted. The resource declaration remains as:
```
resource "libvirt_volume" "vol" {
  name   = "vol"
  pool   = "default"
  volume = "something"
}
```
It's only the XML emitted that changes to allow libvirt's virt-aa-helper to work out what apparmor rules to add.

So the compatibility is around ensuring that existing state files remain working so as not to throw an error, but if you were to destroy and recreate the same resources, or create new ones, the XML generated for libvirt would be altered to use the `<source file='...'/>` tag instead.

flavio:
The code looks good to me, however can you please rebase it against master? This should allow travis to run properly

PR #134, opened 2017-06-16
use official libvirt-go bindings

monstermunchkin:


flavio:
Kudos, this looks good to me. Plus I tested some terraform files I have locally and they worked fine.

PR #137, opened 2017-06-19
Support for raw volumes

jeduardo:
When trying to use the provider I found out that it was hardcoded to default to always create and use qcow2 volumes. 

While I can understand why this default was picked up, I am faced with an use case where I actually want to use raw volumes. For this reason I amended the code to support the `format` and `base_volume_format` as optional, configurable items for new volumes.

I've also added one example about how to use it, which is actually the one I used to test this modification.

Could you please give it a look and tell me what you think?

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -35,6 +35,11 @@ func volumeCommonSchema() map[string]*schema.Schema {
 			Computed: true,
 			ForceNew: true,
 		},
+		"format": &schema.Schema{
+			Type:     schema.TypeString,
+			Optional: true,
+			ForceNew: true,
+		},
 		"base_volume_id": &schema.Schema{
 			Type:     schema.TypeString,
 			Optional: true,
@@ -46,6 +51,11 @@ func volumeCommonSchema() map[string]*schema.Schema {
 			ForceNew: true,
 		},
 		"base_volume_name": &schema.Schema{
+			Type:     schema.TypeString,
+			Optional: true,
+			ForceNew: true,
+		},
+		"base_volume_format": &schema.Schema{
 			Type:     schema.TypeString,
 			Optional: true,
 			ForceNew: true,
@@ -105,6 +115,12 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 	if name, ok := d.GetOk("name"); ok {
 		volumeDef.Name = name.(string)
 	}
+
+	volumeFormat := "qcow2"
+	if _, ok := d.GetOk("format"); ok {
+		volumeFormat = d.Get("format").(string)
+	}
+	volumeDef.Target.Format.Type = volumeFormat
 
 	var (
 		img    image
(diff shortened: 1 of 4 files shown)


flavio:
I have to spend more time reviewing the code, but so far it looks good. I also like the fact you provided an example. We should add more of them...

I've just one question, I've noticed you introduced the `base_volume_format` field:
```hcl
resource "libvirt_volume" "vol-debian8-raw" {
  name = "vol-debian8-raw"
  base_volume_id = "${libvirt_volume.debian8-raw.id}"
  base_volume_format = "raw"
}
```

I find it a bit redundant. Isn't there a way to automatically figure out the format of the base volume without bothering the user with that?

jeduardo:
@flavio thanks for the review. You are right, it is a bit redundant to force the user to specify the base volume.

The documentation for [virStoragePoolListAllVolumes](http://libvirt.org/html/libvirt-libvirt-storage.html#virStoragePoolListAllVolumes) (mirrored by the [golang binding](https://godoc.org/github.com/libvirt/libvirt-go#StoragePool.ListAllStorageVolumes)) specifies that it returns only [virStorageVolPtr](http://libvirt.org/html/libvirt-libvirt-storage.html#virStorageVolPtr)s, which do not contain any information about the actual format. 

The contents of virStorageVolPtr are [actually not made public by libvirt itself](http://libvirt.org/html/libvirt-libvirt-storage.html#virStorageVolInfo) and the only info that be recovered regarding the volume is inside the [StorageVolInfo](https://godoc.org/github.com/libvirt/libvirt-go#StorageVolInfo) structure, which, much like the struct [it mirrors from libvirt](http://libvirt.org/html/libvirt-libvirt-storage.html#virStorageVolInfo), exposes only the minimum information about the volume, not exposing any other info such as format. 

The only solution seems to lie in the [actual XML description](http://libvirt.org/html/libvirt-libvirt-storage.html#virStoragePoolGetXMLDesc) of the volume, which is also [retrievable through the golang binding](https://godoc.org/github.com/libvirt/libvirt-go#StorageVol.GetXMLDesc). From there we can fetch the `<format/>` tag from the XML description and retrieve the volume name.

Digging through your code I found out that you already had utility methods to parse XML definitions for volumes, so I just used these methods to implement an alternate approach which reads the XML definition from the base volume, parses them, and fills up the base volume format for the backing store. 

The CI build failed for some internal reason (apt-get returning 100 whilst provisioning go into the build machine), but the local tests worked as expected. 

Could you please give it another look?

(Also: there is a [libvirt project offering standard XML mappings](https://github.com/libvirt/libvirt-go-xml) for libvirt API responses to be used with golang as being forced to parse the resulting XML [seems to be a common occurrence in so many applications](https://www.berrange.com/posts/2017/01/05/announce-new-libvirt-project-go-xml-parser-model/). Maybe it would be useful to use this one instead of replicating the XML structs inside the plugin code?)

inercia:
I agree with @jeduardo: it would be good to use that library for doing all the marshalling/unmarshalling of lilbvirt's XML. I have filed #143 for keeping track of that improvement.

libvirt/resource_libvirt_volume.go
```diff
@@ -210,6 +229,16 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
...
+		backingStoreXml, err := baseVolume.GetXMLDesc(0)
+		if err != nil {
+			return fmt.Errorf("can't get definitions for base image '%s'", baseVolumeName)
+		}
+		backingStore, err := newDefVolumeFromXML(backingStoreXml)
+		if err != nil {
+			return fmt.Errorf("can't parse definitions for base image '%s'", baseVolumeName)
+		}
+		log.Printf("[DEBUG] base volume format is %s", backingStore.Target.Format.Type)
+		volumeDef.BackingStore.Format.Type = backingStore.Target.Format.Type
```
flavio: This is a repetition of the block you introduced above. Can you DRY the code by creating a dedicated function?

flavio:
Thanks for all the changes you made, I'm going to merge it.

jeduardo:
Hey @flavio, no problem, glad to help. Thanks for merging!

PR #140, opened 2017-06-24
Add autostart option for domain resource.

localghost:


flavio:
Are you sure this is actually working? You haven't update the `domainDef` struct.

localghost:
> Are you sure this is actually working?

I have verified this with `virsh dominfo` and by restarting host machine.


> You haven't update the `domainDef` struct.

If I understand [domain specification](http://libvirt.org/formatdomain.html) and the code correctly, `domainDef` should contain fields that will be translated to XML describing the domain but `autostart` is not part of that XML (it is not mentioned in the XML schema) so I don't think it should be put there (it might even result in error I suspect). I am not sure how it is managed internally by KVM, where it stores this information.
Additionally, I checked the XML (via `virsh dumpxml`) after toggling the domain autostart via `virt-manager` a couple of times and it had no influence on the domain's XML.

flavio:
You are right, I can confirm that too.

flavio:
Can you fix the merge conflicts of your PR? Then I'll merge it

localghost:
@flavio, done

PR #144, opened 2017-07-06
Support for NVRAM template

jeduardo:
Hey all,

When working on deploying some UEFI VMs I stumbled upon the problem that the boot entry was not being carried over to the newly provisioned VMs. Thanks to this, every single new UEFI VM was stuck at the UEFI shell after boot instead of booting the system. 

A solution for this would be to use the [NVRAM template attribute](https://libvirt.org/formatdomain.html#elementsOSBIOS) offered by libvirt, allowing a preconfigured set of UEFI variables to be carried over to the new domain. 

What this pull request does is to add such support to this plugin together with an example of how to use it.

Could you please review it and tell me what you think?

### DIFF
--- a/examples/uefi/libvirt.tf
+++ b/examples/uefi/libvirt.tf
@@ -1,0 +1,38 @@
+provider "libvirt" {
+    uri = "qemu:///system"
+}
+
+resource "libvirt_volume" "ubuntu-cloud-uefi" {
+  name = "ubuntu-cloud-uefi"
+  source = "https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64-uefi1.img"
+}
+
+resource "libvirt_volume" "volume" {
+  name = "vm${count.index}"
+  base_volume_id = "${libvirt_volume.ubuntu-cloud-uefi.id}"
+  count = 1
+}
+
+resource "libvirt_domain" "domain" {
+  count = 2
+  name = "ubuntu-cloud-${count.index}"
+  memory = "512"
+  vcpu = 1
+  # This file is usually present as part of the ovmf firmware package in many
+  # Linux distributions.
+  firmware = "/usr/share/OVMF/OVMF_CODE.fd"
+  nvram {
+    # This is the file which will back the UEFI NVRAM content.
+    file = "/var/lib/libvirt/qemu/nvram/vm${count.index}_VARS.fd"
+    # This file needs to be provided by the user.
+    template = "/srv/provisioning/terraform/debian-stable-uefi_VARS.fd"
+  }
+  disk {
+    volume_id = "${element(libvirt_volume.volume.*.id, count.index)}"
+  }
+  graphics {
+    type = "spice"
+    listen_type = "address"
+    autoport = true
+  }
+
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -3,6 +3,7 @@ package libvirt
 import (
 	"encoding/xml"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


examples/uefi/libvirt.tf
```diff
@@ -0,0 +1,43 @@
+provider "libvirt" {
+    uri = "qemu:///system"
+}
+
+resource "libvirt_network" "dev-cluster" {
+   name = "dev-cluster"
+   domain = "dev-cluster.local"
+   mode = "nat"
+   addresses = ["10.0.200.0/24"]
+}
```
flavio: First of all thanks for having provided an example, this is really appreciated! I would prefer to remove the parts of the `.tf` file that are not related with the main topic of this example: uefi and nvram.

Could you remove the custom network please?

examples/uefi/libvirt.tf
```diff
@@ -0,0 +1,43 @@
...
+
+resource "libvirt_volume" "debian9-base" {
+  name = "debian9-base"
+  source = "http://localhost:8000/debian9-uefi.qcow2"
```
flavio: Can you replace that with a link to an image that can be downloaded also by other users?

examples/uefi/libvirt.tf
```diff
@@ -0,0 +1,43 @@
...
+  name = "vm${count.index}"
+  base_volume_id = "${libvirt_volume.debian9-base.id}"
+  count = 2
+}
+
+resource "libvirt_domain" "domain" {
+  name = "vm${count.index}"
+  memory = "256"
+  vcpu = 1
+  nvram = "/var/lib/libvirt/qemu/nvram/vm${count.index}_VARS.fd"
```
flavio: Can you provide this file inside of the examples folder?

examples/uefi/libvirt.tf
```diff
@@ -0,0 +1,43 @@
...
+  nvram_template = "/srv/provisioning/terraform/debian-stable-uefi_VARS.fd"
```
flavio: I guess that's a binary file, right? Otherwise it would be nice to have that into the examples folder too.

examples/uefi/libvirt.tf
```diff
@@ -0,0 +1,43 @@
...
+  firmware = "/usr/share/OVMF/OVMF_CODE.fd"
+  network_interface {
+    network_name = "dev-cluster"
+  }
```
flavio: This can be dropped to simplify the example

examples/uefi/libvirt.tf
```diff
@@ -0,0 +1,43 @@
...
+  disk {
+    volume_id = "${element(libvirt_volume.volume.*.id, count.index)}"
+  }
+  graphics {
+    type = "spice"
+    listen_type = "address"
+    autoport = true
+  }
+  count = 2
```
flavio: This is just personal taste, but I would move it close to the top of the domain definition.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -540,3 +542,91 @@ func testAccCheckLibvirtScsiDisk(n string, domain *libvirt.Domain) resource.Test
 		return nil
 	}
 }
+
+func TestAccLibvirtDomain_FirmwareNoTemplate(t *testing.T) {
+	// var nvram_path = "/var/lib/libvirt/qemu/nvram/terraform-test-firmware-no-template_VARS.fd"
```
flavio: Can you remove this leftover?

libvirt/resource_libvirt_domain_test.go
```diff
@@ -540,3 +542,91 @@ func testAccCheckLibvirtScsiDisk(n string, domain *libvirt.Domain) resource.Test
...
+	var domain libvirt.Domain
+	var config = fmt.Sprintf(`
+            resource "libvirt_domain" "acceptance-test-domain" {
+                    name = "terraform-test-firmware-no-template"
+                    nvram = "%s"
+  	                firmware = "/usr/share/ovmf/OVMF.fd"
+            }`, nvram_path)
+	// size of an accepted, valid, nvram backing store
+	nvram_dummy := make([]byte, 131072)
+	err := ioutil.WriteFile(nvram_path, nvram_dummy, 0777)
```
flavio: Can you refactor this code to use [io/ioutil/TempFile](https://golang.org/pkg/io/ioutil/#TempFile)?

libvirt/resource_libvirt_domain_test.go
```diff
@@ -540,3 +542,91 @@ func testAccCheckLibvirtScsiDisk(n string, domain *libvirt.Domain) resource.Test
...
+	var config = fmt.Sprintf(`
+            resource "libvirt_domain" "acceptance-test-domain" {
+                    name = "terraform-test-firmware-with-template"
+                    nvram = "%s"
+                    nvram_template = "/usr/share/qemu/OVMF.fd"
+                    firmware = "/usr/share/ovmf/OVMF.fd"
+            }`, nvram_path)
+	// size of an accepted, valid, nvram backing store
+	nvram_dummy := make([]byte, 131072)
+	err := ioutil.WriteFile(nvram_path, nvram_dummy, 0777)
```
flavio: Same comment here about using https://golang.org/pkg/io/ioutil/#TempFile

examples/uefi/libvirt.tf
jeduardo: Sure, not a problem, I will remove it. You are right, it it better to focus on what we really want to demonstrate here.

examples/uefi/libvirt.tf
jeduardo: I will try to find one. For now I have been only using images customized to my needs. From a quick search I could find that there is a Ubuntu Cloud image that might work here. If there is any particular image you want me to point to, I can also do it.

examples/uefi/libvirt.tf
jeduardo: So, this one is a binary file and is actually the destination (backing storage) of the UEFI memory after the domain is deployed. 

I am creating it manually on the tests because the libvirt version installed on the Travis CI build image is not doing it. On the libvirt version I'm using on my test systems (Debian 9 with a backported libvirt 3.0.0) this file is created automatically from the template.

examples/uefi/libvirt.tf
jeduardo: Precisely, the template is also a binary file. It's basically the same snapshot of backing memory for the UEFI firmware that should be copied and used by the new domain after it is defined. 

On the test system I have, I have installed the sample UEFI domain (Debian) through virt-install. A new UEFI backing file for the NVRAM was created with the boot entry for the UEFI-compatible GRUB installed. This is the file I'm using as a template for new installation as the EFI partition will always be in the same location as all new domains are copies of the original installed domain.

examples/uefi/libvirt.tf
jeduardo: No problem, will do.

examples/uefi/libvirt.tf
jeduardo: No problems in doing that. I found it a bit weird to have it down there, but then again that's how it was done on other examples, so I imagine it was a convention or something...

libvirt/resource_libvirt_domain_test.go
jeduardo: Definitely, hate when these things get through. :)

libvirt/resource_libvirt_domain_test.go
jeduardo: Definitely. Will do it and revert back.

examples/uefi/libvirt.tf
flavio: If you can find an image that can be downloaded by others fine, otherwise just replace it with `debian-uefi.qcow2`. Simply drop the `http://` part

examples/uefi/libvirt.tf
flavio: I see, no need to have changes then.

examples/uefi/libvirt.tf
flavio: I see, no need to have changes then.

jeduardo:
Hey @flavio! It took me a while but I've found some time to look into the fixes we discussed. Could you please give it another pass? Thanks!

monstermunchkin:
Hey @jeduardo! Your PR looks good. There are however some conflicting files because PR #148 was merged in the meantime. Could you please resolve those conflicts? Thanks for your good work :)

inercia:
Hi @jeduardo . Are you sure this use case is not already covered by the [`firmware` and `nvram` parameters](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/docs/providers/libvirt/r/domain.html.markdown) in a domain?

examples/uefi/libvirt.tf
```diff
@@ -0,0 +1,36 @@
...
+
+resource "libvirt_domain" "domain" {
+  count = 2
+  name = "ubuntu-cloud-${count.index}"
+  memory = "512"
+  vcpu = 1
+  # This is the file which will back the UEFI NVRAM content.
+  nvram = "/var/lib/libvirt/qemu/nvram/vm${count.index}_VARS.fd"
+  # This file needs to be provided by the user.
+  nvram_template = "/srv/provisioning/terraform/debian-stable-uefi_VARS.fd"
```
inercia: Maybe we could surround these parameters in a common group. Something like:

```hcl
  #...
  nvram {
    file = "/var/lib/libvirt/qemu/nvram/vm${count.index}_VARS.fd"
    template = "/srv/provisioning/terraform/debian-stable-uefi_VARS.fd"
  }
  #...
 }
```

Maybe the template could be provided as a real Terraform template (specially since it depends on the VM):

```hcl
data "template_file" "nvram_contents" {
  template = "${file("nvram/nvram.vars.tpl")}"

  vars {
    some_var = "1111"
  }
}

  #...
  nvram {
    file = "/var/lib/libvirt/qemu/nvram/vm${count.index}_VARS.fd"
    template_contents = "${data.template_file.nvram_contents.rendered}"
  }
  #...
 }
```

jeduardo: Updating the documentation page is actually a very good idea, I'm a bit disappointed I forgot about this. :) 

I also like the idea of encapsulating this under the nvram block, honestly this is how I tried to use the nvram block the first time only to find out it was not a block. Definitely gonna do this.

Finally, you are right, the template is a binary file and it really doesn't sense to template its contents in my point of view. I wish it was possible, for it would be much more convenient when provisioning these UEFI VMs.

jeduardo:
Hey all, I think everything should be there now, including the documentation updates. 

I ended up fixing a couple bugs I stumbled upon in the way and both the local and Travis tests seem to be happy now. Another pass?

docs/providers/libvirt/r/domain.html.markdown
```diff
@@ -83,11 +83,14 @@ Some extra arguments are also provided for using UEFI images:
...
-the domain. However, `libvirt` can manage this automatically (and this is the recommended solution)
-if a mapping for the firmware to a _variables file_ exists in `/etc/libvirt/qemu.conf:nvram`.
-In that case, `libvirt` will copy that variables file into a file specific for this domain.
+* `nvram` - (Optional) this block allows specifying the following attributes related to the _nvram_:
+  * `file` - path to the file backing the NVRAM store for non-volatile variables. When provided, 
+  this file must be writable and specific to this domain, as it will be updated when running the 
+  domain. However, `libvirt` can  manage this automatically (and this is the recommended solution) 
+  if a mapping for the firmware to a _variables file_ exists in `/etc/libvirt/qemu.conf:nvram`.
+  In that case, `libvirt` will copy that variables file into a file specific for this domain.
+  * `template` - (Optional) path to the a file used to override variables from the master NVRAM
```
monstermunchkin: Either "path to _the_ file" or "path to _a_ file".

jeduardo: Oops. This is what reviews are for. Fixed. :)

monstermunchkin:
If you could fix the conflict with `domain.html.markdown` and squash your commits, we can merge it.

jeduardo:
@monstermunchkin aye, conflict fix and commits squashed, hope it looks better now. :)

PR #146, opened 2017-07-14
Fix crash occuring when domain is in state file but doesn't exist in libvirt.

localghost:
This commit fixes crash that occurs if a domain resource is present in Terraform's state file but does not exist in libvirt.

One problem with this change is that I am not sure how to add exposing/regression test for this bug. Could anyone briefly explain how I could add such test?


flavio:
Unfortunately right now there's no easy way to write that test. We would have to convert the `virConn` to an interface and mock the calls. That's something on our TODO list.

The change is limited and looks good to me. I'm going to accept it.

PR #147, opened 2017-07-20
use libvirt-go-xml

monstermunchkin:


monstermunchkin:
Some tests are failing with `operation failed: domain 'terraform-test' already exists with uuid ...` which is not a problem with the tests themselves, but with a failed cleanup of domains etc. Can we somehow clean up the domains?

libvirt/domain_def.go
```diff
@@ -1,216 +1,76 @@
...
 
-	domainDef.Devices.Rng.Model = "virtio"
-	domainDef.Devices.Rng.Backend.Model = "random"
+	rng := libvirtxml.DomainRNG{
+		Model: "virtio",
+		Backend: &libvirtxml.DomainRNGBackend{
+			Model: "random",
+		},
+	}
+	domainDef.Devices.RNGs = append(domainDef.Devices.RNGs, rng)
```
inercia: Couldn't we integrate these definitions in the previous `domainDef` definition? Something like:

```Go
	domainDef := libvirtxml.Domain{
		OS: &libvirtxml.DomainOS{
			Type: &libvirtxml.DomainOSType{
				Type: "hvm",
			},
		},
		Memory: &libvirtxml.DomainMemory{
			Unit:  "MiB",
			Value: 512,
		},
		VCPU: &libvirtxml.DomainVCPU{
			Placement: "static",
			Value:     1,
		},
		CPU:     &libvirtxml.DomainCPU{},
		Devices: &libvirtxml.DomainDeviceList{
                                libvirtxml.DomainChannel{
                		                Type: "unix",
                                		Source: &libvirtxml.DomainChardevSource{
                                			Mode: "bind",
                		                },
                                		Target: &libvirtxml.DomainChannelTarget{
		                                	Type: "virtio",
                                			Name: "org.qemu.guest_agent.0",
                		                },
                                	}
                },
		Features: &libvirtxml.DomainFeatureList{
			PAE:  &libvirtxml.DomainFeature{},
			ACPI: &libvirtxml.DomainFeature{},
			APIC: &libvirtxml.DomainFeatureAPIC{},
		},
	}
```

(and the same for libvirtxml.DomainRNG)

libvirt/resource_libvirt_domain.go
```diff
@@ -602,26 +659,6 @@ func resourceLibvirtDomainUpdate(d *schema.ResourceData, meta interface{}) error
 
 	d.Partial(true)
 
-	if d.HasChange("metadata") {
```
inercia: Don't we need this `metadata` stuff?

libvirt/domain_def.go
```diff
@@ -1,216 +1,76 @@
...
+		VCPU: &libvirtxml.DomainVCPU{
+			Placement: "static",
+			Value:     1,
+		},
+		CPU:     &libvirtxml.DomainCPU{},
+		Devices: &libvirtxml.DomainDeviceList{},
+		Features: &libvirtxml.DomainFeatureList{
+			PAE:  &libvirtxml.DomainFeature{},
+			ACPI: &libvirtxml.DomainFeature{},
+			APIC: &libvirtxml.DomainFeatureAPIC{},
```
monstermunchkin: We can drop `PAE`, `ACPI` and `APIC`. The empty `Features` struct is necessary otherwise Terraform crashes. The `CPU` struct is also needed otherwise `TestAccLibvirtDomain_Cpu` will fail.

libvirt/domain_def.go
monstermunchkin: You're right. I totally overlooked that.

libvirt/resource_libvirt_domain.go
monstermunchkin: No, `metadata` is not needed. It works perfectly without it.

libvirt/domain_def.go
flavio: AFAIK we added `ACPI` and `APIC` at the beginning of the project because they were needed to properly shut down a domain. Without the it looked like Windows 95's "you can now turn off your computer" :)

monstermunchkin:
Let's keep the features as they are (PAE, ACPI and APIC). I integrated the definitions in `domainDef`.

flavio:
The code looks fine to me, but the tests are broken on travis CI.

I did some digging and this is caused by the old version of libvirt that is used by travis. Unfortunately we need our tests to run inside of a vm, hence travis is stuck with ubuntu trusty. I searched across all the ubuntu PPA to find a libvirt 2.0 package for it without any luck.

Then I tried to run libvirt within a docker container and I failed. I tried to adapt the instructions of an old Red Hat post, no luck. I tried some images I found on the Docker Hub, again no luck.

I then tried to run libvirt within a lxc container using lxd. This seems to be working. It's possible to have a ubuntu 17 container running inside of a 14 VM. That requires a hell of a scripting to automate all the steps, but I reached to a point where I could run our tests.

Unfortunately I faced yet another libvirt issue which is tied to how the network is configured inside of the lxc containers plus [this](https://bugs.launchpad.net/ubuntu/+source/qemu/+bug/1492621) bug.

I'm sick of this madness. I propose to completely ditch Travis CI and use the jenkins instance operated by [ci.opensuse.org](http://ci.opensuse.org). In that way we will run the tests against a simple openSUSE image. 🎆

flavio:
BTW, the tests are failing because the domain xml we are generating has a random number generator with a more recent syntax. This makes libvirt unhappy -> a test fails -> a domain is not cleaned up -> all the other tests are failing.

We could try to make the xml work with libvirt 1.2, but what is the point in doing that? The xml is properly formatted for the latest stable release of libvirt: 2.0. It's just the stupid ubuntu trusty image that is stuck in time...

MalloZup:
https://github.com/MalloZup/gitbot for running jenkins on prs (you need  an user) but we use this already at suse mgr

flavio:
closing, I've merged the changes through https://github.com/dmacvicar/terraform-provider-libvirt/issues/148

PR #157, opened 2017-08-08
fix XML name of DHCP host element

monstermunchkin:
Use `<host/>` instead of `<NetworkDHCPHost/>` as libvirt complains
about the latter.


monstermunchkin:
This was caused by the move to libvirt-go-xml. https://github.com/libvirt/libvirt-go-xml/blob/master/network.go#L74-L79 doesn't have an explicit `XMLName` field and therefore the XML tag defaults to the name of the struct.

PR #165, opened 2017-08-29
Adding support for arch and machine type

deasmi:
This adds support for setting the architecture and machine type
of the created virtual machine by adding two new attributes
machine and arch to the domain definition


deasmi:
Will make changes and resubmit

PR #166, opened 2017-08-29
Adding support for arch and machine type

deasmi:
This adds support for setting the architecture and machine type
of the created virtual machine by adding two new attributes
machine and arch to the domain definition

Resubmitted after @monstermunchkin comments

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -147,6 +147,14 @@ func resourceLibvirtDomain() *schema.Resource {
 				Optional: true,
 				Required: false,
 			},
+			"machine": &schema.Schema{
+			        Type: schema.TypeString,
+				Optional: true,
+			},
+			"arch": &schema.Schema{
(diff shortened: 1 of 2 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -239,6 +247,9 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	domainDef.OS.Type.Arch = d.Get("arch").(string)
+	domainDef.OS.Type.Arch = d.Get("machine").(string)
```
monstermunchkin: This should be `domainDef.OS.Type.Machine`.

deasmi: Done I hope.

Sorry for the confusion, I don't often end up contributing to public projects via GitHub.

monstermunchkin:
Thanks a lot for your contribution! :+1:

PR #168, opened 2017-09-01
domain: Add support for DomainBootDevice.

hexfusion:
Hello, this PR adds support for defining DomainBootDevice. A usecase would be declaring the domain to hd boot by default then fallback to network as shown below.

usage
```
    boot_device {
        dev = [ "hd", "network"]
    }
}
```
Thank you for your consideration.

### DIFF
--- a/website/docs/r/domain.html.markdown
+++ b/website/docs/r/domain.html.markdown
@@ -55,6 +55,7 @@ The following arguments are supported:
 * `machine` - (Optional) The machine type,
   you normally won't need to set this unless you are running on a platform that
   defaults to the wrong machine type for your template 
+* `boot_device` - (Optional) A list of devices (dev) which sets boot order
 
 ### UEFI images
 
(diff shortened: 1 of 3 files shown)


hexfusion:
@monstermunchkin any input on these changes?

website/docs/r/domain.html.markdown
```diff
@@ -55,6 +55,7 @@ The following arguments are supported:
 * `machine` - (Optional) The machine type,
   you normally won't need to set this unless you are running on a platform that
   defaults to the wrong machine type for your template 
+* `boot_device` - (Optional) A list of devices (dev) which sets boot order
```
flavio: Can you add an example (like the one inside of the body of the PR)?

hexfusion:
@flavio I appreciate the notes. Request for documentation resolved via 9788ca1.
PTAL
Thanks!

website/docs/r/domain.html.markdown
hexfusion: Good idea, thanks.

PR #174, opened 2017-09-08
Adds fuller support for arch and machine settings

deasmi:
Allows updates to be tracked
Sets sensible defaults

Due to the way libvirt works if you specify machine types like pc,isapc
version

The only other alternative was to translate pc > pc-i440fx-2.9 in
defDomain which would require a libvirt connection when one wouldn't
be available, this felt very hacky.

### DIFF
--- a/libvirt/utils_libvirt.go
+++ b/libvirt/utils_libvirt.go
@@ -4,8 +4,10 @@ import (
 	"encoding/xml"
 	"log"
 
+	"errors"
+	"fmt"
 	libvirt "github.com/libvirt/libvirt-go"
-	"github.com/libvirt/libvirt-go-xml"
+	libvirtxml "github.com/libvirt/libvirt-go-xml"
 )
 
 func getHostXMLDesc(ip, mac, name string) string {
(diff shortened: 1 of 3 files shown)


libvirt/utils_libvirt.go
```diff
@@ -68,3 +70,70 @@ func getHostArchitecture(virConn *libvirt.Connect) (string, error) {
...
+		Emulator string `xml:"emulator"`
+		Machines []Machine `xml:"machine"`
+	} `xml:"guest>arch"`
+
+}
+
+type Machine struct {
+	Name string `xml:",chardata"`
+	Canonical string `xml:"canonical,attr"`
+}
```
monstermunchkin: These structs are already defined in libvirt-go-xml: [Maschine](https://github.com/libvirt/libvirt-go-xml/blob/master/capabilities.go#L109-L113) and [CapsGuest](https://github.com/libvirt/libvirt-go-xml/blob/master/capabilities.go#L134-L138).

libvirt/utils_libvirt.go
```diff
@@ -68,3 +70,70 @@ func getHostArchitecture(virConn *libvirt.Connect) (string, error) {
...
+
+
+
+
```
monstermunchkin: A single empty line is enough.

deasmi: Thanks, I'll review and update.

deasmi:
@inercia Correct, and it is that final read that fails all the tests when a simplistic d.Get("machine") was used, as it has already changed from pc to pc-i440fx-2.9 and so the plan wasn't clean as changing it back to pc was required.

libvirt/utils_libvirt.go
deasmi: @monstermunchkin  Hopefully a cleaner version now

deasmi:
Please not not merge this, I have rethought approach finding similar problems with defaults for emulator setting as well.

I will submit a combined PR for these that adds as stage of preparing the DomainDef object for the current hosts default as above cpu,memory etc. defaults will change from distro to disto and so defaults won't work.

PR #179, opened 2017-09-10
Full support for arch,machine & emulator settings

deasmi:
Sorry for new PR, couldn't seem to reopen #174 

Add emulator setting support

Completes support from arch and machine when reading current status

Adds a new function newDomainDefForConnection that takes
a domainDef and populates with defaults from the current host
which is more sensible than guessing across distributions

Maps canonical machine types <> actual to avoid changes

Due to the way libvirt works if you specific machine types like pc,isapc
version

### DIFF
--- a/libvirt/domain_def.go
+++ b/libvirt/domain_def.go
@@ -1,9 +1,9 @@
 package libvirt
 
 import (
+	libvirt "github.com/libvirt/libvirt-go"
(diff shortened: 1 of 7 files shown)


deasmi:
Close and reopen as travis not actually getting to the tests, seen this a few times no idea why

PR #180, opened 2017-09-10
Full support for arch,machine & emulator settings

deasmi:
Sorry for new PR, couldn't seem to reopen #174

Add emulator setting support

Completes support from arch and machine when reading current status

Adds a new function newDomainDefForConnection that takes
a domainDef and populates with defaults from the current host
which is more sensible than guessing across distributions

Maps canonical machine types <> actual to avoid changes

Due to the way libvirt works if you specific machine types like pc,isapc
version

### DIFF
--- a/libvirt/domain_def.go
+++ b/libvirt/domain_def.go
@@ -1,9 +1,9 @@
 package libvirt
 
 import (
+	libvirt "github.com/libvirt/libvirt-go"
+	libvirtxml "github.com/libvirt/libvirt-go-xml"
 	"os"
-
-	"github.com/libvirt/libvirt-go-xml"
 )
 
 func newFilesystemDef() libvirtxml.DomainFilesystem {
@@ -20,7 +20,9 @@ func newDomainDef() libvirtxml.Domain {
 	domainDef := libvirtxml.Domain{
 		OS: &libvirtxml.DomainOS{
 			Type: &libvirtxml.DomainOSType{
-				Type: "hvm",
+				Type:    "hvm",
+				Arch:    "x86_64",
+				Machine: "pc",
 			},
 		},
 		Memory: &libvirtxml.DomainMemory{
@@ -74,4 +76,25 @@ func newDomainDef() libvirtxml.Domain {
 	}
 
 	return domainDef
+}
+
+func newDomainDefForConnection(virConn *libvirt.Connect) (libvirtxml.Domain, error) {
+	d := newDomainDef()
+	caps, err := getHostCapabilities(virConn)
+	if err != nil {
+		return d, err
+	}
+	guest, err := getGuestForArchType(caps, d.OS.Type.Arch, d.OS.Type.Type)
+	if err != nil {
+		return d, err
+	}
+
+	d.Devices.Emulator = guest.Arch.Emulator
+
+	canonicalmachine, err := getCanonicalMachineName(caps, d.OS.Type.Arch, d.OS.Type.Type, d.OS.Type.Machine)
+	if err != nil {
+		return d, err
+	}
+	d.OS.Type.Machine = canonicalmachine
+	return d, nil
 }
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -148,12 +148,14 @@ func resourceLibvirtDomain() *schema.Resource {
 				Required: false,
 			},
 			"machine": &schema.Schema{
-			        Type: schema.TypeString,
+				Type:     schema.TypeString,
 				Optional: true,
+				Default:  "pc",
 			},
 			"arch": &schema.Schema{
-			        Type: schema.TypeString,
+				Type:     schema.TypeString,
 				Optional: true,
+				Default:  "x86_64",
 			},
 			"boot_device": &schema.Schema{
 				Type:     schema.TypeList,
@@ -162,6 +164,11 @@ func resourceLibvirtDomain() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: bootDeviceSchema(),
 				},
+			},
+			"emulator": &schema.Schema{
+				Type:     schema.TypeString,
+				Default:  "/usr/bin/qemu-system-x86_64",
+				Optional: true,
 			},
 		},
 	}
@@ -208,8 +215,10 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		return fmt.Errorf("The libvirt connection was nil.")
 	}
 
-	domainDef := newDomainDef()
-
+	domainDef, err := newDomainDefForConnection(virConn)
+	if err != nil {
+		return err
+	}
 	if name, ok := d.GetOk("name"); ok {
 		domainDef.Name = name.(string)
 	}
@@ -270,7 +279,8 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 
 	domainDef.OS.Type.Arch = d.Get("arch").(string)
 	domainDef.OS.Type.Machine = d.Get("machine").(string)
-	
+	domainDef.Devices.Emulator = d.Get("emulator").(string)
+
 	if firmware, ok := d.GetOk("firmware"); ok {
 		firmwareFile := firmware.(string)
 		if _, err := os.Stat(firmwareFile); os.IsNotExist(err) {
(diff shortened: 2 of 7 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -268,9 +277,17 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
...
+		domainDef.OS.Type.Arch = guestarch.(string)
+	}
+	if machine, ok := d.GetOk("machine"); ok {
+
+		domainDef.OS.Type.Machine = machine.(string)
+	}
+	if emulator, ok := d.GetOk("emulator"); ok {
+		domainDef.Devices.Emulator = emulator.(string)
+	}
+
```
monstermunchkin: You can skip the whole `GetOk(...)` construct, and just use `Get(...)`.

libvirt/utils_libvirt_test.go
```diff
@@ -32,3 +34,81 @@ func TestGetHostXMLDesc(t *testing.T) {
 		t.Errorf("expected name %s, got %s", name, dd.Name)
 	}
 }
+
+func connect(t *testing.T) *libvirt.Connect {
+	conn, err := libvirt.NewConnect(os.Getenv("LIBVIRT_DEFAULT_URI"))
+	if err != nil {
+		t.Fatalf("Cannot connect")
```
monstermunchkin: This should include the actual error as well.

travis/setup-guest
```diff
@@ -13,7 +13,7 @@ while check_network ; [ $? -ne 0 ];do
 done
 
 apt-get -qq update
-apt-get install -y qemu libvirt-bin libvirt-dev golang-1.8 ovmf
+apt-get install -y qemu-system-common libvirt-bin libvirt-dev golang-1.8 ovmf
```
monstermunchkin: This is covered in one of your other PRs already.

libvirt/domain_def.go
```diff
@@ -75,3 +77,24 @@ func newDomainDef() libvirtxml.Domain {
...
+
+	d.Devices.Emulator = guest.Arch.Emulator
+
+	canonicalmachine, err := getCanonicalMachineName(caps, d.OS.Type.Arch, d.OS.Type.Type, d.OS.Type.Machine)
+	if err != nil {
+		return d, err
+	}
+	d.OS.Type.Machine = canonicalmachine
+	return d, nil
+}
```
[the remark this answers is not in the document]
deasmi: Felt cleaner code wise, they both feel like factories, newDomainDef gives just that, newDomainDefForConnection gives one customised for a connection.

Would make it impossible for get a domainDef without passing a nil which feels odd for a bacis onbject in the packge.

libvirt/resource_libvirt_domain.go
deasmi: Wouldn't that overwrite the defaults with blanks if there wasn't something set in d ?

We need to keep the defaults from the connection otherwise it will see changes on creation.

libvirt/utils_libvirt_test.go
deasmi: Will fix

travis/setup-guest
deasmi: Yes, I can pull the other PR or remove from here, but travis fails without this.

libvirt/resource_libvirt_domain.go
monstermunchkin: Not according to the [documentation](https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceData.Get):

> If the key does exist in the schema but doesn't exist in the configuration, then the default value for that type will be returned. For strings, this is "", for numbers it is 0, etc.

And since we are setting a `Default` value in the [Schema](https://godoc.org/github.com/hashicorp/terraform/helper/schema#Schema), it won't be blank.

deasmi: But isn't 

` domainDef.Devices.Emulator = Get("emulator").(string)`

effectively going to call .(string) on a nil value (default return from Get for interface{}) which will replace with a blank ?

Sorry if I'm just wrong here, I'm pretty new to go.

monstermunchkin: No, `d.Get("emulator")` will not return nil, I'm sure. It'll be either blank (`""`) or a specified default value (our case). Therefore, `.(string)` will not be called on a `nil` value. If it were `nil`, it would panic (which it doesn't) with:
```
panic: interface conversion: interface {} is nil, not string
```

travis/setup-guest
deasmi: @monstermunchkin shall I remove from this PR or close the other ?

libvirt/resource_libvirt_domain.go
deasmi: @monstermunchkin Changes as suggested and pushed

travis/setup-guest
monstermunchkin: I've closed the other PR, but please also remove this commit from this PR since it's not needed.

deasmi:
Can someone kick off travis again please, it's failing due to timeout on test. Test passes fine when run manually.

flavio:
> Can someone kick off travis again please, it's failing due to timeout on test. Test passes fine when run manually.

I just did it.

deasmi:
@flavio tests failed to complete again, at a later point.

```
=== RUN   TestAccLibvirtDomain_FirmwareNoTemplate
SIGQUIT: quit
PC=0x46e4f1 m=0 sigcode=0
```

Is the timeout a travis setting ? 

As VM within VM within Container, or how every many rabbit holes we've gone down seems destined to be slow.

deasmi:
I am a bit concerned about the test harness. 

As you can see I've added a couple of commits, the first rebased into the original add a test for getHostCapabilites which should have been there anyway.

The second I've left separate as it just increases the test timeout to 1200s from default 600s to get Travis to complete

This is needed as calling getHostCapabilities takes a ridiculous amount of time in the test harness.

```
	utils_libvirt_test.go:132: [DEBUG] Get host capabilites took 3.626146377s
```
compared to

```
	utils_libvirt_test.go:132: [DEBUG] Get host capabilites took 9.831542ms
```

when run locally; quite a difference.

@monstermunchkin @flavio could do with advice on what to do here, there is from my perspective an underlying problem with the libvirtd in the test harness making it extremely slow.

libvirt/domain_def.go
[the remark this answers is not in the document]
deasmi: @monstermunchkin Is this going to be allowed as-is or do you require changes ? 

Feels like a separation of responsibility issues, make a new domainDef vs. configure a domainDef for a host.

But if decision is to combine I can refactor easily enough. 

I'd like to get the PR agreed on way or another.

libvirt/resource_libvirt_domain.go
```diff
@@ -268,9 +277,11 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+
```
monstermunchkin: Unnecessary blank line

libvirt/utils_libvirt_test.go
```diff
@@ -32,3 +35,100 @@ func TestGetHostXMLDesc(t *testing.T) {
...
+	
+	name, err := getCanonicalMachineName(caps, arch, virttype, machine)
+
+	if err != nil {
+		t.Errorf("Could not get canonical name for %s/%s", arch, machine)
+		return
+	}
+
+	t.Logf("Canonical name for %s/%s = %s", arch, machine, name)
+
```
monstermunchkin: Unnecessary blank line

libvirt/utils_libvirt_test.go
```diff
@@ -32,3 +35,100 @@ func TestGetHostXMLDesc(t *testing.T) {
...
+	caps,err := getHostCapabilities(conn)
+	if err != nil {
+		t.Errorf("Can't get host capabilties")
+	}
+	if caps.Host.UUID == "" {
+		t.Errorf("Host has no UUID!")
+	}
+
+
+
```
monstermunchkin: Unnecessary blank lines

libvirt/utils_libvirt_test.go
```diff
@@ -32,3 +35,100 @@ func TestGetHostXMLDesc(t *testing.T) {
...
+	elapsed := time.Since(start)
+	t.Logf("[DEBUG] Get host capabilites took %s", elapsed)
+
```
monstermunchkin: Unnecessary blank line

libvirt/domain_def.go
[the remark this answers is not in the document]
monstermunchkin: This can stay as it is.

deasmi:
@monstermunchkin blank lines removed, thank you for your patience.

deasmi:
Can this be merged, or is there concern about the test delays ?

PR #182, opened 2017-09-12
adds tests for fedora

timhughes:
Was trying to work out if an issue was fedora based. Might be useful for adding other distros as well


timhughes:
Might pay to have the fedora tests allowed to fail until we can work them.out

monstermunchkin:
I think the failing tests need to be worked out before we merge them. Keeping them around doesn't make much sense.
What do you think, @inercia and @flavio?

flavio:
Yes, it would be better to fix the tests. However I wonder I long we can go on with this "LXD Russian dolls" situation...

I would like to give a shot at what was suggested here: https://twitter.com/bradrydzewski/status/890900013664141312

MalloZup:
is not a bad pr . we could take this pr if we make the `fedora` tests optional, so they are not running in travis ( since the test are failing, btw to me the failing issue is now related to the same we have on #222),
 so to me seems that when we fix #222 (which is not LXC relevant) we will fix also this tests

MalloZup:
@timhughes i think this pr can be merged if you implement via make, like `make testacc fedora`or something similar syntax. We plan to move ci outside travis

MalloZup:
@timhughes  thank you for this PR.
we are moving to Jenkins CI pipelines so i close this PR, but i consider it on #334.
If you wish feel free to contribute there or on other issues.
The plan is to have `ci` folder where we have then pipelines inside.

PR #183, opened 2017-09-14
We don't need to force format here

jaroug:
Hi,

The format should be specified on volume creation, this reset the utilization of raw format, and kvm is supposed to be anyway able to detect a disk format if not specified.

Best regards,

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -365,7 +365,6 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 			},
 			Driver: &libvirtxml.DomainDiskDriver{
 				Name: "qemu",
-				Type: "qcow2",
 			},
 		}
 


flavio:
LGTM, but I would prefer to check the code locally before merging it

monstermunchkin:
This doesn't work for me. The disk format is not automatically detected.

MalloZup:
i tested this on debian and worked out. It might be dependent on KVM/libvirt version? 
Imho we can live with an additonal line, if removing it cause break on compatibility with other distros

jaroug:
Well, my patch is shitty and need a proper rework. But forcing also break support of raw device :)

MalloZup:
@jaroug yop true :+1:  imho we need an issue for this instead of pr, and  the usecase is that the support of raw device get broken, so we can discuss it further or to implement a generic solution

monstermunchkin:
A generic solution would be to add a `type` option to `disk`:
```hcl
disk {
  volume_id = "foo"
  type      = "qcow2" # or "raw"
}
```

MalloZup:
this pr will fix that in more generic way https://github.com/dmacvicar/terraform-provider-libvirt/pull/380. 
thx for the PR

PR #184, opened 2017-09-14
Updating libvirt-xml-go from upstream

deasmi:
libvirt-xml-go had a bug that appeared with go1.9

This brings us up to date with master.


monstermunchkin:
I'm aware of this issue. However, we use Glide for vendoring, and this change needs to be reflected in the files used by Glide. I'm closing this in favor of #185 which updates all vendored packages.

PR #187, opened 2017-09-15
Fix for Issue#186

rsjethani:


rsjethani:
Ignore due to unclean commit history

Issue #186, opened 2017-09-15
not waiting for DHCP IP lease even after specifying `wait_for_lease = true`


rsjethani:
#  Version Reports:

### Distro  version of host:

```Ubuntu 16.04 ```

### Terraform Version Report
```Terraform v0.10.2```

### Libvirt version

```1.3.1```

### terraform-provider-libvirt plugin version (git-hash)

``` git log```
___
# Description of Issue/Question

### Setup

```
provider "libvirt" {
    uri = "qemu:///system"
}

resource "libvirt_network" "docker-net" {
  name = "docker-net"
  mode = "nat"
  addresses = ["10.0.0.0/24"]
  running = true
}

# qcow2 image for docker master
resource "libvirt_volume" "docker-master" {
  name = "docker-master.qcow2"
  source = "/var/lib/libvirt/images/ubuntu16.04.qcow2"
}

# create docker-master domain
resource "libvirt_domain" "docker-master" {
  name = "docker-master"
  memory = "512"
  vcpu = 1
  network_interface {
    network_id = "${libvirt_network.docker-net.id}"
    wait_for_lease = true
  }
  disk {
    volume_id = "${libvirt_volume.docker-master.id}"
  }
  graphics {
    type = "spice"
    listen_type = "address"
    autoport = "yes"
  }
}
```

### Steps to Reproduce Issue
$ terraform init
$ terraform plan
$ terrafom apply

___
# Additional Infos:
The same behaviour happens if we specify already existing network via `network_name`
DEBUG LOG: 

```
ravi@gs-3355:~/work/build-infra/docker_swarm_3_nodes$ terraform apply
2017/09/15 16:36:54 [INFO] Terraform version: 0.10.2  a1d06eb97378d097d4175c3284ea026937618477
2017/09/15 16:36:54 [INFO] Go runtime version: go1.8.3
2017/09/15 16:36:54 [INFO] CLI args: []string{"/usr/bin/terraform", "apply"}
2017/09/15 16:36:54 [DEBUG] Attempting to open CLI config file: /home/ravi/.terraformrc
2017/09/15 16:36:54 [INFO] CLI command args: []string{"apply"}
2017/09/15 16:36:54 [INFO] command: empty terraform config, returning nil
2017/09/15 16:36:54 [DEBUG] command: no data state file found for backend config
2017/09/15 16:36:54 [DEBUG] New state was assigned lineage "86192502-3f67-4430-b38a-deb7b212be7e"
[...]

PR #190, opened 2017-09-15
FIX waitForLeases invalid key issue

rsjethani:
We are modifying network interface object after adding it as key to
waitForLeases map. This makes the interface object stored in the key
invalid when it is used in functions like hasNetworkAddress() later.

As a result of this the code does not wait for the interface to get
the lease even after specifying `wait_for_lease = true` in the tf
file.

To fix this we move the code to add interface struct object to
waitForLeases map only after all modifications are done i.e. near
to the end of for loop.

For mode details please see [issue#186](https://github.com/dmacvicar/terraform-provider-libvirt/issues/186).

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -484,12 +484,6 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 			Address: mac,
 		}
 
-		// this is not passed to libvirt, but used by waitForAddress
-		waitForLeases[netIface] = false
-		if waitForLease, ok := d.GetOk(prefix + ".wait_for_lease"); ok {
-			waitForLeases[netIface] = waitForLease.(bool)
-		}
-
 		// connect to the interface to the network... first, look for the network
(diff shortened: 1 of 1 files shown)


rsjethani:
@monstermunchkin merged!!!  
I was actually working on adding test cases so as to improve test coverage

monstermunchkin:
@rsjethani sorry, I wasn't aware of that. Feel free to create another PR with the test cases. Increased coverage is always welcome :)

Thanks for your contribution(s) :+1:

flavio:
> I was actually working on adding test cases so as to improve test coverage

@rsjethani: feel free to provide that with another PR ;)

rsjethani:
> @rsjethani sorry, I wasn't aware of that.

@monstermunchkin hey it's fine.

I am actuaaly stuck writing test case for the changes I added. I need to write test case around "wait_for_lease" functionality. I am still going through testing code but so far it seems the iso/volume used for testing gets stuck when tested for "wait_for_lease = true". Should I create a new issue for discussing this problem or some place else. Thanks :)

PR #193, opened 2017-09-18
code cleanup

monstermunchkin:


deasmi:
This feels like an odd way to approach the, admittedly needed, refactoring.

Why are we adding the Terraforma Schema and Connection info into the domainDef object?
Couldn't we just refactor as functions in resource_libvirt_domain?

If we are then it needs renaming as it's no longer a domainDef but some hybrid object with many functions, which I though OO principals were quite against in general. Can't additional functions just be added to resource_libvirt_domain.go, or perhaps a new struct that contains a domainDef, connection and terraform schema then functions on that ?

This also invalidates a number of open PRs, #117,  #111,  #180 and I doubt these will apply once they are accepted or this will apply once they are, unless the expectation is people go away and start their PRs again.

monstermunchkin:
This is just an idea on how to refactor some of the code. It's still work in progress, and I appreciate every kind of input.

> Why are we adding the Terraforma Schema and Connection info into the domainDef object?

 I just thought it looked neater when all the functions looked similar and didn't take arguments. As already stated, this is just one of many ideas.

> Couldn't we just refactor as functions in resource_libvirt_domain?

The reason I moved/added the functions to `domain_def.go` is that they are part of the domain definition and belong there instead of `resource_libvirt_domain.go`.

> This also invalidates a number of open PRs, #117, #111, #180 and I doubt these will apply once they are accepted or this will apply once they are, unless the expectation is people go away and start their PRs again.

You're right, they won't apply. However, the other PRs will be merged before this one. We don't expect anyone to rewrite their PR due to this.

deasmi:
>> Why are we adding the Terraforma Schema and Connection info into the domainDef object?

> I just thought it looked neater when all the functions looked similar and didn't take arguments. As already stated, this is just one of many ideas.

>> Couldn't we just refactor as functions in resource_libvirt_domain?

> The reason I moved/added the functions to domain_def.go is that they are part of the domain definition and belong there instead of resource_libvirt_domain.go.

I think they are things that happen to the domainDef, rather than being part of it, perhaps it's just semantics. 

I was thinking about a new struct resource_libvirt_domain, containing the three bits of data that define that resource, but then that feels odd as it would be used in very different ways across the Create/Update/Read/Delete functions.


Perhaps something like
```
func (Domain *d) configureForConnectionAndDefinition(virConn *c, resourceLibvirtDomain *rd) {
..
d.setArchFor(c,rd)
..
}

func (resourceLibvirtDomain *rd) configureFromDomainDefAndConnection(Domain *d, virConn *c) {
..
  rd.setArchAndMachineFrom(d,c)
..
}
```
I don't think there is any need to retain the virConn and Domain past initialisation is there for them to need to be in the DomainDef object ?

monstermunchkin:
Please review @flavio @dmacvicar @inercia @MalloZup

MalloZup:
@monstermunchkin  you did it! :cupid:  great ! i will review it asap , enjoy :japanese_castle:

flavio:
LGTM, I fixed the conflicts inside of https://github.com/dmacvicar/terraform-provider-libvirt/pull/281

I'm going to close this PR. Thanks @monstermunchkin for the hard work

Issue #171, opened 2017-09-03
Video option


jfgibbins:
```hcl
# Configure the Libvirt provider
provider "libvirt" {
  uri = "qemu:///system"
}

resource "libvirt_domain" "pfSense" {
  name = "pfSense"
  vcpu = 4
  memory = 4096
  running = true
  autostart = true
  disk {
    volume_id = "${libvirt_volume.pfSense.id}"
  }
  network_interface {
    network_name = "pfSense"
    hostname = "pfSense"
  }
}

resource "libvirt_volume" "pfSense" {
  name = "pfSense"
  source = "/var/lib/libvirt/images/pfSense.qcow2"
}
```

Generates error and will not start, but going into virt-manager, changing video from Cirrus to QXL and container boots successfully.  There was also a device called qemu-ga, but no idea what that even is, deleted as well from config.

hexfusion:
Try defining graphics few options that have worked for me.
https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/domain.html.markdown#graphics-devices
```
  graphics {
    type = "vnc"
    listen_type = "address"
  }
```

```
  graphics {
    type = "spice"
    listen_type = "address"
    autoport = true
  }
```
-HTH

jfgibbins:
the options you listed are for VNS vs.SPICE, but not the actual graphics hardware for the guest, but thanks

monstermunchkin:
Feel free to send a pull request adding the video option.

jfgibbins:
I would, but I'm not a programmer.  Do not pass go, and do not collect $200.

deasmi:
I am working on some vnc additions so I'll see if I can include this. No promises however.

jfgibbins:
@deasmi Always appreciated.  It's on my list to start learning go, but currently working on a lot of "infrastructure as code".  Once I get that done, I might have a few working brain cells left to start learning something new.

deasmi:
Can we add this to 0.1 ?

deasmi:
@jfgibbins Can you test if my patch fixes your issue please.

Branch graphicsandvnc on https://github.com/deasmi/terraform-provider-libvirt.git

Pull request #198

MalloZup:
We use now xslt for allowing custom parameters to domain network etc. Look at the doc or examples. Tia

PR #198, opened 2017-09-19
Fuller support for graphics and video settings

### LINKED ISSUE #197: Provide fuller support for graphics and video
The current support for video and graphics settings is somewhat limited, as referenced in #171 

I am working on adding better support so wanted to add an issue to track this, I will submit a [WIP] PR referencing this for others to look at shortly.

I am aiming to support setting listen addresses and optionally port numbers for VNC plus fix #171 initially.

Ideally a 0.1 item.

deasmi:
# Full support for graphics and video settings

This patch adds full support for vnc and spice graphics types
as well as support for changing the video device.

By default a graphics device will no longer be added, this is due
to older code assuming that the graphics type would never be read or
changed. As two of the architectures the code supports are ppc64 and
s390x and these don't support graphics a default wouldn't work.

VNC support now includes setting the listen port and listen address

Closes #197 and #171


libvirt/domain_def.go
```diff
@@ -33,12 +33,6 @@ func newDomainDef() libvirtxml.Domain {
 		},
 		CPU: &libvirtxml.DomainCPU{},
 		Devices: &libvirtxml.DomainDeviceList{
-			Graphics: []libvirtxml.DomainGraphic{
-				libvirtxml.DomainGraphic{
-					Type:     "spice",
-					AutoPort: "yes",
-				},
-			},
```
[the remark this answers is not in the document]
deasmi: It isn't valid if not on x86_64 or i686, there is specific code for s390x and ppc elsewhere and as we now read back what gets created, unlike previously, if left we'd have a mismatch and plan wouldn't be clean. 

Also there is no need for a graphics block at all by default, and IMO in many cloud situations it's just not needed and an unnecessary overhead. Typically ssh is the only connection ever likely.

libvirt/resource_libvirt_domain.go
```diff
@@ -239,24 +243,67 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
...
+
+			if listenType == "address" {
+				listener := libvirtxml.DomainGraphicListener{}
+				listener.Type = listenType
+				listener.Address = listenAddress
 				domainDef.Devices.Graphics[0].Listeners = []libvirtxml.DomainGraphicListener{
-					libvirtxml.DomainGraphicListener{
-						Type: listen_type.(string),
-					},
+					listener,
```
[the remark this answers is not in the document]
deasmi: Thanks, I'll review all these.

deasmi:
@monstermunchkin Can you take a look at new version please, should be much cleaner.

libvirt/resource_libvirt_domain.go
```diff
@@ -244,28 +248,83 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
...
+				return fmt.Errorf("[ERROR] It does not make sense to set autoport to yes and a listen_port")
+			} else if (graphicsMap["autoport"] == nil && graphicsMap["listen_port"] == nil) ||
+				(graphicsMap["autoport"] != nil && graphicsMap["autoport"].(string) == "yes") {
+				domainDef.Devices.Graphics[0].AutoPort = "yes"
+			} else {
+				log.Printf("[INFO] As port has been specified turning autoport off")
+				domainDef.Devices.Graphics[0].AutoPort = "no"
+				d.Set("graphics.autoport", "no")
+				domainDef.Devices.Graphics[0].Port, _ = strconv.Atoi(graphicsMap["listen_port"].(string))
 			}
```
monstermunchkin: IMO this section looks unnecessary complicated, and doesn't cover all cases. For example, this code will fail if `listen_port != nil` and `autoport == "no"`. This shouldn't be, and there's no test case for it.

 Also, go can [check the existence of keys in maps](https://tour.golang.org/moretypes/22). There's no need to check for `nil`. This should be simpler and cover all cases:
```go
autoPort, autoPortOk := graphicsMap["autoport"]
listenPort, listenPortOk := graphicsMap["listen_port"]

// If both autoport is yes and port are set then we should fail
if autoPort == "yes" && listenPortOk {
	return fmt.Errorf("[ERROR] autoport and listen_port cannot be used at the same time")
}

// If neither are set use autoport
if !autoPortOk && !listenPortOk {
	domainDef.Devices.Graphics[0].AutoPort = "yes"
} else {
	// Otherwise use autoport and/or port as specified
	if autoPortOk {
		domainDef.Devices.Graphics[0].AutoPort = autoPort.(string)
	}

	if listenPortOk {
		domainDef.Devices.Graphics[0].Port, _ = strconv.Atoi(listenPort.(string))
	}
}
```

libvirt/resource_libvirt_domain.go
```diff
@@ -861,6 +920,27 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
...
+	if domainDef.Devices.Graphics != nil {
+		graphics := map[string]string{}
+		graphicsInstance := domainDef.Devices.Graphics[0]
+		graphics["type"] = graphicsInstance.Type
+		// Add either the port or autoport but not both
+		if graphicsInstance.AutoPort == "yes" {
+			graphics["autoport"] = "yes"
+		} else {
+			graphics["listen_port"] = strconv.Itoa(graphicsInstance.Port)
+		}
```
monstermunchkin: This section doesn't cover `no` for `AutoPort`.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -252,10 +253,103 @@ func TestAccLibvirtDomain_Graphics(t *testing.T) {
...
+	var config = fmt.Sprintf(`
+            resource "libvirt_volume" "acceptance-test-graphics" {
+                    name = "terraform-test"
+            }
+
+            resource "libvirt_domain" "acceptance-test-domain" {
+                    name = "terraform-test"
+                    graphics {
+                            type = "spice"
+                            autoport = "true"
```
monstermunchkin: This should be `"yes"`, just like the others.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -252,10 +253,103 @@ func TestAccLibvirtDomain_Graphics(t *testing.T) {
...
+            }`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtDomainDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config:      config,
+				ExpectError: regexp.MustCompile(".*We only support graphics types of spice.*"),
```
monstermunchkin: Please provide the _entire_ error message. Don't be lazy ;)

libvirt/resource_libvirt_domain_test.go
```diff
@@ -252,10 +253,103 @@ func TestAccLibvirtDomain_Graphics(t *testing.T) {
...
+            }`)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtDomainDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config:      config,
+				ExpectError: regexp.MustCompile(".*It does not make sense.*"),
```
monstermunchkin: Same as above, please provide the entire error message.

libvirt/resource_libvirt_domain.go
deasmi: Thanks, that's very helpful. Been staring at it for too long.

deasmi:
I'm a bit stuck with this and have opened an issue with terraform hashicorp/terraform#16188.

Almost certainly my lack of understanding somewhere but I don't understand why setting the video type in the schema works and autoport does not.

libvirt/resource_libvirt_domain.go
```diff
@@ -126,7 +126,12 @@ func resourceLibvirtDomain() *schema.Resource {
 			"graphics": &schema.Schema{
 				Type:     schema.TypeMap,
 				Optional: true,
-				Required: false,
+				Computed: true,
```
monstermunchkin: If you remove this line, the test should succeed. At least they do on my local machine.

monstermunchkin:
@deasmi, are you still working on this?

deasmi:
@monstermunchkin I will but haven't had time of late. I even hit a bug with go (https://github.com/golang/go/issues/21945) that stopped me for a while. 

I was struggling with getting config to be persistent, and may need to rewrite to support multiple vnc consoles. I will pick up again.

deasmi:
I am closing this as it won't apply well against current code plus I have no time to work on this currently.

PR #200, opened 2017-09-20
Remove redundant hostname check

rsjethani:
Currently while creating domain we are checking whether hostname
has been set or not. But from the code it's clear that hostname
will _always_ be set. Hostname will be set to domain name by default
which maybe overridden via the optional hostname attribute of the
network interface definition.

Hence removing the unnecessary check.


monstermunchkin:
Thanks :+1:

PR #202, opened 2017-10-12
Build cleanup and updates to build instructions

pfremm:
Adding vendor packages to gitignore.  So now building properly has users install vendor packages with glide.  Updated build directions for the glide install step and dependency on go 1.9 that terraform now has.


flavio:
We discussed for a long time whether to add all the vendored dependencies to our repo or not. Finally we decided to keep them inside of our git repo.

Because of that I'm going to close this PR. I'll update the go version requirement.

PR #203, opened 2017-10-17
Bugfix: let libvirt populate channel source

moio:
Current behavior is to set the source path to empty string, which blocks the Guest Agent from working correctly.


flavio:
@moio and @monstermunchkin: I just updated the package inside of virtualization:containers to include this fix

Issue #204, opened 2017-10-17
SSH provisioner cannot connect


moio:
#  Version Reports:

### Distro  version of host:

Ubuntu 17.04

### Terraform Version Report

Terraform v0.10.7

### Libvirt version

``` virsh --version
2.5.0
```

### terraform-provider-libvirt plugin version (git-hash)

Current master: c034dc8e65b815ac49e1e7d90bf14ac346425529

# Description of Issue/Question

Terraform fails to connect via SSH to newly deployed VMs, reason is libvirt cannot get the machines' IPs despite the Guest Agent is in execution.

Part of the problem is already addressed in: https://github.com/dmacvicar/terraform-provider-libvirt/pull/203

With that PR in, there is still a rather unpredictable behavior: with resources with `count` > 1 only one VM succeeds, others do not.

Note that all VMs are reachable via the network, and all guest agents are running and returning correct data (not necessarily at the same time, but eventually). Example:

```
virsh # qemu-agent-command minion-1 '{"execute":"guest-network-get-interfaces"}'
error: Guest agent is not responding: QEMU guest agent is not connected

(not ready)

virsh # qemu-agent-command minion-1 '{"execute":"guest-network-get-interfaces"}'
{"return":[{"name":"lo","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"127.0.0.1","prefix":8},{"ip-address-type":"ipv6","ip-address":"::1","prefix":128}],"hardware-address":"00:00:00:00:00:00"},{"name":"eth0","ip-addresses":[{"ip-address-type":"ipv6","ip-address":"fe80::402e:93ff:fe43:f6f1","prefix":64}],"hardware-address":"42:2e:93:43:f6:f1"}]}

(agent ready but no DHCP lease yet)

virsh # qemu-agent-command minion-2 '{"execute":"guest-network-get-interfaces"}'
{"return":[{"name":"lo","ip-addresses":[{"ip-address-type":"ipv4","ip-address":"127.0.0.1","prefix":8},{"ip-address-type":"ipv6","ip-address":"::1","prefix":128}],"hardware-address":"00:00:00:00:00:00"},{"name":"eth0","ip-addresses":[{"ip-address-type":"ipv6","ip-address":"fe80::1ca6:f8ff:fe0c:4031","prefix":64}],"hardware-address":"1e:a6:f8:0c:40:31"}]}

(agent ready but no DHCP lease yet, second domain)

[...]

PR #205, opened 2017-10-19
Improved the wait-for-leases mechanism

inercia:
* simplify the wait-for-leases mechanism by just 1) keeping a `waitForLease` set of interfaces 2) insisting periodically on checking if we have an IP address for each interface in that set
* move all the domain-related code to a `domain.go` file and rename all the functions following the same `domain*` scheme.
* add more (probably too verbose and annoying) debug messages.


libvirt/domain.go
```diff
@@ -0,0 +1,171 @@
...
+			log.Printf("[DEBUG] found IPs for MAC=%+v: %+v\n", mac, ifaceWithAddr.Addrs)
+			return true, false, nil
+		}
+	}
+
+	log.Printf("[DEBUG] %+v doesn't have IP address(es) yet...\n", mac)
+	return false, false, nil
+}
+
+func domainGetState(domain libvirt.Domain) (string, error) {
```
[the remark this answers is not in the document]
inercia: I agree with you. But I didn't want to do so many change in this PR and keep most of the code as it is. I would leave this kind of cleanups for a following PR...

libvirt/domain.go
```diff
@@ -0,0 +1,171 @@
...
+				return false, domWaitLeaseStillWaiting, nil
+			}
+		}
+
+		log.Printf("[DEBUG] all the %d IP addresses obtained for the domain", len(waitForLeases))
+		return true, domWaitLeaseDone, nil
+	}
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"waiting-addresses"},
```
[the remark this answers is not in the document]
inercia: Right

libvirt/domain.go
```diff
@@ -0,0 +1,171 @@
...
+	// interfaces that are not attached to networks managed by libvirt
+	// (eg. bridges, macvtap,...)
+	log.Print("[DEBUG] fetching networking interfaces using qemu-agent")
+	interfaces := qemuAgentGetInterfacesInfo(&domain, true)
+	if len(interfaces) > 0 {
+		// the agent will always return all the interfaces, both the
+		// ones managed by libvirt and the ones attached to bridge interfaces
+		// or macvtap. Hence it has the highest priority
+		return interfaces, nil
+	}
```
monstermunchkin: The new qemu-agent code causes our tests to time out. They cannot connect to any qemu-agent because there is none. We could work around this issue:
```go
var skipQemuAgent = "false"

// ...

func domainGetIfacesInfo(domain libvirt.Domain) ([]libvirt.DomainInterface, error) {
	if skipQemuAgent == "false" {
		// get all the interfaces using the qemu-agent, this includes also
		// interfaces that are not attached to networks managed by libvirt
		// (eg. bridges, macvtap,...)
		log.Print("[DEBUG] fetching networking interfaces using qemu-agent")
		interfaces := qemuAgentGetInterfacesInfo(&domain, true)
		if len(interfaces) > 0 {
			// the agent will always return all the interfaces, both the
			// ones managed by libvirt and the ones attached to bridge interfaces
			// or macvtap. Hence it has the highest priority
			return interfaces, nil
		}
	} else {
		log.Println("[DEBUG] skipping qemu-agent")
	}

	// ...
}
```

Then, we could run our tests with `go test -ldflags "-X github.com/dmacvicar/terraform-provider-libvirt/libvirt.skipQemuAgent=true" -v -covermode=count -coverprofile=profile.cov -timeout=1200s ./libvirt`, causing the test cases to always fall back on the libvirt API.

This was the first workaround which came to my mind so it might not be the best. What do you think?

libvirt/domain.go
```diff
@@ -0,0 +1,171 @@
...
+	return err
+}
+
+func domainIfaceHasAddress(domain libvirt.Domain, iface libvirtxml.DomainInterface) (found bool, ignore bool, err error) {
+
+	mac := strings.ToUpper(iface.MAC.Address)
+	if mac == "" {
+		log.Printf("[DEBUG] Can't wait without a MAC address: ignoring interface %+v.\n", iface)
+		// we can't get the ip without a mac address
+		return false, true, nil
```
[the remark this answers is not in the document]
inercia: I'm still reluctant to do any important changes in things that are not directly related to what we are trying to fix in this PR. I would leave this for a subsequent PR...

libvirt/domain.go
inercia: You are right. Tests are timing out because we are insisting on trying connect to the `qemu-agent`. I like your idea of having a _qemu-agent switch_, but I would use an environment variable (ie, `TF_SKIP_QEMU_AGENT`) instead of a global because it seems more natural to me (we already have some other `TF_` vars)

moio:
I confirm this addresses #204, not sure about the Travis CI and review status. @inercia if I can help just ping!

monstermunchkin:
@flavio can this be merged?

pfremm:
Is this still actively being worked on?  Right now I am having issues with remote-exec since I cannot connect to any of the domains I've created with libvirt.

MalloZup:
I think yes, it is only s matter of resolving the conflicts and merge it

libvirt/resource_libvirt_domain.go
```diff
@@ -768,7 +750,7 @@ func resourceLibvirtDomainUpdate(d *schema.ResourceData, meta interface{}) error
 	}
 	defer domain.Free()
 
-	running, err := isDomainRunning(*domain)
+	running, err := domainIsRunning(*domain)
```
MalloZup: yop it is better to have it as category ```domain`` functions

travis/run-tests-inside-guest
```diff
@@ -1,19 +1,23 @@
 #!/bin/bash
 set -x
 
+base="/root/go/src/github.com/dmacvicar"
```
MalloZup: @inercia  would be nice if we can update also the script ``` travis/run-tests-inside-guest``` , this is used when we do 

```make test``` in dev mode :sunflower:

inercia: Maybe if we shared several variables, but I think that is an overkill just for using one common variable...

MalloZup: yop, i think we can leave with the 2 files for this time beeing

libvirt/domain.go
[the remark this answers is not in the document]
MalloZup: we can create an issue for this so we don't forget

libvirt/domain.go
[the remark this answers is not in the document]
MalloZup: also here, i would just create an issue so we can remember and pickup it later

travis/run-tests-inside-guest
pfremm: Thanks!  I am hoping this resolves my timeout issues when using wait-for-lease.

PR #214, opened 2017-11-05
Rename Uri to URI

MalloZup:
Uri --> URI
add some basic comments for exported function(fixing golint)

### DIFF
--- a/libvirt/provider.go
+++ b/libvirt/provider.go
@@ -30,7 +30,7 @@ func Provider() terraform.ResourceProvider {
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	config := Config{
-		Uri: d.Get("uri").(string),
+		URI: d.Get("uri").(string),
 	}
 
 	return config.Client()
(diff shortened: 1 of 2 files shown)


libvirt/provider.go
```diff
@@ -8,7 +8,7 @@ import (
 func Provider() terraform.ResourceProvider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
-			"uri": &schema.Schema{
+			"URI": &schema.Schema{
```
flavio: Let's leave that to `uri` otherwise all the existing terraform files will be broken.

libvirt/provider.go
```diff
@@ -30,7 +30,7 @@ func Provider() terraform.ResourceProvider {
 
 func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	config := Config{
-		Uri: d.Get("uri").(string),
+		URI: d.Get("URI").(string),
```
flavio: Same here, let's leave that to `uri` otherwise all the existing terraform files will be broken.

MalloZup:
ok thx i will do it :running_man:  :rofl:

MalloZup:
thx

Issue #131, opened 2017-06-13
consider moving to terra-farm or core?


josephholsten:
I don't know if libvirt is ready to start joining core, but if it is I'd like to help with the process.

If we want to not actually be in core, some folks are putting together an org like [sous-chefs](https://github.com/sous-chefs) does for chef cookbooks. Then maintainers can help each other out reviewing, merging and releasing changes. The main advantage is having more than one maintainer who can pass the torch. If that sounds interesting, the conversation is happening at https://github.com/terra-farm/meta/issues

If that doesn't sound ideal, please let me know!

flavio:
I personally think this is an exciting idea.

IMHO the code needs some refactoring and we have to tag an initial release of the project before joining.

What do the others think about it?

dmacvicar:
I am more in favour of identifying the cleanups needed to be in core more than an extra project.

If this umbrella would be already something active and popular, I would see things differently. At least, there is more than one person with commit access to the repo already so maintainers are not the problem.

josephholsten:
@dmacvicar totally understand. If there's anything I can do to help, do let me know.

One of the things I'm working on is regularly running acc tests for some providers. I can't tell from the Travis.yml, is libvirt running acc tests on every pr? Or just unit?

I don't know if there's any other shared resources we could use, but let me know if you think of anything.

flavio:
The acceptance tests are run on each PR together with the unit tests. (See https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/.travis.yml#L28)

dmacvicar:
BTW providers were already deleted from core, 
https://github.com/hashicorp/terraform/commit/1ab40eae3532d285ec94bfbf20a5d62ffc1f9d63

and there is a project here:
https://github.com/terraform-providers

I think we should try to move there.

josephholsten:
@dmacvicar absolutely! Would someone like to open an issue in hashicorp/terraform for this? Hopefully, it'll just involve a code review pass and then be migrated.

flavio:
These are the things I would like to see done before moving to https://github.com/terraform-providers:

  * Code cleanup:
     * Drop all the XML definitions by switching to the official `libvirt-go-xml` project. Tracked by issue https://github.com/dmacvicar/terraform-provider-libvirt/issues/143
[...]

PR #216, opened 2017-11-06
WIP: Split DomainCreate function into smaller parts.

MalloZup:
fix partially  #131

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -152,11 +152,6 @@ func resourceLibvirtDomain() *schema.Resource {
 				Optional: true,
 				Default:  "pc",
 			},
-			"arch": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
-				Default:  "x86_64",
-			},
 			"boot_device": &schema.Schema{
 				Type:     schema.TypeList,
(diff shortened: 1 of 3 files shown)


MalloZup:
@inercia @monstermunchkin @flavio do you have something else in mind for refactoring other then this?
https://github.com/dmacvicar/terraform-provider-libvirt/pull/216/files#diff-2f03be069459ebc91fcf0607d4d686d8R169

i was thinking to create an helper variadic function that could take  empty interfaces, with some fixed values for optional true or not, but was kind of overthinking that.

thx in advance :racehorse: :rocket: :robot:

flavio:
I fear @monstermunchkin was working on something similar. Am I wrong?

MalloZup:
@monstermunchkin  if you confirm, ok i will pick up something else thx @flavio

monstermunchkin:
I'm working on this [PR](https://github.com/dmacvicar/terraform-provider-libvirt/pull/193). My approach is also the opposite of what you are doing. IMHO having the `schema.Resource` struct without any helper functions makes it more readable even if the struct is then longer.

MalloZup:
yop @monstermunchkin i am fine then with that. thx for feedback

closing :mobile_phone_off:

PR #221, opened 2017-11-08
Make add/updateHost a bit more resilient

simonswine:
I had plenty of issues with static IP addreses and constant destroy and create of single instances without recreating the network. This fixes these problems with static IP addresses for me.

### DIFF
--- a/libvirt/utils_libvirt.go
+++ b/libvirt/utils_libvirt.go
@@ -46,6 +46,15 @@ func updateHost(n *libvirt.Network, ip, mac, name string) error {
 	return n.Update(libvirt.NETWORK_UPDATE_COMMAND_MODIFY, libvirt.NETWORK_SECTION_IP_DHCP_HOST, -1, xmlDesc, libvirt.NETWORK_UPDATE_AFFECT_CURRENT)
 }
 
+// Tries to update first, if that fails, it will add it
+func updateOrAddHost(n *libvirt.Network, ip, mac, name string) error {
+	err := updateHost(n, ip, mac, name)
+	if virErr, ok := err.(libvirt.Error); ok && virErr.Code == libvirt.ERR_OPERATION_INVALID && virErr.Domain == libvirt.FROM_NETWORK {
+		return addHost(n, ip, mac, name)
+	}
+	return err
+}
+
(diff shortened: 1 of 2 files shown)


MalloZup:
@simonswine thx for the fix!.
 imho as you mention, this works for you, but it is more a workaround then a real fix of the issue,

I am not at moment familiar with that part of code, but the fix has some code smells, because  we rely also for the error handling and updating on a string message, which in future could  be changed from libivrt. 

https://github.com/dmacvicar/terraform-provider-libvirt/pull/221/files#diff-9d553084ac827d330598622dca7aa216R53

So for merging it upstream i would consider a more generic solution for this pb

monstermunchkin:
We could also just remove the `&& strings.Contains(virErr.Message, "couldn't locate an existing dhcp host entry")` part. The error code and domain will most likely not be changed by libvirt.

libvirt/utils_libvirt.go
```diff
@@ -46,6 +47,17 @@ func updateHost(n *libvirt.Network, ip, mac, name string) error {
 	return n.Update(libvirt.NETWORK_UPDATE_COMMAND_MODIFY, libvirt.NETWORK_SECTION_IP_DHCP_HOST, -1, xmlDesc, libvirt.NETWORK_UPDATE_AFFECT_CURRENT)
 }
 
+// Tries to update first, if that fails, it will add it
+func updateAddHost(n *libvirt.Network, ip, mac, name string) error {
```
MalloZup: ```updateOrAddHost ``` maybe is better as function name here

MalloZup:
to me seems good a part of the 2 comments  , and without rely on string as mention before. 
:+1:

monstermunchkin:
@simonswine, are you still working on this?

simonswine:
Yes I am just not happy yet, as I saw some other issues when the mac address of my nodes changes. Hope to make some progress this weekend :slightly_smiling_face:

simonswine:
Thanks for review (and reminder) @monstermunchkin @MalloZup. Your comments should have been addressed now.

The thing I was not happy with is when a host is no longer in the network the delete is failing because it couldn't delete the non-existing host. I will address this in a separate PR later.

PR #226, opened 2017-11-10
Fix libvirt ovmf firmware integration test on non-Ubuntu

dmacvicar:
On SUSE systems, these paths are different.

In order to reuse the code that looks for the right files, we make the template and no-template case sub-tests of the main test containing the lookup code.

Also, if the files are not there, skip the test.

### DIFF
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"os"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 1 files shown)


libvirt/resource_libvirt_domain_test.go
```diff
@@ -607,21 +608,46 @@ func createNvramFile() (string, error) {
...
+		if _, err := os.Stat(template); os.IsNotExist(err) {
+			t.Skip("Can't test domain custom firmware template: OVMF template not found: %s")
+		}
+	}
+
+	t.Run("No Template", func(t *testing.T) {
+		subtestAccLibvirtDomain_FirmwareNoTemplate(t, nvram_path, firmware)
+	})
+	t.Run("With Template", func(t *testing.T) {
+		subtestAccLibvirtDomain_FirmwareTemplate(t, nvram_path, firmware, template)
```
MalloZup: ```nitpick``` can we avoid to use ```nvram_path``` and use camelCase ```nvramPath` (will make golint happy)

Issue #231, opened 2017-11-13
EASY-FIX: make some tests silient, and verbose only on failures


MalloZup:
At moment we have some tests that print a lot of debug data even if they pass.

A test should not print debug info when it pass, otherwise the whole testsuite becomes not human and maintainable. (imagine we have 400 tests with this debug)

MalloZup:
closing, fixed on that pr

PR #237, opened 2017-11-14
Dont print debug info if tests are ok

MalloZup:
the main motivation for this pr, is that we don't need debug info for a passing tests.

When a test fail we are looking for debug info. Otherwise debug a failed test become more diffult with noisy and non relevant information.
**Note:*** i removed only the bigdumps, this took a lot of lines, like more then 13 lines pro test, the 2-3 lines debug are ok imho, but i can remove them of course, but i think they are fine.

I was thinking to having an extra env. variable for printing debug info, but this complicate the whole design for a testsuite.
Imho a testsuite acceptance, should be simple as possible, for maintaining it longterm.
When a test fail, we have debug infos, otherwise the tests are there only for catching regression, so if it pass we don't need extra infos. ( we just want that they pass before merging pr).

Additional extra info on a test, from my POV is a code-smell, in a sense that there is either a lack of an addtional test, or the test is doing to much. (additional info when passing )

Thx for any review and feedback

### DIFF
--- a/libvirt/disk_def_test.go
+++ b/libvirt/disk_def_test.go
@@ -14,28 +14,20 @@ func init() {
 
 func TestDefaultDiskMarshall(t *testing.T) {
 	b := newDefDisk()
-	prettyB := spew.Sdump(b)
-	t.Logf("Parsed default disk:\n%s", prettyB)
-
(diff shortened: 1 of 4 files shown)


MalloZup:
fix #231

PR #239, opened 2017-11-16
Supports disks from the network using QEMU curl backend

dmacvicar:
QEMU supports a [http url for a block device](https://git.qemu.org/?p=qemu.git;a=blob;f=block/curl.c).

This can be used from the command line:

```console
qemu-kvm -m 1G -drive media=cdrom,driver=raw,node-name=cdrom,file.driver=http,file.url=http://download.opensuse.org/tumbleweed/iso/openSUSE-Tumbleweed-DVD-x86_64-Current.iso
```

and from libvirt

```xml
<disk type='network' device='cdrom'>
    <driver name='qemu' type='raw'/>
    <source protocol="http" name="url_path">
      <host name="hostname" port="80"/>
    </source>
    <target dev='hde' bus='ide' tray='open'/>
    <readonly/>
  </disk>
```
This PR allows to add a disk to a domain like this:

```hcl
disk {
  url = "http://download.opensuse.org/tumbleweed/iso/openSUSE-Tumbleweed-DVD-x86_64-Current.iso"
}
```

Which, combined with boot order:

```hcl
  boot_device {
      dev = [ "cdrom", "hd", "network"]
  }
```

Allows you to boot directly from a remote ISO, without the need to load it into the libvirt pool.

### DIFF
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/url"
 	"os"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


libvirt/resource_libvirt_domain_test.go
```diff
@@ -202,6 +203,38 @@ func TestAccLibvirtDomain_ScsiDisk(t *testing.T) {
...
 }
 
+func TestAccLibvirtDomainUrlDisk(t *testing.T) {
+	var domain libvirt.Domain
+	u, err := url.Parse("http://download.opensuse.org/tumbleweed/iso/openSUSE-Tumbleweed-DVD-x86_64-Current.iso")
+	if err != nil {
+		t.Error(err)
+	}
+
+	var configUrl = fmt.Sprintf(`
```
monstermunchkin: golint complaint: it should be `configURL`

libvirt/resource_libvirt_domain_test.go
```diff
@@ -202,6 +203,38 @@ func TestAccLibvirtDomain_ScsiDisk(t *testing.T) {
...
+                    }
+            }`, u.String())
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtDomainDestroy,
+		Steps: []resource.TestStep{
+			resource.TestStep{
+				Config: configUrl,
```
monstermunchkin: Same as above.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -590,6 +623,41 @@ func testAccCheckLibvirtScsiDisk(n string, domain *libvirt.Domain) resource.Test
 	}
 }
 
+func testAccCheckLibvirtUrlDisk(u *url.URL, domain *libvirt.Domain) resource.TestCheckFunc {
```
monstermunchkin: golint complaint: it should be `testAccCheckLibvirtURLDisk`

dmacvicar:
All changes done. Tests pass now.

inercia:
Our code currently supports downloading images, so:

* We support the `If-Modified-Since` header so we do not download the same image if it is already present. Does this `curl` downloader honor that header?
* Our code does the download to the LibVirt pool (so it can be used as a base and shared by multiple VMs). I guess this is a _per-domain_ attribute, right?

dmacvicar:
@inercia this is a qemu block device driver, so the image is not downloaded. Blocks are requested from the block layer using ranges, directly via `http`. That is why it is an attribute of `disk` and not volume. It has nothing to do with libvirt storage pool. `disk` can operate on top of storage volumes, but there are other drivers as well. This `url` is an alternative to `volume_id` at the `disk` level. `source` at the volume level is handled via the provider, at the volume level.

inercia:
@dmacvicar Ok, thanks for the clarification.

PR #243, opened 2017-11-17
(don't merge) splitted it in 2 prs

### LINKED ISSUE #119: Provider ignores "running" parameter for domain
#  Version Reports:

### Distro  version of host:

Ubuntu 14.04

### Terraform Version Report

v0.8.6

### Libvirt version

1.3.1

a07e21b74726d9ce443ac9332417c169b2ccd708
___
Despite "running = false" domain always running.

### Setup
```
provider "libvirt" {
  uri = "qemu+ssh://root@host/system"
}

resource "libvirt_domain" "rs-removeme" {
  name = "rs-removeme"
  vcpu = 1
  memory = 4096
  running = false
}
```

### Steps to Reproduce Issue
1) Create file with state above
2) terraform apply. [See log](http://paste.openstack.org/show/603763/)

Expected result:
VM is not running

Actual result:
VM is running.

___
# Additional Infos:

No Apparmor

MalloZup:
- [x] do pr #255  

- [ ]  Create other pr after #255 is merged for other changes

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -79,12 +79,6 @@ func resourceLibvirtDomain() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
-			"running": {
-				Type:     schema.TypeBool,
-				Optional: true,
-				Default:  true,
-				ForceNew: false,
-			},
 			"domain_shutoff": {
 				Type:     schema.TypeBool,
 				Optional: true,
@@ -885,12 +879,6 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
 	// or it will show as changed
 	d.Set("emulator", domainDef.Devices.Emulator)
 
-	running, err := isDomainRunning(*domain)
-	if err != nil {
-		return err
-	}
-	d.Set("running", running)
-
 	var disks []map[string]interface{}
 	for _, diskDef := range domainDef.Devices.Disks {
 		// network drives do not have a volume associated
@@ -1248,12 +1236,7 @@ func destroyDomainByUserRequest(d *schema.ResourceData, domain *libvirt.Domain) 
 			return fmt.Errorf("Couldn't destroy libvirt domain: %s", err)
 		}
 	}
-	statusDomain, err := isDomainRunning(*domain)
-	if err != nil {
-		return err
-	}
 
-	d.Set("running", statusDomain)
 	return nil
 }
 
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -85,6 +85,13 @@ func resourceLibvirtDomain() *schema.Resource {
 				Default:  true,
 				ForceNew: false,
 			},
+			"domain_shutoff": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: false,
+				Required: false,
+			},
 			"cloudinit": {
 				Type:     schema.TypeString,
 				Required: false,
@@ -713,7 +720,7 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 			}
 		}
 	}
-
+	destroyDomainByUserRequest(d, domain)
 	return nil
 }
 
@@ -1217,6 +1224,37 @@ func isDomainRunning(domain libvirt.Domain) (bool, error) {
 	}
 
 	return state == libvirt.DOMAIN_RUNNING, nil
+}
+
+func destroyDomainByUserRequest(d *schema.ResourceData, domain *libvirt.Domain) error {
+	if !d.Get("domain_shutoff").(bool) {
+		return nil
+	}
+
+	domainID, err := domain.GetUUIDString()
+
+	if err != nil {
+		return fmt.Errorf("Error retrieving libvirt domain id: %s", err)
+	}
+
+	log.Printf("Destroying libvirt domain %s", domainID)
+	state, _, err := domain.GetState()
+	if err != nil {
+		return fmt.Errorf("Couldn't get info about domain: %s", err)
+	}
+
+	if state == libvirt.DOMAIN_RUNNING || state == libvirt.DOMAIN_PAUSED {
+		if err := domain.Destroy(); err != nil {
+			return fmt.Errorf("Couldn't destroy libvirt domain: %s", err)
+		}
+	}
+	statusDomain, err := isDomainRunning(*domain)
+	if err != nil {
+		return err
+	}
+
+	d.Set("running", statusDomain)
+	return nil
 }
 
 func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml.DomainDisk, error) {
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -116,6 +116,69 @@ func TestAccLibvirtDomain_Volume(t *testing.T) {
 	})
 }
 
+// we want to destroy (shutdown volume after creation)
+func TestShutoffDomain(t *testing.T) {
+	var domain libvirt.Domain
+	var volume libvirt.StorageVol
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtDomainDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+			            resource "libvirt_volume" "acceptance-test-volume" {
+			                    name = "terraform-test"
+			            }
+			            resource "libvirt_domain" "acceptance-test-domain" {
+			                    name = "terraform-test"
+													domain_shutoff = true
+			                    disk {
+			                            volume_id = "${libvirt_volume.acceptance-test-volume.id}"
+			                    }
+			            }`),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLibvirtDomainExists("libvirt_domain.acceptance-test-domain", &domain),
+					testAccCheckLibvirtVolumeExists("libvirt_volume.acceptance-test-volume", &volume),
+					testAccCheckLibvirtDomainStateEqual("libvirt_domain.acceptance-test-domain", &domain, "shutoff"),
+				),
+				ExpectNonEmptyPlan: true,
+			},
+		},
+	})
+}
+
+func TestShutoffMultiDomainsRunning(t *testing.T) {
+	var domain libvirt.Domain
+	var domain2 libvirt.Domain
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtDomainDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+   					resource "libvirt_domain" "domainoff" {
+	               name = "domainfalse"
+	               vcpu = 1
+	               domain_shutoff = true
+	          }
+            resource "libvirt_domain" "domainok" {
+	               name = "domaintrue"
+	               vcpu = 1
+	               domain_shutoff = false
+	          }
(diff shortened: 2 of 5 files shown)


MalloZup:
@dmacvicar @monstermunchkin  you can have a look for reviewing once you have free time. Thx in advance happy weekend

dmacvicar:
Can you explain some context? I have no idea of this "feature" of libvirt. What is the usecase?, motivation? implementation overview? advantages? disadvantages?

MalloZup:
@dmacvicar sure, i'm fixing with this pr the issue #119.

So this is more a bugfix then a new feature.

The issue was, that 

Despite "running = false" domain always running.

In the doc, we are saying: (https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/domain.html.markdown)running

 - (Optional) Use false to turn off the instance. If not specified, true is assumed and the instance, if stopped, will be started at next apply.

As by the issue stated, setting false has no effect. So i'm fixing this, when set to false, the domain is shut off

libvirt/resource_libvirt_domain.go
```diff
@@ -28,6 +28,12 @@ type DomainMeta struct {
 // PoolSync exported pool sync
 var PoolSync = NewLVirtPoolSync()
 
+// destroyDomainRequestedByUser if in a tf we have running =false
+// we want to shutdown the domain after creation.
+// we need to cache this var, since the running var is used for other operations.
+// if this var is true, we will destroy in the create. (this is init in read)
+var destroyDomainRequestedByUser bool
```
monstermunchkin: I would shorten this:
```
destroyDomainRequestedByUser tracks the desired state of the domain. If set to true, the domain will be destroyed after its creation.
```

libvirt/resource_libvirt_domain_test.go
```diff
@@ -115,6 +115,38 @@ func TestAccLibvirtDomain_Volume(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+			            resource "libvirt_volume" "acceptance-test-volume" {
+			                    name = "terraform-test"
+			            }
+
+			            resource "libvirt_domain" "acceptance-test-domain" {
+			                    name = "terraform-test"
+													running = false
```
monstermunchkin: Too much indentation.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -490,6 +522,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
...
+			return fmt.Errorf("Libvirt domain not found")
+		}
+
+		// 1) test that domain is not up
+		domainRuns, err := IsDomainRunning(*domain)
+		if err != nil {
+			return err
+		}
+		if domainRuns {
+			return fmt.Errorf("Domain should be destroyed! TEST FAILED")
```
monstermunchkin: `Domain should not be running` No need for `TEST FAILED` since `go test` will show that the test has failed.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -490,6 +522,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
...
+		}
+		// 2 check that status is shutdown
+		state, err := GetDomainState(*domain)
+		if err != nil {
+			return fmt.Errorf("could not get domain state: %s", err)
+		}
+
+		if state != "shutoff" {
+			return fmt.Errorf("Domain state should be == shutoff, but is %s, TEST FAILED", state)
```
monstermunchkin: `Domain state should be 'shutoff' but is '%s'`. Again, no need for `TEST FAILED`.

dmacvicar:
Ok now I understand. I was confusing the resource destroy with domain destroy. The jargon for libvirt resource destroy is undefine, so it looks good to me that the domain is "destroyed" if you don't want it running.

Now, I did not implement this "running" functionality (IIRC at the beginning domains were not defined), so you may want someone that knows that concept to give it an OK.

libvirt/resource_libvirt_domain.go
```diff
@@ -698,6 +704,8 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	defer destroyDomainByUserRequest(d, domain)
```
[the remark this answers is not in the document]
MalloZup: @monstermunchkin  why we don't need `defer` here? imho is more elegant to use it. 
I mean this destroying function is a `cleanup` function

MalloZup:
@monstermunchkin thx for your review ! :sunflower: 

i was thinking that the defer function is much more elegant there.

libvirt/resource_libvirt_domain.go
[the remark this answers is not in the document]
dmacvicar: What @monstermunchkin means it that you want to call destroyDomainByUserRequest _there_ and not later. Any programmer reading the code would think "why is he deferring something he wants to do now?".
While in the other example `defer domain.Free()` you are indeed intending "Call `domain.Free()` once all the rest of the code is executed, and _not_ before, because I will be using `domain`".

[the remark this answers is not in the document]
MalloZup: @dmacvicar @monstermunchkin  thx for clarification. 
I put `defer` there, because we are shutting down a libvirt domain, and imho it should be added later like `defer domain.Free`, for one reason only: 
- imagine the domain is shut off (down). and people are adding after that function other operations on domain. 
The question is , are all domain operation possible on a shutdown domain without errors? 
So that's why i put this in a `defer` because to me seemed a function that need to go @ end

libvirt/resource_libvirt_domain.go
MalloZup: good

MalloZup:
@monstermunchkin @dmacvicar ok all comments adressed. :dog2: thx

dmacvicar:
@MalloZup The fact that your function is called destroy it does not mean it is cleanup of a resource. You have no resource that you created there.

Now changing topics. I smell something fishy about `var destroyDomainRequestedByUser bool`being global. Why did you make it global?
Isn't the value going to be overwritten by two domains in the same .tf file?

I think there is a different place to store state for the resource. I'd look either in the properties itself or something around DomainMeta.

MalloZup:
@dmacvicar  ok thx for the catch.  

The global var is a code smell that i will address this asap.

dmacvicar:
More than a code smell, is a bug. It would make each domain overwrite data from the other. It is not just a question of style.

MalloZup:
@dmacvicar, i tried to  say it with lightness. Of course i know it that it is a bug ... that is also why i want create some tests

libvirt/resource_libvirt_domain.go
```diff
@@ -28,6 +28,9 @@ type DomainMeta struct {
 // PoolSync exported pool sync
 var PoolSync = NewLVirtPoolSync()
 
+// destroyDomainRequestedByUser tracks the desired state of the domain. If set to true, the domain will be destroyed after its creation.
+var destroyDomainRequestedByUser bool
```
inercia: Do we really need a global here? Isn't it a domain attribute?

libvirt/resource_libvirt_domain.go
```diff
@@ -1215,6 +1220,38 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
 	return disk, nil
 }
 
+func destroyDomainByUserRequest(d *schema.ResourceData, domain *libvirt.Domain) error {
+	// if the running is true in the tf file, we don't destroy the domain
+	if !destroyDomainRequestedByUser {
```
inercia: As I said, couldn't we get the `destroyDomain` as an attribute? like `d.Get('destroyDomain')` or something like that?

libvirt/resource_libvirt_domain.go
MalloZup: @inercia  yop this i will fix, is a mistake from my part.

libvirt/resource_libvirt_domain.go
MalloZup: i'm thinking now how to solve this, i will address this asap

MalloZup:
@inercia thx for review. i will ping you guys when i am done. I'm writing a new ACC-tests and will continue it later the fix.
:crossed_fingers:

libvirt/resource_libvirt_domain_test.go
```diff
@@ -116,6 +116,71 @@ func TestAccLibvirtDomain_Volume(t *testing.T) {
...
+					testAccCheckLibvirtVolumeExists("libvirt_volume.acceptance-test-volume", &volume),
+					testAccCheckLibvirtDomainIsDestroyed("libvirt_domain.acceptance-test-domain", &domain),
+				),
+			},
+		},
+	})
+}
+
+// we have 4 running=true and 4 running=false domain
+func TestRunningFalseDomainConcurenncyMultiDomainsRunning(t *testing.T) {
```
MalloZup: open for name improvements :tangerine:

libvirt/resource_libvirt_domain_test.go
```diff
@@ -523,6 +593,103 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
 	return nil
 }
 
+func testAccCheckLibvirtDomainIsRunning(n string, domain *libvirt.Domain) resource.TestCheckFunc {
```
MalloZup: this  Helper Function can be refactored with the other function.

I am in favor with the generic refactoring

```golang
func testAccCheckLibvirtDomaiStatusEqual (n string, domain *libvirt.Domain, domainStatusExpected string) 
```

MalloZup:
so i fixed all issue.
what is new here:

- added multidomain tests for avoid concurency race condition. ( 3 domain run 3 are shut down expected)
- added test function DomainStateEqual
- the destroy is controlled over a `domain_shutoff` property and not over global var
- added documentation 

thx for reviews in advance

MalloZup:
@dmacvicar @inercia @monstermunchkin  so i fixed and adressed your comments.

Feel free to review it. Take your :watch:  and :tea: . I will :tent: and :running_man:  :rofl:

website/docs/r/domain.html.markdown
```diff
@@ -31,8 +31,8 @@ The following arguments are supported:
   will be created.
 * `memory` - (Optional) The amount of memory in MiB. If not specified the domain
   will be created with 512 MiB of memory be used.
-* `running` - (Optional) Use `false` to turn off the instance. If not specified,
-  true is assumed and the instance, if stopped, will be started at next apply.
+* `domain_shutoff` - (Optional) Use `true` to turn off the instance. If not specified,
```
MalloZup: i could change this with `domain_shutdown` open for name suggestions

monstermunchkin: This is confusing:
* in the docs, you are replacing `running` with `domain_shutoff`
* however, the option `running` still exists

Why do we need both options?

MalloZup: from my POV the problem was already there in the variable `running`.

We are using this property, for doing other operations, like refresh etc. The `running` is kind of internal property that we use in the domain for know if the domain is up or not, and then we do other stuff. 

The `shutoff_domain` var is only an option that we need to say that after creation we need to shutoff.

This is why i created the other var.
 I could create the runnig var in the meta Domain struct. if you want

monstermunchkin: I just looked at the code, and cannot find a reason to keep the `running` option. Setting it in a `.tf` file, won't change anything. Am I missing something or is this option useless?

MalloZup: i think the option is useless. i will remove it

MalloZup:
@monstermunchkin  ok good catch :+1: :1st_place_medal: . btw we had the same `running` in network, and i removed it ( which caused also the cleanup/removal of the network_update function.) :sunflower: :sake:

website/docs/r/domain.html.markdown
MalloZup: domain_shutoff or domain_shutdown ?

MalloZup:
@monstermunchkin  thx  for feedback! appreciate it. 

So locally `bei mir` the test are ok. travis will take some time

MalloZup:
#255

MalloZup:
i am closing this, and i have done another pr #257

PR #256, opened 2017-11-29
check uniqueness of storage volume name

monstermunchkin:
The storage volume name needs to be unique. If this is not the case, the
storage volume will be overwritten which is not problematic itself.
However, `terraform destroy` will fail since it will try and delete the
same storage volume twice.

Consider the following example:
```hcl
provider "libvirt" {
  uri = "qemu:///system"
}

resource "libvirt_volume" "disk_1" {
  name   = "vol_1"
  source = "http://example.org/image.qcow2"
}

resource "libvirt_volume" "disk_2" {
  name   = "vol_1"
  source = "http://example.org/image.qcow2"
}
```

Running `terraform apply` works perfectly. However, running `terraform destroy`
would fail with the below message if the uniqueness were to be ignored.
```
Storage volume not found: no storage vol with matching name 'vol_1'
```

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -105,11 +105,14 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 		return pool.Refresh(0)
 	})
 
-	volumeDef := newDefVolume()
-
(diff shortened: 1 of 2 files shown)


MalloZup:
lgtm

PR #261, opened 2017-12-11
Fix support for raw volumes

solidnerd:
Hey,

thanks for this nice project. I use it currently very heavily. I found a bug when I try to use `raw` volumes then I got the following message:

```console
* libvirt_domain.template: Error creating libvirt domain: virError(Code=1, Domain=10, Message='internal error: process exited while connecting to monitor: 2017-12-11T19:57:09.278580Z qemu-system-x86_64: -drive file=/var/lib/libvirt/images/template_docker,format=qcow2,if=none,id=drive-virtio-disk1: Image is not in qcow2 format')
```

The problem is that the image is correctly converted as raw image, but the libvirt definition is not correct and assumes that it's an qcow2 image.

My solution is to check the file extension like you do when using a volume from url. Futhermore I added the file exentsion to qcow2 volumes so that anything except qcow2 will be defined as raw.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -422,6 +422,10 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 				return fmt.Errorf("Error retrieving volume file: %s", err)
 			}
 
+			if !strings.HasSuffix(diskVolumeFile, ".qcow2") {
+				disk.Driver.Type = "raw"
+			}
+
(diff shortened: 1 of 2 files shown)


inercia:
This seems to be related to (or maybe in conflict with) #183.

MalloZup:
@solidnerd  we are going to merge soon this PR which will adress this pb in more generic way,
thank you for your PR :sunflower: 

https://github.com/dmacvicar/terraform-provider-libvirt/pull/380

PR #263, opened 2017-12-21
fix difference detection when using an existing bridge network

munkyboy:
given the terraform:

```json
resource "libvirt_domain" "myvm" {
  network_interface {
    bridge = "some-bridge"
    network_name = "existing-bridge-network"
  }
}
```

`terraform plan` would always think it needs to re-apply the domain


munkyboy:
code coverage went down but AFAICT, the block I edited did not have code coverage previously.
https://coveralls.io/builds/14755903/source?filename=libvirt%2Fresource_libvirt_domain.go#L1070

PR #268, opened 2018-01-07
UEFI/OVMF booting: Don't check for existance of firmware/NVRAM

night199uk:
This check fails if we're provisioning a remote libvirt instance as the files won't exist on the local machine anyway.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -8,7 +8,6 @@ import (
 	"log"
 	"net"
 	"net/url"
-	"os"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


flavio:
What happens after if these files are not available and the checks are not in place?

night199uk:
apologies, was away for a bit. if the files don't exist the domain will fail to start. this is one of the only places I could find in terraform-provider-libvirt where we check for the existence of the files though. this makes us consistent with other places and gets remote UEFI VMs running again.

checking for files on a remote server is a difficult problem without some channel outside virsh (e.g. ssh). i don't think we can solve that problem here. but it is inconsistent right now that non-UEFI VMs work fine, and UEFI VMs break.

inercia:
@night199uk please rebase your PR so we can proceed with the merge...

dmacvicar:
Closing. Anybody is welcome to contribute a rebase.

PR #269, opened 2018-01-07
Modify cloudinit to use 'mkisofs' instead of 'genisoimage'.

night199uk:
genisoimage was debian's fork of mkisofs so the two are compatible, but mkisofs is available on more OS (e.g. macOS via brew). The genisoimage Ubuntu package at least installs 'mkisofs' also.

### DIFF
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -150,7 +150,7 @@ func (ci *defCloudInit) createISO() (string, error) {
 
 	isoDestination := filepath.Join(tmpDir, ci.Name)
 	cmd := exec.Command(
-		"genisoimage",
+		"mkisofs",
 		"-output",
 		isoDestination,
 		"-volid",


multani:
But then, Debian doesn't ship mkisofs anymore and this patch breaks the cloudinit resource on Debian :(

What would be the proper way of supporting these different mkisofs-variants so that I could propose an alternative pull request which hopefully would work for everybody?

night199uk:
On Ubuntu at least, genisoimage these days is a symlink to mkisofs if you install the mkisofs package. I think the genisoimage package is also just linked to mkisofs.
Is that not the same on Debian?

multani:
@night199uk Not that I'm aware. Actually, [genisoimage is now deprecated in favor of xorriso](https://wiki.debian.org/genisoimage) apparently. xorriso doesn't provide ``mkisofs`` either, but has a ``xorrisofs`` command which behaves as ``mkisofs``.

PR #275, opened 2018-01-22
Expose IP addresses for domains at the top level

pdf:
Accessing a list, nested inside a map, nested inside a list, is very
difficult in Terraform, so the existing structure makes it impossible to
extract IP addresses for all domains, when combined with `count` in a
domain definition.

Exposing `ip_address` on the domain makes IP addresses much more
accessible.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -117,6 +117,14 @@ func resourceLibvirtDomain() *schema.Resource {
 					Schema: networkInterfaceCommonSchema(),
 				},
 			},
+			"ip_address": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 			"graphics": {
 				Type:     schema.TypeMap,
 				Optional: true,
@@ -1082,6 +1090,14 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	log.Printf("[DEBUG] read: ifaces for '%s':\n%s", domainDef.Name, spew.Sdump(netIfaces))
 	d.Set("network_interface", netIfaces)
+
+	var ipAddresses []string
+	for _, iface := range ifacesWithAddr {
+		for _, addr := range iface.Addrs {
+			ipAddresses = append(ipAddresses, addr.Addr)
+		}
+	}
+	d.Set("ip_address", ipAddresses)
 
 	if len(ifacesWithAddr) > 0 {
 		d.SetConnInfo(map[string]string{


pdf:
Happy to update examples, etc if the change is approved.

inercia:
I don't think getting the IP aaddresses is so hard. For example, you could print them in the Terraform file with something like:

```
output "ips" {
  value = ["${libvirt_domain.node.*.network_interface.0.addresses.0}"]
}
```

pdf:
That doesn't work @inercia - Terraform can't index a splatted list like that.  Even if it did work, it would only return a single address, from a single interface.

inercia:
I'm afraid this _does_ work, @pdf. I just got that example from [this piece of code](https://github.com/inercia/caasp-tf/blob/master/terraform.tf#L212): it provides a list of IP addresses for all the domains in the cluster.

So I'm closing this issue as this is already provided by the Terraform language in a more general way...

pdf:
@inercia did you actually test it? Because that's exactly the same code I had, that prompted me to write this PR, because it does _not_ work when there is more than one node:

```
resource "libvirt_domain" "node" {
  name   = "${var.name}-${count.index}"
  memory = 1024
  vcpu   = 4

  cloudinit = "${libvirt_cloudinit.node.*.id[count.index]}"

  network_interface {
    network_name = "kube"
  }

  disk = ["${conditional_list_of_map.node_disks.*.result[count.index]}"]

  graphics {
    type        = "spice"
    listen_type = "address"
    autoport    = "yes"
  }

  console {
    type        = "pty"
    target_type = "serial"
    target_port = 0
  }

  count = 3
}

output "ip_addresses" {
  //value = ["${libvirt_domain.node.*.ip_address}"]
  value = ["${libvirt_domain.node.*.network_interface.0.addresses.0}"]
}
```

```
Error: Error running plan: 1 error(s) occurred:

* module.infra.module.libvirt.module.controller.output.ip_addresses: Resource 'libvirt_domain.node' does not have attribute 'network_interface.0.addresses.0' for variable 'libvirt_domain.node.*.network_interface.0.addresses.0'
```

pdf:
Yeah, looks like the reason your plan doesn't complain with the same code is that you're not actually reading that output.

inercia:
@pdf That code is actually being used in production and uses the format in [Terraform documentation](https://www.terraform.io/docs/commands/output.html). Maybe your version of Terraform does not support that. Please check you are using the latest version.

pdf:
@inercia look closely - you're not reading the output anywhere inside your code.  If you want to see what the issue is, move that output block (and its dependencies, obviously) into a module, then try to read it, like:

```hcl
"${module.whatevermodulename.nodes[0]}"
```

No where that I've seen in the Terraform documentation does it show reading an index value from a splat, that is, the `0`s (zeroes) in this stanza:

```
libvirt_domain.node.*.network_interface.0.addresses.0
```

I am indeed using the latest version of Terraform:

```
$ terraform version
Terraform v0.11.2
+ provider.conditional (unversioned)
+ provider.libvirt (unversioned)
```

Of course, arguably the broken behaviour is a bug in Terraform, but with them trying to move to HCL2, the chances of getting anything merged that affect HCL1 are slim to none, and there's no way right now of knowing whether this will be valid syntax under HCL2 either.

inercia:
I ave not much experience with output from modules, so maybe you are right and there is problem there. Maybe you should file a bug report in the Terraform repo...

pdf:
Here's one issue that references the problem (there are more besides, reporting the same problem).  In this issue, one of the Terraform devs suggests exposing the nested value as a top-level value, as I've done here, because they're not going to fix anything until they get the new HCL finished, which is going to take a long time:

https://github.com/hashicorp/terraform/issues/17048

pdf:
Does the ticket above make the current state of Terraform, and the position of its devs more clear @inercia? And so that I know whether or not I have to make our users to rely on a fork for this functionality going forward, can you confirm whether or not you are willing to accept a PR for it?  Is there anything that can be altered in this PR to make it more acceptable?

inercia:
I think we should not try to overcome the limitations in the Terraform core by doing hacks in this provider. The problem does not lie in this plugin, so it should be solved on the core (probably when HCL2 is available).

So, if we export IP addresses in the plugin we would be duplicating the source of truth and messing up the code just for being used in modules, and some other users could require the kind of changes for other data like, for example, MAC addresses.

I think the solution for this is to wait for HCL2...

pdf:
HCL2 will be a breaking change for everyone anyway, and could be another 6-12 months away (there are no firm estimates, and even the design is still in flux).  The majority of other providers expose a var to access IP addresses at the top level, because they're a requirement for connecting machines/services together (and output from modules is the only way to actually make use of IP addresses when integrating multiple components).

I suppose I can understand your position, but it means we're forced to make users install a fork for usable libvirt support. :disappointed:

PR #285, opened 2018-02-24
Fix broken tests due to incorrect usage of t.Skip()

multani:
Current tests were broken with the following error:

    # github.com/dmacvicar/terraform-provider-libvirt/libvirt
    libvirt/resource_libvirt_domain_test.go:874: Skip call has possible formatting directive %s
    libvirt/resource_libvirt_domain_test.go:882: Skip call has possible formatting directive %s
    FAIL	github.com/dmacvicar/terraform-provider-libvirt/libvirt [build failed]

t.Skip with formatting should be t.Skipf()


flavio:
Thanks!

PR #289, opened 2018-03-15
Add support for readonly disks

cbaines:
These changes add support for readonly disks, e.g.
```
  disk {
    file = "..."
    readonly = true
  }
```

I've tested this, and it somewhat works, but I get a confusing error regarding refreshing the state:
```
* libvirt_domain.domain-ubuntu: libvirt_domain.domain-ubuntu: Error retrieving volume for disk: virError(Code=50, Domain=18, Message='Storage volume not found: no storage vol with matching path '/gnu/store/d0nai2f8wqs231hsp4lrzqf04kpbnwmb-vm-image-and-system/image'')
```

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -158,6 +158,11 @@ func resourceLibvirtDomain() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
+						"readonly": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
 					},
 				},
 			},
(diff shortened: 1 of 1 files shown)


cbaines:
> I've tested this, and it somewhat works, but I get a confusing error regarding refreshing the state:

I've figured this out now. If the disk is readonly, it has no corresponding volume. I've now adjusted the code to not look for one in this case.

dmacvicar:
@cbaines I don't feel comfortable with that explanation. What do you mean that if a disk is readonly, it has no corresponding volume?

cbaines:
> @cbaines I don't feel comfortable with that explanation. What do you mean that if a disk is readonly, it has no corresponding volume?

Prior to my latest changes, I was getting an error when using this code. The actual machine would work fine, and the disk would be correctly configured, but as I say in the description, I was getting an error.

Unfortunately, it's a while since I looked at this, so I've forgotten the sources I used at the time, and I didn't put them in the commit message. But, I'm reasonably sure this is correct. If you create a virtual machine using this, I believe readonly disks won't appear in one of the storage volumes displayed by virt-manager, there just isn't any need, they are not going to change in any way, or take up more space, they are readonly.

What do you think?

MalloZup:
Closing doing stale and lack of answer. Feel free to reopen this and or ask for infosinfo thank you and enjoy software and also life:)

PR #291, opened 2018-04-08
Remove duplicated append to disks

cbaines:
I haven't encountered a problem with this, just noticed from looking at the code that it appears to be duplicated, as it happens immediately after the if statement.


inercia:
It seems you are right @cbaines. Thanks.

PR #302, opened 2018-06-01
add Makefile cleanup target

MalloZup:
This is usefull when tests fails and you need to remove the unclean resources that hangs on your systems

I dunno if i covered all possibly resources that we create during tests, i would propose that if make cleanup doesn't clean up, the name can be added  later

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,4 +27,8 @@ golint:
 gofmtcheck:
 	bash travis/run-gofmt
 
+
+clean:
+	./travis/cleanup.sh
+
 .PHONY: build install test vet fmt golint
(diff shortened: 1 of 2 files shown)


MalloZup:
CC @inercia

dmacvicar:
Doesn't this belong in the terraform test layer?

It has all this facilities to create resources, but they are not destroyed when the testsuite fails.

MalloZup:
exactly, when the test-suite fails there are all resources not destroyed.
Imagine you have a dev. workflow where you run the testsuite and to see your dev change works.
Testsuite fails, you need always to destroy the resource by hand, (volumes, domain, networks) by hand. 

Additionaly bonus from my pov would be also that we could keep track of the name of tests-resource created, since atm there is no convention about the names, but having this script could encourage contributors to have a fixed name for cleaning them up.

MalloZup:
@dmacvicar  i'm going to merge this since with the broken testsuite atm this spare time, i just do 

`make clean` and all domain which remains there go destroyed.

As you mentioned when things go wrong on terraform layer testsuite domain/networks are not cleaned-up

PR #270, opened 2018-01-08
Add cloud-init support for the network-config file.

night199uk:
This allows creating static and custom networking configurations.

PR #267, opened 2018-01-07
Allow for specifying custom meta-data for the cloud-init image.

night199uk:
Works exactly the same way as specifying custom user-data does.

PR #309, opened 2018-06-28
Add cloud-init support for the network-config file.

night199uk:
This allows creating static and custom networking configurations.

Re-opening #270 with documentation added.

### DIFF
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -25,6 +25,10 @@ const userData string = "user-data"
 // metaData is the filename expected by cloud-init
 const metaData string = "meta-data"
 
+// networkConfig is the filename expected by cloud-init
+const networkConfig string = "network-config"
+
+// CloudInitUserData struct
 type defCloudInitUserData struct {
 	SSHAuthorizedKeys []string `yaml:"ssh_authorized_keys"`
 }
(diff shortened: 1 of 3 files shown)


night199uk:
This should be rebased; all tests passing and documentation added. Hopefully this is gtg.

dmacvicar:
My question to the reviewers is if we want to go this path, and starting to add support for all the cloudinit schema.

inercia:
> My question to the reviewers is if we want to go this path, and starting to add support for all the cloudinit schema.

Not sure about this. On one hand, the network configuration seems to be a top-level element of the cloud-init schema, and it is important enough for being managed by the terraform/libvirt provider. On the other hand, this could open a can of worms where we would end up having all the possible cloud-init stuff...

My personal view would be to go for the network configuration, but nothing else.

night199uk:
My understanding is that there are only 3 files that can exist in the ISO?
meta-data
user-data
network-config

I don't think you need to add the entire cloud-init schema. At least allowing folks to place those 3 files avoids having to implement the schema and the user can source those files as they see fit (I use terraform templates).

MalloZup:
I would really like if we improve also the tests coverage for this feature

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/pull/267

MalloZup:
Closing this pr because it will be adressed on https://github.com/dmacvicar/terraform-provider-libvirt/pull/410, were we allow the new field, were the user can add cloud init network data. Thank you for contributing and your time spend on this pr ( reviewer and contributors). Have a nice day

PR #314, opened 2018-06-30
Bug fix: create network interfaces attached to non-DHCP networks

Bischoff:
If one declares a network without DHCP, then declaration of interfaces connected to that network are silently ignored.

The problem relies in the `continue` instruction which skips the last line of the loop on the interfaces that creates the interface in libvirt.

This PR fixes that problem by adding the missing instructions before the `continue`. It is been tested to work.

This is my first contribution to terraform-provider-libvirt and I'm a n00b in go language, so please double-check the fix with attention. Thank you.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -1283,55 +1283,51 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
 				return fmt.Errorf("Error retrieving network name: %s", err)
 			}
 			networkDef, err := newDefNetworkfromLibvirt(network)
-			if !HasDHCP(networkDef) {
-				netIface.Type = "network"
-				netIface.Source = &libvirtxml.DomainInterfaceSource{
-					Network: networkName,
+
+			// only for DHCP, we update the host table of the network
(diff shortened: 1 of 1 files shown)


Bischoff:
@MalloZup Sorry, no, I will not write a test case, at least not now. It looks complicated, and I have too little time right now.

But feel free to create a test case yourself. English description is: "declare two things:
* a private network with no DHCP (for example in mode "none" and no address range declared)
* a libvirt instance with at least one interface connected to that network
and check the provider creates that card."

Documentation how to create a network is here: https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/network.markdown ; documentation how to declare an interface is here: https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/domain.html.markdown#handling-network-interfaces .

If you want that we create it together, no worries, but I won't write one from scratch, especially with my current level in go.

dmacvicar:
*update* discussed with Eric in person and we agree to refactor it so that the host table is only updated for the DHCP case, which prevents the bug of the skipped "append" in the first place.

Bischoff:
Thanks a lot Duncan.

PR #288, opened 2018-03-13
[WIP] More configuration options for networks

inercia:
This PR does some cleanups in the `libvirt_network` block, adds some sub-blocks like `dns {}`, `dhcp {}` and includes new attributes like `routes = []`

PR #321, opened 2018-07-11
Decouple definition of subnet addresses from definition of DHCP range

Bischoff:
### What does this PR do? ###
If no addresses are declared for a network, then the bridge to the network does not get an IP address on the real host, making the network (and the VMs using it) unreacheable from the real host.

If addresses are declared, then a DHCP server is started on the network, but this is not always wished.

This PR adds a new `dhcp` boolean flag, that lets someone, in conjunction with existing `addresses` list, set up an address for the bridge while not starting a DHCP server. This flag defaults to `true`, thus keeping compatibility with existing `main.tf` files.

### Example ###
This setup:
```
(...)
resource "libvirt_network" "private_network" {
  name = "mynetwork"
  mode = "none"
  addresses = [ "192.168.75.0/24" ]
  dhcp = false
}
(...)
```
results in:
```
# virsh net-dumpxml mynetwork
<network connections='1'>
  <name>mynetwork</name>
  <uuid>10de3502-198d-482e-8034-5cc1950905d3</uuid>
  <bridge name='virbr2' stp='on' delay='0'/>
  <mac address='52:54:00:86:59:bd'/>
  <ip family='ipv4' address='192.168.75.1' prefix='24'>
  </ip>
</network>
```

Please notice the absence of unwanted:
```
    <dhcp>
      <range start='192.168.75.2' end='192.168.75.254'/>
    </dhcp>
```

The bridge gets an IP address as expected:
```
# ip address show
64: virbr2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 52:54:00:86:59:bd brd ff:ff:ff:ff:ff:ff
    inet 192.168.75.1/24 brd 192.168.75.255 scope global virbr2
       valid_lft forever preferred_lft forever
```

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -229,8 +229,10 @@ func getDomainInterfacesFromNetworks(domain libvirtxml.Domain,
 		networkDef, err := newDefNetworkfromLibvirt(network)
 		macAddresses := make(map[string][]string)
 		for _, ips := range networkDef.IPs {
-			for _, dhcpHost := range ips.DHCP.Hosts {
-				macAddresses[dhcpHost.MAC] = append(macAddresses[dhcpHost.MAC], dhcpHost.IP)
+			if ips.DHCP != nil {
+				for _, dhcpHost := range ips.DHCP.Hosts {
+					macAddresses[dhcpHost.MAC] = append(macAddresses[dhcpHost.MAC], dhcpHost.IP)
+				}
 			}
 		}
 		networkMacAddresses[networkName] = macAddresses
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -75,6 +75,12 @@ func resourceLibvirtNetwork() *schema.Resource {
 					Type: schema.TypeString,
 				},
 			},
+			"enable_dhcp": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
+				Default:  true,
+			},
 			"autostart": {
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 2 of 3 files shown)


dmacvicar:
So what is the equivalent of `dhcp=X` in the libvirt configuration? Is it also a boolean?

I think the problem is that we are adding dhcp as default. However, the dhcp configuration in libvirt is not done through boolean flags, and this is my main concern.

Bischoff:
@dmacvicar Well, the libvirt provider does not mimic at all the libvirt XML syntax. Instead it makes it easier by computing a lot of things from reasonable defaults and avoids redundancy.

Compare:
```
  <ip family='ipv4' address='192.168.75.1' prefix='24'>
    <dhcp>
      <range start='192.168.75.2' end='192.168.75.254'/>
    </dhcp>
  </ip>
```
with:
```
  addresses = [  "192.168.75.0/24" ]
  dhcp = true
```

If you are not satisfied with this anymore and want to mimic libvirt, then a) it's a major refactoring and b) you will lose on conciseness and consistency.

Bischoff:
I just had a look at AWS and OpenStack providers, for comparaison. I do not really understand the AWS approach. For Openstack one reads on http://www.terraform.io/docs/providers/openstack/r/networking_subnet_v2.html#enable_dhcp

  "enable_dhcp - (Optional) The administrative state of the network. Acceptable values are "true" and "false". 
   Changing this value enables or disables the DHCP capabilities of the existing subnet. Defaults to true."

I will rename `dhcp` to `enable_dhcp` immediately.

Bischoff:
To make things explicit, here is what a 1:1 equivalence with libvirt could look like
```
subnets = [
   { host_address = "192.168.75.1", dhcp = { start = "192.168.75.2", end = "192.168.75.254" } },
   { host address = "fde0::5:1", dhcp = { start = "fde0::5:2", end = "fde0::5:fe" } }
]
```
We would gain more configuration possibilities, but lose a lot of comfort.

inercia:
I think that maybe this should be part of #288, where _disable-dhcp_ part is already implemented...

inercia:
> Sure, if you want to make it on a wider scale, as discussed in #323, no problem on my side. @inercia, when will #288 be ready?

@Bischoff Mostly complete, I'm trying to fix the tests right now...

Bischoff:
ping @inercia - any progress ?

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/pull/385

MalloZup:
@Bischoff so the pr #385 is adreessing this PR in more generic way.

I took the intial draft from @inercia and improved it . 

This will work for sumaform creating the network.  ( i need to fix small issues but the main functionality is already working). feel free to test it with sumaform TIA :rose:

MalloZup:
@Bischoff  i am closing this since i backported your changes to that PR.

Thank you for your efforts :sun_with_face:

Bischoff:
Sure. Thanks for your merging efforts.

Issue #320, opened 2018-07-11
Update dependencies


dmacvicar:
Originally reported [here](https://github.com/dmacvicar/terraform-provider-libvirt/issues/295)

Updating dependencies fails:

```
ok 12:35 root@dc00 ~/go/src/github.com/terraform-provider-libvirt 
# make build
bash travis/run-gofmt
==> Checking that code complies with gofmt requirements...
*** checking pkg libvirt with gofmt
 pkg libvirt has no lint gofmt errors!

*** running gofmt on main.go
==> go fmt OK !!! ***
golint ./libvirt
go vet .
go build
# github.com/terraform-provider-libvirt/libvirt
libvirt/disk_def.go:14:7: unknown field 'Type' in struct literal of type libvirtxml.DomainDisk
libvirt/domain.go:199:24: cannot use networkInterface.Source.Network (type *libvirtxml.DomainInterfaceSourceNetwork) as type string in append
libvirt/domain_def.go:13:7: unknown field 'Type' in struct literal of type libvirtxml.DomainFilesystem
libvirt/domain_def.go:40:10: unknown field 'Type' in struct literal of type libvirtxml.DomainGraphic
libvirt/domain_def.go:41:14: unknown field 'AutoPort' in struct literal of type libvirtxml.DomainGraphic
libvirt/domain_def.go:46:10: unknown field 'Type' in struct literal of type libvirtxml.DomainChannel
libvirt/domain_def.go:48:11: unknown field 'Type' in struct literal of type libvirtxml.DomainChannelTarget
libvirt/domain_def.go:49:11: unknown field 'Name' in struct literal of type libvirtxml.DomainChannelTarget
libvirt/domain_def.go:57:12: unknown field 'Model' in struct literal of type libvirtxml.DomainRNGBackend
libvirt/resource_libvirt_domain.go:668:50: domainDef.OS.KernelArgs undefined (type *libvirtxml.DomainOS has no field or method KernelArgs)
libvirt/resource_libvirt_domain.go:668:50: too many errors
make: *** [Makefile:4: build] Error 2

fail 12:35 root@dc00 ~/go/src/github.com/terraform-provider-libvirt 
#
```

dmacvicar:
I already started working on this one. I figured it out, and I am doing progress.

dmacvicar:
I have completed the port of the code, but as I am moving on running/porting tests, I realized the first compatibility victim:

[...]

PR #322, opened 2018-07-11
Update libvirtxml and other dependencies

dmacvicar:
This PR refactors all the code building XML structures to match the upstream [libvirtxml](https://github.com/libvirt/libvirt-go-xml) refactorings and improvements.

All other dependencies are also updated through a generic `glide update`

Some notes for the reviewers:

* 2 testcases are still failing as the ignition provider used in the test is no longer compatible. We need to decide what to do here.
* One additional fix popped up during the refactoring: https://github.com/dmacvicar/terraform-provider-libvirt/commit/7d60f0dcd06508c886052cf7bf35720dc89e50e9

Addresses https://github.com/dmacvicar/terraform-provider-libvirt/issues/320


libvirt/resource_libvirt_domain.go
```diff
@@ -967,31 +965,58 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
...
-			domainDef.Devices.Graphics[0].AutoPort = "no"
+			listenType = "none"
 		}
-		if listenType, ok := d.GetOk(prefix + ".listen_type"); ok {
-			domainDef.Devices.Graphics[0].Listeners = []libvirtxml.DomainGraphicListener{
-				{
-					Type: listenType.(string),
-				},
+
+		switch graphicsType {
```
MalloZup: :+1:

dmacvicar:
I think I found a way to get the ignition test working. It was not related to this PR (we are using terraform 0.11.x before this PR), but the way that glide was ran (using `--strip-vendor`), so pinning the right versions and using `--strip-vendor` should make our provider and the ignition provider share the same vendor libraries. It should also reduce the amount of duplicated vendor files, as they would be flattened. I will update this PR.

To keep git repo size down, I will squash the new commits.

Issue #290, opened 2018-03-25
Support multiple Provider instances with different URIs


kawaja:
#  Version Reports:

### Distro  version of host:

``` Ubuntu 16.04.4 LTS ``` (xenial)

### Terraform Version Report

Terraform v0.11.6-dev (725281f4aa8c166d5d8bd003c2d8f8f4b4bcbc46+CHANGES)

### Libvirt version

On Terraform host:
```# virsh --version
1.3.1
```

On Libvirt host(s):
```# virsh --version
3.2.0
```

### terraform-provider-libvirt plugin version (git-hash)

```# git log
d37e941778af77694f63306d4cae3250885abd87
```
___
# Description of Issue/Question

I want to be able to use a single terraform config to address multiple libvirt hosts, in order to build a cluster of VMs deployed on these different hosts.

### Setup

After reading through the Terraform documentation, I wrote something like this. In `cluster.tf`:

```
provider "libvirt" {
  alias = "host00"
  uri = "qemu+ssh://root@192.168.100.1/system"
}

provider "libvirt" {
  alias = "host01"
  uri = "qemu+ssh://root@192.168.100.2/system"
}

...

module km1 {
  source "./domain"
  host "km1"
  ram "2048"
  providers {
    libvirt = "libvirt.host00"
  }
}

module k01 {
  source "./domain"
  host "k01"
  ram "4096"
  providers {
    libvirt = "libvirt.host01"
  }
}

module k02 {
  source "./domain"
  host "k02"
  ram "4096"
  providers {
    libvirt = "libvirt.host00"
  }
}
```

and then in `domain/main.tf`:

```
variable "host" {
  type = "string"
}

variable "ram" {
  type = "string"
}

...

# this is a temporary workaround required in the current version of terraform
# and should be removed in the future (see https://www.terraform.io/docs/modules/usage.html)
provider "libvirt" {
}

...

resource "libvirt_domain" "host" {
  depends_on = ["null_resource.fix-qcow2-image"]
  name = "${var.host}"
  memory = "${var.ram}"
  vcpu = 1

...
}
```

When I `apply` this, however, it starts all domains on one libvirt host, rather than spreading them across both hosts. In looking at the `terraform-provider-libvirt` code, this is because the first connection is cached in a global variable when the first Provider instance is created and the subsequent Provider instance uses this cached value rather than creating the second connection.

[...]

PR #325, opened 2018-07-17
Support multiple provider instances with different URIs (#290)

dmacvicar:
Terraform supports this using
https://www.terraform.io/docs/configuration/providers.html#multiple-provider-instances

The current code had global connection and mutex instances that prevented this from
working correctly.

While the global objects and mutexes are gone with this PR, we still reuse the client per URI, in order to get the same mutex for the same hypervisor, and then close all opened connections at the end.

All global data is moved into the Client instance.

Fixes: https://github.com/dmacvicar/terraform-provider-libvirt/issues/290

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ $ terraform destroy
 
 Look at more advanced examples [here](examples/)
 
+### Using multiple hypervisors / provider instances
+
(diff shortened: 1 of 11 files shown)


MalloZup:
@dmacvicar  should we disable the coverage tool? since we plan to disable travis-ci i don't see any advantage to have it

dmacvicar:
I increased the threshold to 5%

dmacvicar:
@kawaja let me know if you can give this a test

inercia:
Related to #282

PR #328, opened 2018-07-18
examples: Fix 'autoport' value from string to boolean

hwoarang:
The schema was changed to require a boolean value for the 'autoport'
argument. Fixes the following error:

Error: libvirt_domain.domain-ubuntu: graphics.0.autoport: cannot parse '' as bool: strconv.ParseBool: parsing "yes": invalid syntax


MalloZup:
thx @hwoarang

PR #333, opened 2018-07-24
skip qemu-agent when running make testacc

MalloZup:
since qemu-agent is causing failures on testacc and also because is faster in this way.


MalloZup:
merged in ohter pr

PR #348, opened 2018-07-31
Fix spurious reading network_name when using bridge as the source

dmacvicar:
In the past, Source contained Network and Bridge as a string. Now because the Source is exclusive, it is an union of both.

There is no apparent reason why one would have a Network name as a source when
using Bridge as the source.


Bischoff:
Thumbs up: tested and works

Issue #352, opened 2018-08-09
Implement out-of-band destroy of resources


dmacvicar:
[Documentation](https://www.terraform.io/docs/extend/writing-custom-providers.html) is very clear that resources can be destroyed out-of-band and that the provider has to handle that gracefully.

Right now, if the expected resource is not found via libvirt, the providers errors. Some additional logic has to be implemented, like checking the specific error (DOMAIN_NOT_FOUND), same with volumes.

I plan to implement this.

Issue #324, opened 2018-07-17
cloud-init volume race conditions


MalloZup:
when the volume is created, it is not removed on the event of a failure, so the next time you will find the volume being referenced in the state file, but not phisically present in the pool...
```
Error: Error refreshing state: 2 error(s) occurred:

* libvirt_cloudinit.admin: 1 error(s) occurred:

* libvirt_cloudinit.admin: libvirt_cloudinit.admin: error while retrieving remote ISO: Can't retrieve volume /var/images/admin_cloud_init.iso
* libvirt_volume.img: 1 error(s) occurred:

* libvirt_volume.img: libvirt_volume.img: can't retrieve volume /var/images/devel
```

MalloZup:
https://github.com/kubic-project/automation/issues/367

MalloZup:
i will investigate on this on freecycles

dmacvicar:
Can we go&describe over the steps of creation that may fail in between? then we can define a way to make it more atomic.

What kind of failures are you _seeing_ during creation?

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/commit/d6ff0ccd1fa160be716175dae88f10837917c1c1

gcavalcante8808:
@MalloZup can u provide some information about the discussion you had about libvirt?

I have some issues with racing conditions when I try to use terraform ansible modules from nbering with the libvirt module. If I create the hosts without the "ansible_host" resource and then add the configuration, it works flawlessly.

If I try to use a complete TF file with domain and ansible configurations, terraform returns the following error:

* ansible_host.etcd: Resource 'libvirt_domain.etcd' does not have attribute 'network_interface.0.addresses.0' for variable 'libvirt_domain.etcd.*.network_interface.0.addresses.0'

TF File:
```
variable "number_of_etcds" {
  default = "1"
}

resource "libvirt_volume" "etcd-vol" {
  count ="${var.number_of_etcds}"
  name = "etcd_vol-${count.index}"
  base_volume_id = "${libvirt_volume.centos-qcow2-latest.id}"
}

resource "libvirt_domain" "etcd" {
  count ="${var.number_of_etcds}"
  name = "etcd-${count.index}"
  memory = "512"
  vcpu = 1

  cloudinit = "${libvirt_cloudinit.commoninit-latest.id}"

  network_interface {
    network_name = "default"
    wait_for_lease = 1
  }

  console {
    type        = "pty"
[...]

Issue #88, opened 2016-11-28
libvirt_volume.main_disk: Can't retrieve volume xxx


MalloZup:
## How-to-reproduce:

clone sumaform
Take the main.tf-libvirt-testing example.
```console
Terraform get
Terraform apply
```
run it 3/4 times.  ( 2 times will fail with error : 
```
libvirt_volume.main_disk: Can't retrieve volume  sles12sp1
```
This issue block the provissioning and for production or dayly work is annoyng, because we have to reapply terraform apply.

I cannot sadly determine the cause of this issue, the disk created is there but terraform fails to retrieve it. 
It's a race condition. I'm quite sure that the sumaform is not the root cause of the problem.

For additional logs, info, ping me

MalloZup:
Doing 2 times terraform apply workarounds the problem

MalloZup:
@inercia  @moio 
here we have the log with debug enabled:

https://gist.github.com/MalloZup/0413641e9f350d22f225f0982f577fef

Looking at this debug, to me it looks like the issue is that terraform want to use the volume, but the volume is not there, and the plugin says "it's created".


cloning ```https://github.com/moio/sumaform```
Here what i used  as tf.file

```console 
provider "libvirt" {
  uri = "qemu+tcp://myREMOTE_SERVER_Desperados.server.com/system"
}

module "base" {
  source = "./modules/libvirt/base"

  cc_username = "UC7"
  cc_password = "Desperados"

  pool = "ssd"
  network_name = ""
  bridge = "br0"
  name_prefix = "dg"
  use_avahi = false
  domain = "mgr.suse.de"
  images = ["sles12sp2"]
}

module "vanilla" {
  source = "./modules/libvirt/host"
  base_configuration = "${module.base.configuration}"
  name = "vanilla"
  image = "sles12sp2"
}
```

So here we create the disks, globally, https://github.com/moio/sumaform/blob/master/modules/laibvirt/base/main.tf#L5


then for the host we use them here:

https://github.com/moio/sumaform/blob/master/modules/libvirt/host/main.tf

MalloZup:
this is a invalid issue, due to uncorrect use of the plugin. thx for support, closing

moio:
> this is a invalid issue, due to uncorrect use of the plugin. thx for support, closing

Why? Did you find a workaround? I now have the same problem.

MalloZup:
@moio  was at the time we did not lock resource esplictely via hcl
[...]

PR #353, opened 2018-08-10
Handle gracefully out-of-band destruction of volume and cloud-init resources

dmacvicar:
Terraform states in the "Writing providers guide":

>If the ID is updated to blank, this tells Terraform the resource no longer exists (maybe it was destroyed out of band). Just like the destroy callback, the Read function should gracefully handle this case.

One can as well (according to my experiment and the added tests results), implement the `Exists` callback, which will make terraform first check and then call `Read`. This is the reason why adidng graceful handling of out-of-band destruction to `Domain` had no effect: it already implements `Exists`.

This should improve the user experience by preventing the user getting errors when state is already created by terraform, but some of the resource were changed behind terraform's back.

The PR includes some commits that refactor cloud init to be more like the other resources in term of file-names, function names and reuse more of volume resource code.

Addresses #352 and probably (no evidence) related ones like #324 and #88


multani:
YES please! :)

I implemented by trial and error [something for the same problems](https://github.com/dmacvicar/terraform-provider-libvirt/commit/b2815e56105e79311cf6e735536bec44e652aeb2), which I never reported upstream due to the lack of unit tests :(

I wasn't aware of `Exists` so I didn't try, will it bypass `Read` if `Exists` returns False?

In any case, you could have a look at my (3 lines) patch as you are setting ID to `""` in the volume but not for cloud-init. I'm not sure if it's necessary or not with `Exists` though :man_shrugging:

dmacvicar:
@multani Originally I implemented handling the case of removed resources on `Read`. Then I realized if you implement `Exists` terraform uses it in a way that it gracefully handles out-of-band destruction by itself. I realized because I implemented a testcase for `libvirt_domain` resource that I was expecting to see fail, and it didn't.

multani:
@dmacvicar Oh I see. That's even better then!

PR #360, opened 2018-08-19
add initrd del. in make clean target

MalloZup:
cleanup-initrd volume in utility script


MalloZup:
merging because cosmetical but usefull to me :sunflower:

PR #366, opened 2018-08-21
 remove unused/deprecated function remoteImageSize

MalloZup:
this function was merged in utils.go during refactoring.
The old function was merged here:
https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/utils_volume.go#L76

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -4,8 +4,6 @@ import (
 	"encoding/xml"
 	"fmt"
 	"log"
-	"net/http"
-	"strconv"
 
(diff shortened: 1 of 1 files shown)


libvirt/resource_libvirt_volume.go
```diff
@@ -67,18 +65,6 @@ func resourceLibvirtVolume() *schema.Resource {
 	}
 }
 
-func remoteImageSize(url string) (int, error) {
```
MalloZup: https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/utils_volume.go#L76

PR #373, opened 2018-08-24
Revert "Do not use qemu-agent by default to gather network information"

MalloZup:
Reverts dmacvicar/terraform-provider-libvirt#354

# problem

Since this merge the bridge mode doesn't work anymore 

as i discussed with @inercia  in gitter, we have  a problem with this when using `bridge` mode.

I just test it before migrating sumaform to tf11. ( we use bridge in prod)

# what we can do for fixing pb:
- we could implement a function that enable qemu-agent when bridge is active.

To me this could be ok, but as we hit already the pb with bridge, the question is:

` how many networking configurations could require the enabling of qemu-agent`?

What i fear is that we have to much corner cases that would require the enablement in that particolar networking modes.
To me seems really difficult to maintain and that is why i would enable by default the qemu-agent and disable it if not needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,20 +120,6 @@ Look at more advanced examples [here](examples/)
 
 You can target different libvirt hosts instantiating the [provider multiple times](https://www.terraform.io/docs/configuration/providers.html#multiple-provider-instances). [Example](examples/multiple).
 
-### Using qemu-agent
-
-From its documentation, [qemu-agent](https://wiki.libvirt.org/page/Qemu_guest_agent):
-
->It is a daemon program running inside the domain which is supposed to help management applications with executing functions which need assistance of the guest OS.
-
-Until terraform-provider-libvirt 0.4.2, qemu-agent was used by default to get network configuration. However, if qemu-agent is not running, this creates a delay until connecting to it times-out.
-
-In current versions, we default to not to attempt connecting to it, and attempting to retrieve network interface information from the agent needs to be enabled explicitly with `TF_USE_QEMU_AGENT`. Note that you still need to make sure the agent is running in the OS, and that is unrelated to this option.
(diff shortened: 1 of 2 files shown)


moio:
For sumaform users at least reverting is OK, forcing developers to set an environment variable is not so good.

What about making this a configuration variable in the domain, where it belongs, instead of using a "global" environment variable?

MalloZup:
to me the main concern is the question: should we really disable by default the qemu-agent?
 if we have the qemu-agent enabled by demand for specific configurations, we will end-up maintaining and fixing the bugs that are related to it. Like the one is on sumaform with bridged mode. and i think there are other networking conf that led to this

To me we should disable it and if some user want more speed they can enable the flag without.

we are not gaining much speed disabling it afaik and is better to have more stability in longterm  has little more performance.

dmacvicar:
I will not revert this. Using the agent by default was annoying for everyone. I prefer in this case:

* Document that bridge mode needs the env variable set. This is not that terrible, as if you use the qemu agent, people need to be aware to run the daemon too, so bridge mode would work only on certain images running the daemon anyway. My problem is that even if you wanted to use it, and you lack the daemon, this leads to annoyances with the timeouts.
* As @moio mentions, make the usage part of the domain schema. Or even further:
  * given that we even misname the variable as "use the qemu agent", but we only use it for network information, it may belong in the network interface definition instead?
* In any case, this starts to look like moving screws and switches around. I will only change it further with a better overall proposal and overall picture.

MalloZup:
ok thx @dmacvicar  closing i opened a follow-up issue. 

Yop having in network schema afaik should be more appropriate then domain.
We might also need to research if there is also other type of networking mode that could need it

PR #375, opened 2018-08-26
Print err message instead of throwing it away

eparis:


MalloZup:
putting @dmacvicar as coreviewer just for notification

libvirt/coreos_ignition_def.go
```diff
@@ -168,7 +168,7 @@ func newIgnitionDefFromRemoteVol(virConn *libvirt.Connect, id string) (defIgniti
 
 	volume, err := virConn.LookupStorageVolByKey(key)
 	if err != nil {
-		return ign, fmt.Errorf("Can't retrieve volume %s", key)
+		return ign, fmt.Errorf("Can't retrieve volume %s: %v", key, err)
```
MalloZup: why we are using `%v` instead of `%s` ?
https://github.com/dmacvicar/terraform-provider-libvirt/blob/a73e75d277487024e79941644839a67002838aa5/libvirt/coreos_ignition_def.go#L86

dmacvicar: '%v' seems right. Not all 'error' are a string, except for the default one (`errorString`), as 'error' is an interface, and using '%v' may show additional fields. It is used like that in the examples at https://blog.golang.org/error-handling-and-go

PR #376, opened 2018-08-26
Add localonly domain option to libvirt-network

eparis:
This allows one to tell dnsmasq to own all of the domain in question.
Instead of forwarding unknown names. This way you can point your local
DNS at the libvirt DNS for the domain in question, and the domain in
question will not forward unknown hosts back to your local DNS.

### DIFF
--- a/libvirt/resource_libvirt_network_test.go
+++ b/libvirt/resource_libvirt_network_test.go
@@ -2,12 +2,159 @@ package libvirt
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/libvirt/libvirt-go"
+	"github.com/libvirt/libvirt-go-xml"
 )
+
+func getNetworkDef(s *terraform.State, name string) (*libvirtxml.Network, error) {
+	var network *libvirt.Network
+	rs, ok := s.RootModule().Resources[name]
+	if !ok {
+		return nil, fmt.Errorf("Not found: %s", name)
+	}
+	if rs.Primary.ID == "" {
+		return nil, fmt.Errorf("No libvirt network ID is set")
+	}
+	virConn := testAccProvider.Meta().(*Client).libvirt
+	network, err := virConn.LookupNetworkByUUIDString(rs.Primary.ID)
+	if err != nil {
+		return nil, err
+	}
+	networkDef, err := newDefNetworkfromLibvirt(network)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading libvirt network XML description: %s", err)
+	}
+	return &networkDef, nil
+}
+
+func TestAccCheckLibvirtNetwork_LocalOnly(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtNetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+				resource "libvirt_network" "test_net" {
+					name      = "networktest"
+					domain    = "k8s.local"
+					addresses = ["10.17.3.0/24"]
+					dns {
+						local_only = true
+					}
+				}`),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("libvirt_network.test_net", "dns.0.local_only", "true"),
+					checkLocalOnly("libvirt_network.test_net", true),
+				),
+			},
+		},
+	})
+}
+
+func checkLocalOnly(name string, expectLocalOnly bool) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		networkDef, err := getNetworkDef(s, name)
+		if err != nil {
+			return err
+		}
+		if expectLocalOnly {
+			if networkDef.Domain == nil || networkDef.Domain.LocalOnly != "yes" {
+				return fmt.Errorf("networkDef.Domain.LocalOnly is not true")
+			}
+		} else {
+			if networkDef.Domain != nil && networkDef.Domain.LocalOnly != "no" {
+				return fmt.Errorf("networkDef.Domain.LocalOnly is true")
+			}
+		}
+		return nil
+	}
+}
+
+func TestAccCheckLibvirtNetwork_DNSForwarders(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtNetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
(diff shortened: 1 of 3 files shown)


MalloZup:
@eparis  thank you for you pr !

We have already a big PR that is adressing this one.
However i really prefer this PR in term that is more smaller and compact change. Adding only 1 thing.

For the actual implementation take care between boolean and yes/no xml in libvirt.

Here:
https://github.com/dmacvicar/terraform-provider-libvirt/pull/288/files#diff-042e2374df248e4277a3506b065eb28aR110

and here:
https://github.com/dmacvicar/terraform-provider-libvirt/pull/288/files#diff-042e2374df248e4277a3506b065eb28aR232

We might add also some testacceptance for variable, take inspiration from : 
https://github.com/dmacvicar/terraform-provider-libvirt/pull/288/files#diff-620e6bdaa84869e53998f4dbbdf65204R51

Finallly i would add some documentation about  this var, here: 

https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/network.markdown

eparis:
Updated, based on the feedback.  I borrows heavily from #288 but still only adds this one new things (and moves dns_forwarder_

MalloZup:
@eparis ok thx for PR we will have a look ! :sunflower:

Bischoff:
I'm not sure. I also have a small PR #321 that tackles insufficiencies of libvirt-network.

Either we apply Inercia's big PR
OR
we apply the small PRs (this one and #321) and Inercia has to rebase

Given that #288 has been staled for months, I would prefer the second solution.

Bischoff:
I would suggest a name other that "localonly" for your new setting. My PR too does "local only" networks but for dhcp. Your PR seems to be centered on DNS. What about a name like "dont_forward" ?

eparis:
And now I openned #382 with more DNS extensions.  I'm a fan of merging the 3 small PRs rather than the 1 large PR...

MalloZup:
@eparis thx i will have a look asap :+1:

MalloZup:
@eparis thx for pr, we can wait for others pov. the only thing to me is the testing of the feature.

We are checking of the attributes beeing set, sofar so good, but it will be interesting to test the real mechanism of the feature behind it.

Atm i don't have an idea how to do it properly i will search a little, if you have some inspiration feel free, that would be great! :+1:

libvirt/resource_libvirt_network_test.go
```diff
@@ -48,6 +48,48 @@ func TestNetworkAutostart(t *testing.T) {
...
+		CheckDestroy: testAccCheckLibvirtNetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
+				resource "libvirt_network" "test_net" {
+					name      = "networktest"
+					mode      = "nat"
+					domain    = "k8s.local"
+					addresses = ["10.17.3.0/24"]
+					dns {
```
MalloZup: for this kind of testing we need to parse the networking xml definitio of the network and check if the xml entity is set correctly.

Then we can assume that it works correctly, i will post you an example soon, i"m dooing it for the dhcp.

eparis: any luck on a pointer to a real test?

MalloZup: @eparis  sorry for delay yop, look here: 

https://github.com/dmacvicar/terraform-provider-libvirt/pull/385/files#diff-620e6bdaa84869e53998f4dbbdf65204R63

I will help you more on this pr next week :rose: :sunrise:

eparis: That's awesome, I'll check it out!

eparis:
Hopefully I sufficiently ripped off your test code :) It seems to work.

libvirt/resource_libvirt_network_test.go
```diff
@@ -48,6 +50,130 @@ func TestNetworkAutostart(t *testing.T) {
...
+		return nil, err
+	}
+	networkDef, err := newDefNetworkfromLibvirt(network)
+	if err != nil {
+		return nil, fmt.Errorf("Error reading libvirt network XML description: %s", err)
+	}
+	return &networkDef, nil
+}
+
+func testAccCheckLibvirtNetworkLocalOnlyStatus(name string, LocalOnlyStatus bool) resource.TestCheckFunc {
```
MalloZup: `small nitpick` testAccCheckLibvirtNetworkLocalOnlyStatus should be called:


```ModuleTest_ + Testname``` 
``` TestAccCheckLibvirtNetwork_LocalOnlyStatus```

I know that this not documented so you coud not know it but we are trying to move all the testnames in this direction.

So when running Acceptance test we get more the debug context

File: libvirt/resource_libvirt_network_test.go [hunk 3]
```diff
@@ -48,6 +50,130 @@ func TestNetworkAutostart(t *testing.T) {
...
+		} else {
+			if networkDef.Domain != nil && networkDef.Domain.LocalOnly != "no" {
+				return fmt.Errorf("networkDef.Domain.LocalOnly is true")
+			}
+		}
+		return nil
+	}
+}
+
+func testAccCheckLibvirtNetworkDNSForwarders(name string, expected []libvirtxml.NetworkDNSForwarder) resource.TestCheckFunc {
```
MalloZup: nitpick: testAccCheckLibvirtNetworkDNSForwarders ->  `TestAccCheckLibvirtNetwork_DNSForwarders`

libvirt/resource_libvirt_network_test.go
```diff
@@ -48,6 +50,130 @@ func TestNetworkAutostart(t *testing.T) {
...
+				}`),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("libvirt_network.test_net", "dns.0.local_only", "true"),
+					testAccCheckLibvirtNetworkLocalOnlyStatus("libvirt_network.test_net", true),
+					resource.TestCheckResourceAttr("libvirt_network.test_net", "dns.0.forwarders.#", "3"),
+					resource.TestCheckResourceAttr("libvirt_network.test_net", "dns.0.forwarders.0.address", "8.8.8.8"),
+					resource.TestCheckResourceAttr("libvirt_network.test_net", "dns.0.forwarders.1.address", "10.10.0.67"),
+					resource.TestCheckResourceAttr("libvirt_network.test_net", "dns.0.forwarders.1.domain", "my.domain.com"),
+					resource.TestCheckResourceAttr("libvirt_network.test_net", "dns.0.forwarders.2.domain", "hello.com"),
+					testAccCheckLibvirtNetworkDNSForwarders("libvirt_network.test_net", []libvirtxml.NetworkDNSForwarder{
```
MalloZup: nice :+1:

Bischoff:
In the DHCP patch, we do:
```
dhcp { enabled = true }
```
in this patch we do:
```
bind { standalone = false }
```
Would it make sense to align to same keywords ?

MalloZup:
we can do the 
`dns { enabled = false }`

in another PR afaik and track it on separate issue, from my pov

Bischoff:
ok, makes sense

MalloZup:
@dmacvicar feek free to have a look, to me looks good

MalloZup:
:+1: :hugs: :tada:  thx flavio and @eparis

PR #377, opened 2018-08-27
use qemu-agent variable as terraform variable for domain

### LINKED ISSUE #374: bridge mode follow-up issue from #373 qemu-agent
follow-up issue from 
https://github.com/dmacvicar/terraform-provider-libvirt/pull/373

MalloZup:
fix #374

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.4.4 
+
+HIGHLIGHTS:
+
+* qemu-agent variable is not anymore env. variable, instead it is 
+  domain terraform variable
 ## 0.4.3 (August 14, 2018)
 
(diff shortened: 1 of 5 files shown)


MalloZup:
for sumaform the behaviour is invisible for users, see https://github.com/moio/sumaform/commit/740c5ab7ecd30910f38e9e7ad9e135b2299acacb

moio:
Note also that it is possible to emulate the old behavior since Terraform allows to set variables via environment variables, in case there is a actual need for those:

https://www.terraform.io/docs/configuration/environment-variables.html#tf_var_name

MalloZup:
@dmacvicar  feel free to review it, tia :sunflower:

Issue #358, opened 2018-08-14
could not open disk image /dev/vghost/vol1: Image is not in qcow2 format


mjiao:
#  Version Reports:

### Distro  version of host:

```Centos7 ```

### Terraform Version Report
(Provided by running `terraform -v`.)
0.11.7
### Libvirt version

```virsh --version```
3.9.0
### terraform-provider-libvirt plugin version (git-hash)

``` git log```
0.4.2
___
# Description of Issue/Question

### Setup
(Please provide the full **main.tf** file for reproducing the issue (Be sure to remove sensitive info)
```
variable "domain_name" {
  type = "string"
}

variable "hypervisor" {
  type = "string"
  default = "aaa.in"
}

provider "libvirt" {
    uri = "qemu+ssh://root@${var.hypervisor}/system"
}

resource "libvirt_volume" "base_vol" {
  name = "${var.domain_name}_base_vol"
  source = "/home/jenkins/output-images/centos7-base"
}

resource "libvirt_volume" "base_disk" {
  name = "${var.domain_name}_base_disk"
  base_volume_id = "${libvirt_volume.base_vol.id}"
}

resource "libvirt_volume" "vol1" {
  name = "vol1"
  # vghost is a lvm storage pool
  pool = "vghost"
  size = 10000000000
  format = "raw"
}

resource "libvirt_domain" "terraform_test" {
  name = "${var.domain_name}"
  disk {
    volume_id = "${libvirt_volume.base_disk.id}"
    scsi = "true"
  }
  disk {
    volume_id = "${libvirt_volume.vol1.id}"
  }
  vcpu = "1"
  memory = "2048"
  running = "true"
  network_interface = {
    bridge = "br1"
    hostname = "${var.domain_name}"
    # wait_for_lease = 1
  }
  autostart = "true"
}
```
### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
terraform init
terraform apply -var domain_name=test
___
# Additional Infos:

The storage pool used is based on lvm:
  pool = "vghost" 
It seems that it always uses format qcows in the disk scope in the domain definition.
```
Error: Error applying plan:

1 error(s) occurred:

2018/08/15 00:57:34 [DEBUG] plugin: waiting for all plugin processes to complete...
* libvirt_domain.terraform_test: 1 error(s) occurred:

[...]

PR #380, opened 2018-08-28
Automatic disk driver selection based on volume format and automatic volume format detection

dmacvicar:
This PR started as a goal to avoid issues like #358 to happen, which required re-work in various areas of disk and volume handling.

* The provider was using _file=_ to point to the image in the disk _XML_ definition, however, depending on the _pool_ type, one may need to use _block_ or something else.
  * This PR changes the way to refer to a volume in the disk _XML_ description to use disks of type _volume_, where the _libvirt_ _XML_ definition points to the _pool_ and _volume name_ and it is libvirt who does the right translation of that into _file, block, etc_ depending on the _pool_ type. So, instead of adding all the logic in the provider, we outsource it to libvirt.

* We used to hardcode the disk driver to use _qcow2_ as format unless we detect an _ISO_. This creates problems with _pool_ like _LVM_.
  * This PR changes the default disk driver to _RAW_ (however I may think to remove the default at all) and sets the driver explicitly based on the volume format as specified by libvirt (for existing volumes).
  * As we also support uploading images from remote urls and local files, we add a feature to detect _qcow2_ files based on the file header, so that we set the format at upload time, which makes then the driver selection work, because it is based on the volume format and nothing else.

* Because the volume _format_ attribute is optional, however, for uploaded images (`source=`) it is detected, we set this attribute to _computed_. It can still be explicitly set, however, if it is eg. set to raw for an image we are uploading to libvirt from a remote url, and the provider detect it is a qcow2 file, it will complain.

Before merging this, I would like some feedback from #358. I did some testing reproducing it and it was positive, but there may be corners here and there to still take into account.

In general, this makes the provider better on the path to support all pool types.


libvirt/resource_libvirt_domain_test.go
```diff
@@ -205,7 +261,24 @@ func TestAccLibvirtDomain_ScsiDisk(t *testing.T) {
...
 	var domain libvirt.Domain
-	u, err := url.Parse("http://download.opensuse.org/tumbleweed/iso/openSUSE-Tumbleweed-DVD-x86_64-Current.iso")
+
+	fws := fileWebServer{}
+	if err := fws.Start(); err != nil {
+		t.Fatal(err)
+	}
+	defer fws.Stop()
+
+	isoPath, err := filepath.Abs("testdata/tcl.iso")
```
MalloZup: nice catch

MalloZup:
to me seems really good, i want to check one thing only before commenting one line

PR #392, opened 2018-09-04
Tests: dhcp_test should be expected true..

MalloZup:
when enabled

this is just nitpick. I overlooked it when i was testing the update ( testing 2 steps in same tests)  function which we canot tests because we have some other issues.


MalloZup:
just merging because is cosmetical

PR #395, opened 2018-09-05
libvirt/resource_libvirt_network_test: reuse getNetworkDef

eparis:
This just simplifies and reuses some code in a helper function.


eparis:
@MalloZup This fixes head so it will build tests again.

PR #405, opened 2018-09-07
remove unused hash function

MalloZup:
this function was removed during a refactor but the commiter forgot to remove the function declaration and the related test

the concerned function was removed here:
https://github.com/dmacvicar/terraform-provider-libvirt/commit/f130f9848e61924bded600602a91822cf76e0959

anyway terraform have already an helper function for that so we don\t need it.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -1,8 +1,6 @@
 package libvirt
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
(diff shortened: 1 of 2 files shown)


MalloZup:
@dmacvicar @flavio jfyi  ( for this is not needed a review)

MalloZup:
`
PASS
coverage: 73.3% of statements
ok  	github.com/dmacvicar/terraform-provider-libvirt/libvirt	153.634s
`

PR #407, opened 2018-09-10
randomize volume names (part of idempotent testacc)

MalloZup:
we need to randomize also the name of the volume additional to resource name


MalloZup:
testacc ok

Issue #384, opened 2018-08-29
Research on better implementation of cloudinit 


MalloZup:
https://www.terraform.io/docs/providers/template/d/cloudinit_config.html

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/issues/313 and many other issues. Goal of this is to be spare code more generic

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/pull/267

multani:
One issue that is a bit painful with current implementation is that the user data is rewritten before being saved in the state: if the original content (the one written in the Terraform configuration) differs from the one saved in the state due to the rewritting, Terraform will decide to recreate the cloud-init resource and its dependencies.

To prevent this, we have to take care of writing in our Terraform configuration a user data field which will contain the exact same order/content (the best we could!) as the one which will be saved in the state. This is a bit counter intuitive (and bit of an annoyance at the same time :/)

I'm not sure the `libvirt_cloudinit` resource needs to provide the `local_hostname` and `ssh_authorized_key`: it's nice if you are only interested by these but as soon as you have a custom user data, this falls apart and it becomes difficult to manage correctly.

MalloZup:
Yop i agree, that was the motivator for this issue

MalloZup:
fixed ! thx

PR #410, opened 2018-09-11
redesign cloud_init resource to cloud_init_disk, extend fields etc.

### LINKED ISSUE #313: template_file user_data containing a tab: terraform plan forces new resource
#  Version Reports:

### Distro  version of host:

```
cat /etc/*release | grep PRETTY_NAME
PRETTY_NAME="Ubuntu 16.04.4 LTS"
```

### Terraform Version Report

```
terraform --version
Terraform v0.11.7
+ provider.libvirt (unversioned)
+ provider.template v1.0.0
```

### Libvirt version

```
virsh --version
1.3.1
```

```
echo $CGO_ENABLED
1
```

https://github.com/dmacvicar/terraform-provider-libvirt/commit/b11e5b79ff40dc8f03d762d92226fabe315955dc

___

This could possibly be a problem with `terraform` on not the `terraform-provider-libvirt`.

The default example https://github.com/dmacvicar/terraform-provider-libvirt/blob/21c507a4c44b510196b8a4a527aca921e9cd9b11/examples/ubuntu/ubuntu-example.tf is modified due to https://github.com/dmacvicar/terraform-provider-libvirt/issues/311, and uses `template_file`

```
cat ubuntu-example.tf
# instance the provider
provider "libvirt" {
  uri = "qemu:///system"

  template = "${file("${path.module}/cloud_init.cfg")}"
  #template = "${trimspace("${file("${path.module}/cloud_init.cfg")}")}"

}

# Use CloudInit to add our ssh-key to the instance
resource "libvirt_cloudinit" "commoninit" {
          name           = "commoninit.iso"
	  user_data          = "${data.template_file.user_data.rendered}"
          ssh_authorized_key = ""
        }

```

```
cat cloud_init.cfg 
#cloud-config
ssh_authorized_keys: []
write_files:
- content: |
    #I'm a	tab
```

```
TF_LOG=debug terraform apply -auto-approve
TF_LOG=debug terraform plan
...
  ~ update in-place
-/+ destroy and then create replacement

2018/06/29 19:50:38 [DEBUG] plugin: waiting for all plugin processes to complete...
-/+ libvirt_cloudinit.commoninit (new resource required)
      id:        "/var/lib/libvirt/images/commoninit.iso;5b36712e-d554-ab48-551a-fefc7cdd9f2b" => <computed> (forces new resource)
      name:      "commoninit.iso" => "commoninit.iso"
      pool:      "default" => "default"
      user_data: "#cloud-config\nssh_authorized_keys: []\nwrite_files:\n- content: \"#I'm a\\ttab\\n\"\n" => "#cloud-config\nssh_authorized_keys: []\nwrite_files:\n- content: |\n    #I'm a\ttab\n" (forces new resource)

  ~ libvirt_domain.domain-ubuntu
      cloudinit: "/var/lib/libvirt/images/commoninit.iso;5b36712e-d554-ab48-551a-fefc7cdd9f2b" => "${libvirt_cloudinit.commoninit.id}"


Plan: 1 to add, 1 to change, 1 to destroy.
```

Replace the tab in ` #I'm a	tab` by a space and the plan ends correctly `No changes. Infrastructure is up-to-date.`.

MalloZup:
## TODOS:

- [x] Remove todos in codebase

- [x] add some negative tests

- [x] documentation
- [x] split in readable commits

- [x] rename cloud_init into `cloud_init_disk`

fix #313 #384 

example:

```hcl

# instance the provider
provider "libvirt" {
  uri = "qemu:///system"
}

# We fetch the latest ubuntu release image from their mirrors
resource "libvirt_volume" "ubuntu-qcow2" {
  name = "ubuntu-qcow2"
  pool = "default"
  source = "https://cloud-images.ubuntu.com/releases/xenial/release/ubuntu-16.04-server-cloudimg-amd64-disk1.img"
  format = "qcow2"
}

# Use CloudInit to add our ssh-key to the instance
resource "libvirt_cloudinit" "test" {
          name           = "test.iso"
          user_data          = "${data.template_file.user_data.rendered}"
		network_config = <<EOF
		network:
		version: 2
		ethernets:
		eno1:
		dhcp4: true
		EOF
}


  template = "${file("${path.module}/cloud_init.cfg")}"
}

# Create the machine
resource "libvirt_domain" "domain-ubuntu" {
  name = "ubuntu-terraform"
  memory = "512"
  vcpu = 1

  cloudinit = "${libvirt_cloudinit.test.id}"

  # IMPORTANT
  # Ubuntu can hang if an isa-serial is not present at boot time.
  # If you find your CPU 100% and never is available this is why
  console {
    type        = "pty"
    target_port = "0"
    target_type = "serial"
  }

  console {
      type        = "pty"
      target_type = "virtio"
      target_port = "1"
  }

  disk {
       volume_id = "${libvirt_volume.ubuntu-qcow2.id}"
  }
  graphics {
    type = "spice"
    listen_type = "address"
    autoport = true
  }
}
```
cloudinit file
```
#cloud-config
ssh_pwauth: True
chpasswd:
  list: |
     root:linux
     cloud-user:atomic
  expire: False
```

### DIFF
--- a/examples/ubuntu/ubuntu-example.tf
+++ b/examples/ubuntu/ubuntu-example.tf
@@ -21,10 +21,19 @@ data "template_file" "user_data" {
   template = "${file("${path.module}/cloud_init.cfg")}"
 }
 
+
+data "template_file" "network_config" {
+  template = "${file("${path.module}/network_config.cfg")}"
+}
+
+# for more info about paramater check this out
+# https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/cloudinit.html.markdown
 # Use CloudInit to add our ssh-key to the instance
+# you can add also meta_data field
 resource "libvirt_cloudinit" "commoninit" {
           name           = "commoninit.iso"
           user_data          = "${data.template_file.user_data.rendered}"
+          network_config = "${data.template_file.network_config.rendered}"
 }
 
 # Create the machine
--- a/examples/ubuntu/ubuntu-example.tf
+++ b/examples/ubuntu/ubuntu-example.tf
@@ -17,12 +17,15 @@ resource "libvirt_network" "vm_network" {
    addresses = ["10.0.1.0/24"]
 }
 
+data "template_file" "user_data" {
+  template = "${file("${path.module}/cloud_init.cfg")}"
+}
+
 # Use CloudInit to add our ssh-key to the instance
 resource "libvirt_cloudinit" "commoninit" {
           name           = "commoninit.iso"
-          ssh_authorized_key = "<ssh-key-here>"
-        }
-
+          user_data          = "${data.template_file.user_data.rendered}"
+}
 
 # Create the machine
 resource "libvirt_domain" "domain-ubuntu" {
--- a/examples/ubuntu/ubuntu-example.tf
+++ b/examples/ubuntu/ubuntu-example.tf
@@ -30,7 +30,7 @@ data "template_file" "network_config" {
 # https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/cloudinit.html.markdown
 # Use CloudInit to add our ssh-key to the instance
 # you can add also meta_data field
-resource "libvirt_cloudinit" "commoninit" {
+resource "libvirt_cloudinit_disk" "commoninit" {
           name           = "commoninit.iso"
           user_data          = "${data.template_file.user_data.rendered}"
           network_config = "${data.template_file.network_config.rendered}"
@@ -42,7 +42,7 @@ resource "libvirt_domain" "domain-ubuntu" {
   memory = "512"
   vcpu = 1
 
-  cloudinit = "${libvirt_cloudinit.commoninit.id}"
+  cloudinit = "${libvirt_cloudinit_disk.commoninit.id}"
 
   network_interface {
     network_name = "vm_network"
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -28,7 +28,6 @@ type defCloudInit struct {
 	NetworkConfig string `yaml:"network_config"`
 }
 
-// TODO: check better this one maybe do in otherplaces
 func newCloudInitDef() defCloudInit {
 	return defCloudInit{}
 }
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -10,46 +10,27 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"github.com/hooklift/iso9660"
-	"github.com/imdario/mergo"
 	libvirt "github.com/libvirt/libvirt-go"
 	"github.com/mitchellh/packer/common/uuid"
-	"gopkg.in/yaml.v2"
 )
 
-// userData is the filename expected by cloud-init
-const userData string = "user-data"
-
-// metaData is the filename expected by cloud-init
-const metaData string = "meta-data"
-
-type defCloudInitUserData struct {
-	SSHAuthorizedKeys []string `yaml:"ssh_authorized_keys"`
-}
-
-type defCloudInitMetaData struct {
-	LocalHostname string `yaml:"local-hostname,omitempty"`
-	InstanceID    string `yaml:"instance-id"`
-}
+const userDataFileName string = "user-data"
+const metaDataFileName string = "meta-data"
+const networkConfigFileName string = "network-config"
 
 type defCloudInit struct {
-	Name        string
-	PoolName    string
-	MetaData    defCloudInitMetaData
-	UserDataRaw string `yaml:"user_data"`
-	UserData    defCloudInitUserData
+	Name          string
+	PoolName      string
+	MetaData      string `yaml:"meta_data"`
+	UserData      string `yaml:"user_data"`
+	NetworkConfig string `yaml:"network_config"`
 }
 
-// Creates a new cloudinit with the defaults
-// the provider uses
+// TODO: check better this one maybe do in otherplaces
 func newCloudInitDef() defCloudInit {
-	return defCloudInit{
-		MetaData: defCloudInitMetaData{
-			InstanceID: fmt.Sprintf("created-at-%s", time.Now().String()),
-		},
-	}
+	return defCloudInit{}
 }
 
 // Create a ISO file based on the contents of the CloudInit instance and
@@ -159,8 +140,9 @@ func (ci *defCloudInit) createISO() (string, error) {
 		"cidata",
 		"-joliet",
 		"-rock",
-		filepath.Join(tmpDir, userData),
-		filepath.Join(tmpDir, metaData))
+		filepath.Join(tmpDir, userDataFileName),
+		filepath.Join(tmpDir, metaDataFileName),
+		filepath.Join(tmpDir, networkConfigFileName))
 
 	log.Printf("About to execute cmd: %+v", cmd)
 	if err = cmd.Run(); err != nil {
@@ -171,8 +153,7 @@ func (ci *defCloudInit) createISO() (string, error) {
 	return isoDestination, nil
 }
 
-// Dumps the userdata and the metadata into two dedicated yaml files.
-// The files are created inside of a temporary directory
+// write user-data,  meta-data network-config in tmp files and dedicated directory
 // Returns a string containing the name of the temporary directory and an error
 // object
 func (ci *defCloudInit) createFiles() (string, error) {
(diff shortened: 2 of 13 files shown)


libvirt/cloudinit_def.go
```diff
@@ -183,16 +173,9 @@ func (ci *defCloudInit) createFiles() (string, error) {
...
-	// Create files required by ISO file
-	mergedUserData, err := mergeUserDataIntoUserDataRaw(ci.UserData, ci.UserDataRaw)
-	if err != nil {
-		return "", fmt.Errorf("Error merging UserData with UserDataRaw: %v", err)
-	}
-	userdata := fmt.Sprintf("#cloud-config\n%s", mergedUserData)
-
 	if err = ioutil.WriteFile(
-		filepath.Join(tmpDir, userData),
-		[]byte(userdata),
```
MalloZup: to me: this can be more simplyfied

MalloZup:
@dmacvicar @multani what do you think about this?

The main goal of this is to get rid of this custom values we manage our self.
Instead we delegate to cloud_init resources. 
( i did not also looked at this: https://www.terraform.io/docs/providers/template/d/cloudinit_config.html 
and also a custom_data source yet.)

The main problem, was here: 
https://github.com/dmacvicar/terraform-provider-libvirt/issues/384#issuecomment-419075667

From my pov, since we are still in a "beta" phase, we should change this.

I see only advantages on this PR, since we don't reinvent the wheel and the user can create their user_data fields with their values.

examples/ubuntu/ubuntu-example.tf
```diff
@@ -17,10 +17,16 @@ resource "libvirt_network" "vm_network" {
    addresses = ["10.0.1.0/24"]
 }
 
+data "template_file" "user_data" {
+  template = "${file("${path.module}/cloud_init.cfg")}"
+  #template = "${trimspace("${file("${path.module}/cloud_init.cfg")}")}"
```
MalloZup: need to be improved too

libvirt/cloudinit_def.go
```diff
@@ -13,33 +13,23 @@ import (
...
 const metaData string = "meta-data"
 
-type defCloudInitUserData struct {
-	SSHAuthorizedKeys []string `yaml:"ssh_authorized_keys"`
-}
-
 type defCloudInitMetaData struct {
-	LocalHostname string `yaml:"local-hostname,omitempty"`
-	InstanceID    string `yaml:"instance-id"`
+	InstanceID string `yaml:"instance-id"`
```
MalloZup: need to check this more

multani:
@MalloZup I would :tada: this 1000 times if possible!

It's going to break lot of deployments, but I think it's definitely worthwhile considering the simplification it brings at so many level :+1:

libvirt/cloudinit_def.go
multani: I would say these are implementation details of what cloud-init support or not.

I would favor the approach where user can set whatever they want here, hopefully something cloud-init support (adding links towards the relevant cloud-init documentation would be great I guess).
Having default values like today for the Instance ID is nice (although I guess the instance ID itself should be changed, I have something currently named like `created-at-2018-09-05 18:53:16.5935747 +0200 CEST m=+37.022498553` :D)

Also, might be good to have support [for the other cloud-init fields](https://git.launchpad.net/cloud-init/tree/cloudinit/sources/DataSourceNoCloud.py#n71)?

multani: Actually, I think you need to let the metadata field, but:

* make it a string so that user can populate it as they want + document that if it's being filled in by users, they have to conform to what cloud-init needs (apparently, only `instance-id` is required)
* if it's empty, generate a `instance-id` field like it is done today

Since `meta-data` will not be filled in automatically by libvirt, if a cloud-init datasource is provided by the Terraform provider, then I would say it's up to the user to set it correctly! :)

multani:
On Thu, 13 Sep 2018, at 16:24, Dario Maiocchi wrote:
> I was looking for vendor-data, is something that you use or ever heard
> about it? is this wort to add? 🤔 tia
I never heard nor used it, but it's in cloud-init code so, may be
worth adding?

multani:
@dmacvicar I think renaming to `libvirt_cloudinit_disk` makes its intent clearer :+1: 

However, I think there's a distinction between what [the built-in Terraform data](https://www.terraform.io/docs/providers/template/d/cloudinit_config.html) does and what `libvirt_cloudinit_disk` should do, that would mandate to keep `user_data` instead of `config` as you did:

* AFAIK, `template_cloudinit_config` is only concerned about providing the "user" part of what cloud-init reads, ie. the "user-data", as the rest is supposed to be injected by the cloud provider directly. For example for an instance on AWS EC2:

  - http://169.254.169.254/2009-04-04/user-data/ can be set by the user creating the instance (and thus by Terraform and  `template_cloudinit_config`)
  - http://169.254.169.254/2009-04-04/meta-data/ however is set directly by AWS and can't be set by the user. It's out of scope of `template_cloudinit_config`.

* As there no such thing as "the cloud provider" for Libvirt in essence, the "Terraform libvirt provider" should act as "the cloud provider" in the sense of what I described above, and should provide a way of setting `user-data` AND `meta-data`

* As `template_cloudinit_config` is not made for managing `user-data` AND `meta-data`, I would keep them separated and have separate fields in `libvirt_cloudinit_disk` for each of these values.

So I think I would just go like this:

```tf
  name      = "commoninit.iso"
  user_data = "${data.template_file.config.rendered}"

  meta_data = <<EOF
instance-id: foo-bar
EOF

  vendor-data = <<EOF
#something here
EOF
}
```

MalloZup:
the usefull doc i found was this one mostly : https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html#datasource-nocloud

dmacvicar:
Yes, I tend to confuse the data sources _Config Drive_ and _NoCloud_ by name. _NoCloud_ is indeed the datasource we are using.

Now, I understand _NoCloud_ only supports _user-data_ and _meta-data_, and _network-config_. Which one is the 4th field?

Then I'd agree to keep the fields named _user-data_ and _meta-data_, and _network-config, and renaming the provider _libvirt_cloudinit_disk_ or libvirt_cloudinit_nocloud_disk.

I am not sure if I understand your argument about _template_cloudinit_config_. But it does not affect the discussion but mostly the examples. I agree we need to keep the fields separate in the resource, as we need a way to know which part goes to which file.

The part I like about this design is that in theory one could still write a data provider for _user-data_ (eg. _libvirt_cloudinit_user_data_)that has all the fields separated in a "type-safe" way (_ssh_key_, etc) and then inject it as _.rendered_ in the _user_data_ field of the _libvirt_cloudinit_disk_. This proposal separates the concerns of the provider, leaving it only with the responsibility of building the disk and attaching it, and still leave open to use _template_file_, or read a file with _file()_, or even use the _http_ data source.

MalloZup:
The 4th  field is vendor, but i need to read more about it.
 https://git.launchpad.net/cloud-init/tree/cloudinit/sources/DataSourceNoCloud.py#n71

MalloZup:
I dont see much value on vendor data field. From my pov we can add it later, if needed. I will continue without it , if you think we need it i will add it a the end

dmacvicar:
> The 4th field is vendor, but i need to read more about it.
> https://git.launchpad.net/cloud-init/tree/cloudinit/sources/DataSourceNoCloud.py#n71. Imhi is not super needed but l will get more info on doc before
> 
> I dont see much value on vendor data field. From my pov we can add it later, if needed. I will continue without it , if you think we need it i will add it a the end

I understand _vendor-data_ as equivalent to _user-data_ in capabilities, only that provided by the cloud, and not by the admin. I don't see how it is useful except that you are simulating a real cloud (hence the name _nocloud_), so you setup your images to seed from the metadata service and with less priority from nocloud. In QA you test with nocloud, and simulate the vendor values.

MalloZup:
good news: i have tested the current branch and sofar everything works.

I could setup a ubuntu img with cloud-init and change password.

Another good news is that istance-id can be empty because it will be popoulated automatically by cloud-init:

https://github.com/cloud-init/cloud-init/blob/master/cloudinit/sources/DataSourceNoCloud.py#L40

( so we don't need to set it by default). :tada: 

I am trying to verify the networking and metadata effects more in details but sofar i think it works.

If you want to experiment with this experimental branch feel free. :+1: 

At moment i am also looking to improve little more the testacc we have but is not so easy because we don't want to boot a qcow upstream etc, this take to much, if you have also suggestion for that feel free

:+1: :thermometer: thx

MalloZup:
```hcl

# instance the provider
provider "libvirt" {
  uri = "qemu:///system"
}

# We fetch the latest ubuntu release image from their mirrors
resource "libvirt_volume" "ubuntu-qcow2" {
  name = "ubuntu-qcow2"
  pool = "default"
  source = "https://cloud-images.ubuntu.com/releases/xenial/release/ubuntu-16.04-server-cloudimg-amd64-disk1.img"
  format = "qcow2"
}

# Use CloudInit to add our ssh-key to the instance
resource "libvirt_cloudinit" "test" {
          name           = "test.iso"
          user_data          = "${data.template_file.user_data.rendered}"
		network_config = <<EOF
		network:
		version: 2
		ethernets:
		eno1:
		dhcp4: true
		EOF
}


  template = "${file("${path.module}/cloud_init.cfg")}"
}

# Create the machine
resource "libvirt_domain" "domain-ubuntu" {
  name = "ubuntu-terraform"
  memory = "512"
  vcpu = 1

  cloudinit = "${libvirt_cloudinit.test.id}"

  # IMPORTANT
  # Ubuntu can hang if an isa-serial is not present at boot time.
  # If you find your CPU 100% and never is available this is why
  console {
    type        = "pty"
    target_port = "0"
    target_type = "serial"
  }

  console {
      type        = "pty"
      target_type = "virtio"
      target_port = "1"
  }

  disk {
       volume_id = "${libvirt_volume.ubuntu-qcow2.id}"
  }
  graphics {
    type = "spice"
    listen_type = "address"
    autoport = true
  }
}
```
```

#cloud-config
ssh_pwauth: True
chpasswd:
  list: |
     root:linux
     cloud-user:atomic
  expire: False
#cloud-config
ssh_pwauth: True
chpasswd:
  list: |
     root:linux
     cloud-user:atomic
  expire: False
```

dmacvicar:
@MalloZup I would like to do a release during the weekend (0.4.4). Would you mind that we target this for a 0.5 together with my work on the resizable base volumes? (because I would need to re-do the cloudinit example for partition resize)

MalloZup:
@dmacvicar yop no problem, i would like to spend more time on this also :smile:, so i am perfect fine with 0.5.0. i created the changelog for the (0.4.4 https://github.com/dmacvicar/terraform-provider-libvirt/issues/412)
hope it helps and thx for feedback

libvirt/cloudinit_def_test.go
```diff
@@ -41,16 +30,15 @@ func TestCloudInitCreateFiles(t *testing.T) {
 		t.Errorf("Unexpected error %v", err)
 	}
 	defer os.RemoveAll(dir)
-
-	for _, file := range []string{userData, metaData} {
```
MalloZup: this variable were global shared between files and declared in other files. for the simple for loop is better to have them explicit as string

MalloZup:
ok i did also some manual  test everything  is ok. 

@dmacvicar note that i changed also the commits names following the schema on Contributing.

Thank you for feedback

Issue #369, opened 2018-08-22
Support Volume Resizing from Base Volume


BlamKiwi:
#  Description
You support using base images for volumes OR creating an empty volume of a particular size. It would be good if you also supported created a volume from a base, but also resize it to a larger capacity. [This could then be reconfigured by cloud-init to provide a larger root volume. ](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#grow-partitions)

I've come across this use case a couple of times where I want to resize the cloud image as part of infrastructure deployment. Especially when its not easy or appropriate to mount sub-volumes to get around the small base image capacity. This is common when the install step is larger than the final rootfs requirements. (Eg: GitLab VM, Jenkins VM)

# Example

```
resource "libvirt_volume" "ubuntu-qcow2" {
  name = "ubuntu-qcow2"
  pool = "default"
  source = "..."
  format = "qcow2"
}

# Create a 5GB root volume
resource "libvirt_volume" "rootfs" {
  name = "rootfs"
  base_volume_id = "..."
  size = "5120"
}

# Use CloudInit to add our ssh-key to the instance
resource "libvirt_cloudinit" "commoninit" {
          name           = "commoninit.iso"
          ssh_authorized_key = "<ssh-key-here>"
          growpart = "/"
        }
```

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/issues/357 same issue proposal afaik

MalloZup:
i think we can close this issue @BlamKiwi @dmacvicar

dmacvicar:
yeah, I haven't got much feedback on the PR, but lets close this and continue discussion in the PR.

canit00:
I understand there's a couple of work-arounds 1) of them being extending the image prior to deploying, but that's not ideal. Also, followed the suggestion above but this isn't working & would prefer to tooling to have this feature which would allow dynamic deployments.

```
resource "libvirt_volume" "os_image" {
  name	 = "${var.hostname}-rootfs.qcow2"
  pool	 = "libvirt"
  source = var.image
  format = "qcow2"
}

# Use default 40G rootfs
resource "libvirt_volume" "rootfs" {
  name	 = "${var.hostname}-rootfs.qcow2"
  pool	 = "libvirt"
  size   = "42949672960"
}

# Define KVM domain
resource "libvirt_domain" "compute" {
  name   = var.hostname
  memory = var.memoryMB
[...]

Issue #357, opened 2018-08-14
Unable to set disk size when deploying from cloud image


mattymo:
#  Version Reports:

### Distro  version of host:

```  Ubuntu.. ```

### Terraform Version Report
Terraform v0.11.7

### Libvirt version
1.3.1

### terraform-provider-libvirt plugin version (git-hash)

``` git log```
___
# Description of Issue/Question

How can you use a cloud image (of Ubuntu for example) and create a libvirt_volume larger than the default qcow2 size provided by the source image?

### Setup
(Please provide the full **main.tf** file for reproducing the issue (Be sure to remove sensitive info)

Relevant bit:
```
resource "libvirt_volume" "ubuntu-qcow2" {
  name = "ubuntu-qcow2"
  pool = "default" 
  source = "https://cloud-images.ubuntu.com/releases/xenial/release/ubuntu-16.04-server-cloudimg-amd64-disk1.img"
  format = "qcow2"
}

resource "libvirt_volume" "volume" {
  name = "volume-${count.index}"
  base_volume_id = "${libvirt_volume.ubuntu-qcow2.id}"
  size = 42949672960 # 40gb
  count = 3
}
```
### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
```
Error: Error applying plan:

3 error(s) occurred:

* libvirt_volume.volume[2]: 1 error(s) occurred:

* libvirt_volume.volume.2: 'size' needs to be specified if no 'source' or 'base_volume_id' is given
* libvirt_volume.volume[0]: 1 error(s) occurred:

* libvirt_volume.volume.0: 'size' needs to be specified if no 'source' or 'base_volume_id' is given
* libvirt_volume.volume[1]: 1 error(s) occurred:

* libvirt_volume.volume.1: 'size' needs to be specified if no 'source' or 'base_volume_id' is given
```
___
# Additional Infos:

Do you have SELinux or Apparmor/Firewall enabled? Some special configuration?
Have you tried to reproduce the issue without them enabled?

mattymo:
Workaround I have for now:
wget $CLOUD_IMAGE
qemu-img resize $CLOUD_IMAGE 40G
Put the path to image in my libvirt.tf

dmacvicar:
[...]

PR #416, opened 2018-09-14
Able to specifiy size when using backing volume (and able to resize partitions with cloud-init)

dmacvicar:
This PR allows to set the size of a volume larger than its backing volume, which allows to resize the partition then using cloud-init, and not be limited by the backing image.

Addresses #369 and #357

* ~~~Waiting for https://github.com/dmacvicar/terraform-provider-libvirt/pull/410 to be merged to re-do the example.~~~ 
* ~~~I may introduce an integration test, however, it is not trivial without a fully cloud-init enabled boot image.~~~ Not going to add tests here.


dmacvicar:
Can you elaborate more how do picture such test?

MalloZup:
@dmacvicar  yop ( i didn't have a terminal on we :smile: ) , i had in mind to check the size parameter but i had looked now with a terminal, it is already covered in other tests ( the resize tests we cannot test it without wasting resource and time for the cloudinit) :+1:

PR #420, opened 2018-09-17
improve/refactory directory file-layout architecture design.

MalloZup:
# what does this PR do? : 

This pr remove the ` libvirt/utils_libvirt.go` and ` libvirt/utils_libvirt_test.go` and move the network functions to network and domain to domain function files.

I want to keep it simple for not conflicting to other PRs.

move network helper function to network files and domain to specific domain function files.

Previous behaviour we had generic files names


libvirt/network_def.go
```diff
@@ -57,3 +59,43 @@ func newNetworkDef() libvirtxml.Network {
 		return d
 	}
 }
+
+func getHostXMLDesc(ip, mac, name string) string {
```
MalloZup: this function will be usefull when we will update the network ( see current limitations)

PR #426, opened 2018-09-20
Fix wait_for_lease boolean in test and doc

MalloZup:
`wait_for_lease` = true/false


MalloZup:
Merging since not crucial

PR #427, opened 2018-09-20
Remove deprecated getDomainInterfacesFromNetworks

### LINKED ISSUE #425: investigate on function getDomainInterfacesFromNetworks  ( if it is still needed)
# Infos

I was debugging for getting the network interface information from domain.

I noticed that i got always a `NIL` interface with  the following function:  ```getDomainInterfacesFromNetworks```

The function was added here: https://github.com/dmacvicar/terraform-provider-libvirt/commit/7b2adfc1f28d76399d63aaf50ff914268efbb87a

As we can see on the testacc codecoverage, the relavant part of the code which  populate the NetworkINterface  returned by the function return is never called.

The **red** is code never executed

screenshot from 2018-09-19 16-04-09

MalloZup:
# What does this PR?

So basically i found out that `getDomainInterfacesFromNetworks` function is never used. ( see testscode coverage) and i try it out manually. (see here https://github.com/dmacvicar/terraform-provider-libvirt/issues/425#issue-361766883, for more details)
This is since https://github.com/dmacvicar/terraform-provider-libvirt/commit/7b2adfc1f28d76399d63aaf50ff914268efbb87a

After analyzing a little, if you look at 
https://github.com/dmacvicar/terraform-provider-libvirt/blob/2150005da77ddc623992685a8a574c8c170c37e6/libvirt/domain.go#L244

```if mac == domMac {```
this is not correct instead should be

```if strings.ToUpper(mac) == domMac {```

Once you uppercase the mac the function will work.

## why i want still want  remove it:

I am still want to remove this function because since the merge it never worked.

Now even if we could fix it to make it working, i don't see any advantage for having it.

This function is also impacting considerable the performance of the plugin ( i noticed when doing the testacc tests were slow once i have fixed the function.)

Other then performance it add code complexity which in the network part we still have, so i would just remove it for simplifing the codebase.

For code complexity motivation i mean following:

when we do https://github.com/dmacvicar/terraform-provider-libvirt/blob/2150005da77ddc623992685a8a574c8c170c37e6/libvirt/domain.go#L150

we are basically trying in 3 ways to get the Interfaces. ( 1 is with the qemu-agent is ok)

Then we have the 2 which are doing the same thing,  so i get the impression that we are `trying` to  do things in order to work or getting the interface.

https://github.com/dmacvicar/terraform-provider-libvirt/blob/2150005da77ddc623992685a8a574c8c170c37e6/libvirt/domain.go#L170

Last argument for removal is that if until now we never experienced a bug without it, why we should keep  it ? :small_airplane: 

fix #425

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -26,7 +26,7 @@ const domWaitLeaseDone = "all-addresses-obtained"
 var errDomainInvalidState = errors.New("invalid state for domain")
 
 func domainWaitForLeases(domain *libvirt.Domain, waitForLeases []*libvirtxml.DomainInterface,
-	timeout time.Duration, domainDef libvirtxml.Domain, virConn *libvirt.Connect, rd *schema.ResourceData) error {
+	timeout time.Duration, rd *schema.ResourceData) error {
 	waitFunc := func() (interface{}, string, error) {
 
 		state, err := domainGetState(*domain)
@@ -46,7 +46,7 @@ func domainWaitForLeases(domain *libvirt.Domain, waitForLeases []*libvirtxml.Dom
 
 		// check we have IPs for all the interfaces we are waiting for
 		for _, iface := range waitForLeases {
-			found, ignore, err := domainIfaceHasAddress(*domain, *iface, domainDef, virConn, rd)
+			found, ignore, err := domainIfaceHasAddress(*domain, *iface, rd)
 			if err != nil {
 				return false, "", err
 			}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by MalloZup** [comment]
```
PASS
coverage: 74.0% of statements
```

MalloZup:
This was.the commit msg by @eamonnotoole `Add new method to get information on domain interfaces

We've found that, when executing destroy/create cycles on
a domain where the virtual network(s) to which it is attached
are persisted across these cycles then we get network failures
on recreation of the domain since the old entries for
the domain are still present in the virtual network(s). This issue
should (hopefully) be fixed by commit
a20d2a9.  We've also found that,
on occasion, we don't get complete information on a domain's
interfaces from ListAllInterfaceAddresses. This patch implements
a new method getDomainInterfacesFromNetworks which is called from
domainGetIfacesInfo after the qemu agent method has been used if
the qemu agent method doesn't return any information.  This new
method builds the interface information from information collected
from the networks themselv`

MalloZup:
As side comment, as.wrote in the desc, even if theyou function had purpose, the function always returned nil because of the uppercase check mismatch

MalloZup:
done opened a follow-up issues so we can work on it later

Issue #419, opened 2018-09-17
Unusable mouse cursor in Windows resources


unsubtleguy:
#  Version Reports:

### Distro  version of host:

``` Fedora/NA ```

### Terraform Version Report
``` Terraform v0.11.8``` 

### Libvirt version
``` 4.1.0``` 

### terraform-provider-libvirt plugin version (git-hash)
``` c46f0f88a5b74319af7d1f1b219f1a7412f215de``` 
___
# Description of Issue/Question

When using Windows guests with the libvirt provider, by default the mouse input is completely unusable. I've had to manually work around by adding an evtouch input device (which results in being able to interact with the machine and use a mouse).

The bug is known, and is a KVM/libvirt problem rather than the provider - however I'd like to be able to not do this manually, given the role of Terraform.

Tl;dr:
* Terraform libvirt provider does not allow customisation of input devices
* When using modern (>2008) Windows(TM) Operating systems, when using the generic ps/2 input device, input is borderline unusable
  * By this I mean with great effort, you can move the mouse pointer around and use the machine - but the users mouse pointer does not align with the VMs mouse pointer
* However, when an `EvTouch USB Graphics Tablet`  input device is added, the mouse is responsive rather than not synchronising with the users mouse pointer

I've made a quick modification locally which appeared to work - and with quick testing, it added the evtouch device to deployed resources.

This would be of great benefit to anyone using this provider for Windows guests - and does not appear to harm Linux guests.

### Proposal

Ideally, the provider would just allow you to modify this on a per-resource basis, but the following appeared to work as a proof of concept:
```
$ git diff .
diff --git a/libvirt/domain_def.go b/libvirt/domain_def.go
index 09a41717..7a3aaa2f 100644
--- a/libvirt/domain_def.go
+++ b/libvirt/domain_def.go
@@ -8,6 +8,8 @@ import (
        libvirtxml "github.com/libvirt/libvirt-go-xml"
 )

+var tabletBus uint = 0
+
 // note source and target are not initialized
 func newFilesystemDef() libvirtxml.DomainFilesystem {
        return libvirtxml.DomainFilesystem{
@@ -35,6 +37,22 @@ func newDomainDef() libvirtxml.Domain {
                },
                CPU: &libvirtxml.DomainCPU{},
                Devices: &libvirtxml.DomainDeviceList{
+                               Inputs: []libvirtxml.DomainInput{
+                                       libvirtxml.DomainInput{
[...]

Issue #350, opened 2018-08-04
Add Network Model Configuration


ncsurfus:
#  Version Reports:

### Distro  version of host:

``` Ubuntu```

### Terraform Version Report
Terraform v0.11.7

### Libvirt version

```1.3.1```

### terraform-provider-libvirt plugin version (git-hash)
0.4
``` git log```
___
# Description of Issue/Question

Add support for other network model besides virtio. 
In file: resource_libvirt_domain.go, function: setNetworkInterfaces, there is this line of code which hardcodes the network model to virtio. Ideally, virtio should be the default but it would be nice to be able to set this to other values to add support for a wider amount of devices. I modified this locally to e1000 to validate it added network support for my specific use case.
```
		netIface := libvirtxml.DomainInterface{
			Model: &libvirtxml.DomainInterfaceModel{
				Type: "virtio",
			},
		}
```

___
# Additional Infos:

N/A

**Comment by dmacvicar** [comment]
What is the use-case nowadays to not use `virtio`?

**Comment by ncsurfus** [comment]
My use case is for Cisco vIOS and other virtualized routers/switches. The network model support for them is generally poor since their physical counterparts utilize ASICs and these virtualized devices are often meant for a testing / learning environment. Previously, I’ve hand built virtualized network environments to validate code against network devices. My goal is to be able to stand up these environments as part of the testing process (rather than a pre-built environment that I hope never breaks). Besides having to manually change the network model type, I’ve had a ton of success using this project!

**Comment by axxyhtrx** [comment]
+1

**Comment by dmacvicar** [comment]
I am in favor of this, even if it is not in the top of the list of the things I would spend my time on.

Before even accepting more "options" that change the schema, I am toying with the idea to add some hooks to the resources that would allow to set arbitrary values in the _libvirt_ _xml_ definition, that would help everyone with corner cases or custom configurations.

I was thinking something around the lines of:

```hcl
libvirt_domain "foo" {
   # ....
   xml = {
      "//devices/nic[0]/model/@type" = "e1000"  
  }
}
```
As a hook that would set the given _xpath_ values before creating the actual resource and after the provider has its own version of the _xml_ description done.

We would solve a much broader array of corner cases, even if we may choose to standardize some of these in the schema later, people would not need to wait and will have a workaround.

Let me think about it for some more time.

MalloZup:
@ncsurfus we fixed it. Have look with latest release and doc on xslt.
[...]

PR #431, opened 2018-09-23
Initial prototype of a XML transformation hook based on XSLT

dmacvicar:
Sometimes the provider does not allow to set some libvirt attributes. Good examples are #350 (the provider does not allow to change the NIC model)  #419 (one needs to add a touch device).

This PR explores the idea of allowing for a transformation hook before creating the domain. The hook is in the form of a _XSLT_ stylesheet. This allow for almost unlimited capabilities when "editing" the generated XML.

```hcl
provider "libvirt" {
  uri = "qemu:///system"
}

resource "libvirt_domain" "xslt-demo-domain" {
  name = "xslt-demo-domain"
  memory = "512"

  network_interface {
    network_name = "default"
  }

  xml {
    xslt = "${file("nicmodel.xsl")}"
  }
}
```

_nicmodel.xsl_:

```xml
<?xml version="1.0" ?>
<xsl:stylesheet version="1.0"
                xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
  <xsl:output omit-xml-declaration="yes" indent="yes"/>
  <xsl:template match="node()|@*">
     <xsl:copy>
       <xsl:apply-templates select="node()|@*"/>
     </xsl:copy>
  </xsl:template>

  <xsl:template match="/domain/devices/interface[@type='network']/model/@type">
    <xsl:attribute name="type">
      <xsl:value-of select="'e1000'"/>
    </xsl:attribute>
  </xsl:template>

</xsl:stylesheet>
```

Some notes about the implementation:

* The current implementation depends on _xsltproc_ to do the processing
* It works well for properties that are not supported by the provider. Otherwise the provider will insist in changing them back. I have an acceptance test for both cases: changing the nic model and changing the network name.
* I am not sure if it is worth to have the `xslt`property nested into an `xml` element.
* Only implemented for domain resources, but 95% of the code is generic. A few lines are needed in every resource creation and schema to add support for it.

The idea is of this PR is to gather some feedback and foresee problems I may not be seeing yet.

//cc @ncsurfus @unsubtleguy

## TODO

- [X]  Gather feedback
- [ ] Decide on the element name
- [X] Write some docs
- [X] Add missing resources
- [X] Hook the new diff function
- [X] Do a no-op if xslt is empty

### DIFF
--- a/examples/xslt/main.tf
+++ b/examples/xslt/main.tf
@@ -1,0 +1,16 @@
+provider "libvirt" {
+  uri = "qemu:///system"
+}
+
+resource "libvirt_domain" "xslt-demo-domain" {
+  name = "xslt-demo-domain"
+  memory = "512"
+
+  network_interface {
+    network_name = "default"
+  }
+
+  xml {
+    xslt = "${file("nicmodel.xsl")}"
+  }
+}
--- a/examples/xslt/nicmodel.xsl
+++ b/examples/xslt/nicmodel.xsl
@@ -1,0 +1,17 @@
+<?xml version="1.0" ?>
+<xsl:stylesheet version="1.0"
+                xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+  <xsl:output omit-xml-declaration="yes" indent="yes"/>
+  <xsl:template match="node()|@*">
+     <xsl:copy>
+       <xsl:apply-templates select="node()|@*"/>
+     </xsl:copy>
+  </xsl:template>
+
+  <xsl:template match="/domain/devices/interface[@type='network']/model/@type">
+    <xsl:attribute name="type">
+      <xsl:value-of select="'e1000'"/>
+    </xsl:attribute>
+  </xsl:template>
+
+</xsl:stylesheet>
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -348,6 +348,20 @@ func resourceLibvirtDomain() *schema.Resource {
 				Default:  false,
 				ForceNew: false,
 			},
+			"xml": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"xslt": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ForceNew: true,
+						},
+					},
+				},
+			},
 		},
 	}
 }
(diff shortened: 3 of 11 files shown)


MalloZup:
i will have a look more in details on this during the week.

Generally i think is a good idea to have this kind of `custom` parser for modify the resource.
This will give User the possibility to implement their own Custom Resource if not supported, and also can give us the possibility to improve and focus on others `core zone` of the provider, instead of corner cases. :+1: :heart: :cupid: 

A part of this, i think as reminder we should keep in mind take care also in the mechanism following:

* `updating` resource. 
In the cloudinit (in the past) and also currently in networking resource, the update of resource is not rockstable.

In this case we should keep in mind that an user can modify the created initial xlst with other parameters in the file and this imho should force the recreation of the `domain` or other `resource`.

Also as corner case , we should not have a Diff if the xlst has spaces/tab etc, so we don't force recreation of resource for nothing ( see past cloudinit bug).

I will have look more and think about it more with calm :smiley:

Bischoff:
That is a great idea, and I think I might need that feature very soon! :heart: :heart: :heart:

Bischoff:
@dmacvicar What about skipping the transformation if `xslt.(string)` is empty (for example)?

Rationale: make transformation optional based on some other boolean. For example, in my use case, I need to leave the XML unchanged if there is no need to change the declared "hardware type".

dmacvicar:
We can make it a no-op if _xslt.(string)_ is empty. But then you would end with:

```hcl
xslt = "${var.hardware_type ? file(var.hardware_type) : ""}" 
```

Is that ok?

Bischoff:
@dmacvicar yes, that was exactly the idea.

That would spare writing the `noop.xsl`boilerplate file.

Bischoff:
@MalloZup I am using this feature for the test suite, so I would need it packaged.

For me it is good enough in this current state, and it just works smoothly.

dmacvicar:
I still need to do a last pass before merging. Hook the diff function and add the transformation to the other resources.

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_cloud_init_test.go#L72
 You can test in testacc that we do not have diffs see the example above

MalloZup:
i will revert the PR today

dmacvicar:
I think I have addressed all points. If you are fine with the nested `xml.xslt` attribute, this should be ready for review or merge.

This is a very advanced feature that should not be for day to day use, but to address shortcomings of the schema and corner cases, like our teams running testsuites with it.

//cc @Bischoff

Bischoff:
Thanks @dmacvicar !!!

PR #433, opened 2018-10-02
WIP: Add networking autoinst example

MalloZup:
this is just a missing example for networking autoinstall


MalloZup:
closing this becaue we have an infinite loop at moment so we don\t need it atm

PR #440, opened 2018-10-11
DON't merge POC: Libvirt events

MalloZup:
@dmacvicar commented: 
>I don't want to rush with this PR, as I was not a fan of the previous timeout loop either.
>
>I see that using `SRC_AGENT` without the agent connected fails immediately if no timeout is set, so >that is good. It won't get stuck.
>
>```console
>virsh # domifaddr 1
 >Name       MAC address          Protocol     Address
>-------------------------------------------------------------------------------
 >vnet0      7e:ba:be:47:7d:d8    ipv4         192.168.122.187/24
>
>virsh # domifaddr --source agent 1
>error: Failed to query for interfaces addresses
>error: Guest agent is not responding: QEMU guest agent is not connected
>```
>As using the agent at all is not a property of the domain, may be we could do the following:
>
>* If using the agent, at creation register `virConnectDomainEventRegisterAny` with >`VIR_DOMAIN_EVENT_ID_AGENT_LIFECYCLE` and create a simpl for >`VIR_CONNECT_DOMAIN_EVENT_AGENT_LIFECYCLE_STATE_CONNECTED` event, before issuing the command
>
>Testing this with _virsh_ produces the event as I expect:
>
>```console
>virsh # event --domain 1 --all
>```
>
>(starts _qemu-guest-agent_ on the machine)... and I get:
>
>```console
>event 'agent-lifecycle' for domain doman_ubuntu_resized: state: 'connected' reason: 'channel event'
>events received: 1
>```
>
>Would this be a better approach than timeouts?

___ 
USEFUL LINKS:

* https://github.com/libvirt/libvirt-go/blob/master/domain_events.go#L1321
* https://github.com/gofort/libvirt-cli/blob/6f7d3b9dc1f0ea021a86be85d046bd17dd4cc8ec/vendor/src/github.com/rgbkrk/libvirt-go/events_test.go#L9
* https://gist.github.com/toabctl/53f26989ad7634a3168b/
* https://github.com/rgbkrk/libvirt-go/blob/57a0dabced1462b94bafffa08b5a56c0a03b9d05/domain_1.2.14_test.go#L105
* https://www.redhat.com/archives/libvirt-users/2017-May/msg00035.html


MalloZup:
# Results: 

@dmacvicar so i played a bit with libvirt events ( was funny) :smile: . I have added some usefull examples in my last comment (https://github.com/dmacvicar/terraform-provider-libvirt/pull/432#issuecomment-427175738).

Here my facit:

i think for our current usecase libvirt event is overkill and will add more complexity. ( event handling etc).

Furthermore and i think the most important thing is : 
`VIR_CONNECT_DOMAIN_EVENT_AGENT_LIFECYCLE_STATE_CONNECTED` this event is sent** only 1 time**( if the domain has started). But what if the domain already exist?
In our use cases in terraform, we need to consider the situation where a domain **already exists** and we do for example `apply` or `refresh` or `plan`, so updates operation. In this case we endup in weird situation because the event was already sent and we cannot really count on it. 

Imho for solving this problem, the best and simple solution is to have the approach i did in https://github.com/dmacvicar/terraform-provider-libvirt/pull/432/commits/ec86c07288a78f2d4fc0a97f3e6f2d6ad6b74813
but without a refresh function from terraform. Additionaly to that patch, we don't check the QEMU_AGENT erorrs. ( otherwise we fail), and we use the loop timeout already present from `wait_for_lease` option.
( in this way we avoid to have 2 timeouts). In this way, we return empty interfaces in errors, or we use already the loop `wait_for_lease =true`. ( this mean that also for qemu-agent we need this var).

I think in that case is not worth to check error handling, we just need to return empty interfaces.

Anyway with my current poc i tested already the limits of this events.
```
018-10-06T16:13:46.769+0200 [DEBUG] plugin.terraform-provider-libvirt: 2018/10/06 16:13:46 TeneroTango: INSIDE EVENT!
2018-10-06T16:13:46.795+0200 [DEBUG] plugin.terraform-provider-libvirt: 2018/10/06 16:13:46 TeneroTango: INSIDE EVENT!
2018-10-06T16:13:51.769+0200 [DEBUG] plugin.terraform-provider-libvirt: 2018/10/06 16:13:51 ERROR: unable to get ifaces via qemu-agent virError(Code=86, Domain=10, Message='Guest agent is not responding: Guest agent not available for now')
2018-10-06T16:13:51.795+0200 [DEBUG] plugin.terraform-provider-libvirt: 2018/10/06 16:13:51 ERROR: unable to get ifaces via qemu-agent virError(Code=86, Domain=10, Message='Guest agent is not responding: Guest agent not available for now')
```

I got the error with this code
```golang
	gaCallback := func(c *libvirt.Connect, d *libvirt.Domain, eva *libvirt.DomainEventAgentLifecycle) {
		log.Printf("TeneroTango: INSIDE EVENT!")
		allInterfaces, err = domain.ListAllInterfaceAddresses(libvirt.DOMAIN_INTERFACE_ADDRESSES_SRC_AGENT)
		if err != nil {
			log.Printf("ERROR: unable to get ifaces via qemu-agent %s", err)
			...
	}

	gaCallbackID, err := virConn.DomainEventAgentLifecycleRegister(&domain, gaCallback)
```

imho we should use events for the autoyast feature but not for this one, because we don't gain much.

THe main problem are the hidden one, like when we shutdown/reboot or change the status of domain this will trigger events, and this we have complexity +  we need to consider the cases of broken connection/etc.

MalloZup:
@dmacvicar honestly i think libvirt-event are overkill for this feature and will have problems.

You can play with the POC here: https://github.com/dmacvicar/terraform-provider-libvirt/pull/440

Basically i think that we will have mre problems  then benifits,  i will continue the old PR **without** event and i will use not a new loop but use the old one, you will see on that other PR

MalloZup:
for listening in the loop: https://github.com/dmacvicar/terraform-provider-libvirt/pull/440/files#diff-f6f8db75203b0db83110f024f21cc4ebR226

dmacvicar:
I still think if it adds unnecessary complexity we can keep the method of looping and use the qemu agent connect function instead of the callbacks. 

I think if the user selected to use the agent, we should loop until we can finally connect to it, and rely on terraform general creation timeout instead.

MalloZup:
@dmacvicar  i agree with you:

this PR contain all the minimal changes for what you describe:
https://github.com/dmacvicar/terraform-provider-libvirt/pull/432

( this pr i want to keep it for implementing the autoyast)

Issue #218, opened 2017-11-08
Implementing AutoYaST support


dmacvicar:
# Motivation

We are using the provider internally heavily to test non-released products, and this means sometimes installing the machine from scratch and no images are available.

https://hackweek.suse.com/16/projects/implement-kernel-cmdline-and-slash-or-autoyast-slash-kickstart-support-in-terraform-provider-libvirt

# Approach

* While passing the profile via `fw_config` would be awesome, it is not supported. Parallel efforts will be done to move into that direction.
* The approach will likely be a `ISO` image containing `/autoinst.xml` using `autoyast=device://`
* Some code may be shared with `cloud-init` support code
* ~~I believe (without much evidence) the Ignition code can be simplified, and there is no need to upload the profile to a volume. In that case, a fw_config based version could be generic for both AutoYaST *_and_* Ignition.~~ volume is needed to have fw_config file= local to the libvirt daemon vs terraform.

# Task List

- [ ] Refactor cloud-init code to make it generic (ISO with X content + kernel option)
- [ ] Implement AutoYaST
- [ ] Implement unit tests
- [ ] Look if integration tests make sense

MalloZup:
@dmacvicar i am not sure that this feature could make sense in the direction we are going.

For me an user should be able to build the qcow image before using it.

For example, an user need to use kiwi for building their qcow ( this is equivalent of using autoyast on the highlevel)

Mixing and adding autoyast can also add more errors and maintainance on provider.

I would vote to close it because from the user perspective is not needed ( and also the codebase would be incosistent)

dmacvicar:
I think this would be a great addition for SUSE internally, as we deal a lot with testing installations. However, I understand nobody is wants to work on it now.

MalloZup:
I reopen this one because after thinking and your last comment is valid. So we might add it.

MalloZup:
@dmacvicar  i would be happy to continue this feature and finish it. ( after some discussion i think this still relevant for SUSE/openSUSE). ( we might also add open doors for Kickstart etc, but let see)

Would be helpful if you can share the status of what is missing to make it working, and if the actual status is reflecting the card. ( missing code, features etc).

We can sync on irc or github which medium you wish

dmacvicar:
You should read the link https://hackweek.suse.com/16/projects/implement-kernel-cmdline-and-slash-or-autoyast-slash-kickstart-support-in-terraform-provider-libvirt , as it has most of the information.

[...]

PR #456, opened 2018-10-19
[on-hold] Enable network autoinstallation

MalloZup:
## what does the feature:
#218 
This feature will add a new a bool variable `network_autoinstall`.

This variable will wait and block ( inside  a timeout can be also possible) until the reboot event from domain is caught.

When we catch the reboot_event of the domain, we will shutdown the domain and remove the `kernel` and `initrd`  components. ( this was why we did the infinite loop).

Once removing the kernel and initrd from domain, we can restart the domain and finish to create the resource.

The `network_autoinstall` variable is binded with `kernel` and `initrd`.

As main advantages with this implementation is that we can autoinstall also other distros in this way.

I attached an example

## Examples:
the example.tf : https://github.com/dmacvicar/terraform-provider-libvirt/pull/456/files#diff-104b8f76e49dd7878b6d3f38e41dc89a

## Details:

For details i did a first POC with comments for the next functions.

https://github.com/dmacvicar/terraform-provider-libvirt/pull/456/files#diff-2f03be069459ebc91fcf0607d4d686d8R537

@dmacvicar CC

I think this way is the simplest one, at moment i don't see problems with the implementations but i might oversee something :smile:


MalloZup:
@dmacvicar what is from your pov the best method for removing the `kernel` and `initrd` from the existing domain?

Sofar i am thinking that the current solution is to dump the current XML description of the domain,
remove the initrd, kernel, create a new-copy domain , and remove the old one.
I cannot find another way of updating it.
What do you think?

MalloZup:
@dmacvicar  feel free to have a look

I could today having a real network-installation VM with this patch :rocket: :rocket: 
( i used an internal sles example autoyast, because i had some issues with leap15 autoyast profiles, we can fix it later outside this PR)

At moment we might thinks about:

- [ ]  error handling inside the `callaback`: we are using `panic`, instead of return. Is still ok for you?
       if not i would need to find a way to catch the errors inside the callback and return later if error found.
      ( i need to check this out)

libvirt/resource_libvirt_domain.go
```diff
@@ -528,6 +535,76 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	// network-installation used in combination with kernel and initrd parameters.
+	NetworkAutoinstallEnabled := d.Get("network_autoinstall").(bool)
+	if NetworkAutoinstallEnabled {
```
dmacvicar: I think it would be better to have all the autoinstall related code in a separate file?

libvirt/resource_libvirt_domain.go
```diff
@@ -528,6 +535,76 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
...
+		stop := make(chan int)
+		rebootCallBack := func(c *libvirt.Connect, d *libvirt.Domain) {
+			log.Printf("[DEBUG:] Libvirt-events: Caught reboot event!")
+			err := destroyDomain(domain)
+			if err != nil {
+				panic(err)
+			}
+			domainDef, _ := getXMLDomainDefFromLibvirt(domain)
+			domainDef.OS.Kernel = ""
```
dmacvicar: This completely changes the semantics of _initrd_ and _kernel_, as they will be gone after reboot, and then terraform would attempt to re-add them.

dmacvicar:
I am still not convinced about this.

* It adds lot of complexity to the domain file (it should be separated)
* It changes the semantics of _initrd_ and _kernel_ and I am not sure if after applying again terraform would try to fix that again
* It has a confusing name for what it does

* So, if we want to name it _network_autoinstall_ I would expect much more, like detecting the proper way of detecting how to boot from the url and proper re-setting of the original _initrd_ and _kernel_ values.

In its current state, where everything is manual, and the only thing it does it waiting for reboot, I'd then name it for what it does:

* It could be named "wait_for_reboot" where one could specify the number of reboots to wait for. Then we document how this can be useful for an autoinstall when specifying initrd and kernel.
* It could be named "wait_for_agent" where we don't accept the domain as created until the agent connects for first time. Then we document how this is useful for autoinstall if you specify the agent in the profile.

libvirt/resource_libvirt_domain.go
MalloZup: Yes, but there is no workaround for it. We need to remove them

MalloZup: I think the user should be aware that he need.to change the tf file

libvirt/resource_libvirt_domain.go
MalloZup: If you have name suggestion feel free :)

MalloZup:
Thx for review. I agree that the name is not sexy, but i choosed that name with the future in mind. If we choose "wait for reboot" etc, this will be deprecated, once we got to dinamically sets of initrd and kernel. Meaning we have 2 possible ways which one is deprecated.
Having that var now will avoid api breaking changes
I can have a look now for adding support autodetection to opensuse/ubuntu, it is maybe a low hanging fruit  let see. (Wrote from.my phone)

libvirt/resource_libvirt_domain.go
dmacvicar: I would choose the name depending how we end merging this.

So, either _domain_reboot.go_ or _domain_autoinstall.go_....

MalloZup:
@dmacvicar  i adreessed all the comments.

I had also a first look howto implement the automatically `kernel` `initrd` autodetection.

For the auto-detection we will need to implement a `parser` which auto-detect from the OS present from the URL string , then append to the baseString the initrd and Kernel path.

I had a look also on how `ubuntu` does, and they have  a `initrd.gz` zipped... so this will make more complex the logic for the future. :smile:  Meaning some distros have zipped initrd in official mirrors etc.

To me honestly, the `auto-detection` is something fragile because 1) parser (parser are fragile by nature :smile: )  2) we will rely on OS-specification that might change.

Thinking retrospectively the choice of having `initrd` and `kernel` resource is a really stable one without any maintainance cost. ( we delegate all the logic/download to users).

Anyway, i'm open if we want to rename the variable `network_autoinstall` to one name of your suggestion, but i think that name is not 100 perfect but we keep the door open if in the future someone would like to write/and maintain (:smile: ) that parser

MalloZup:
@dmacvicar once you have time feel free to share your take on this, and if we can get it inside :pray: :smile:

MalloZup:
Agreed and share your pov there, i will rename it accordingly thx

MalloZup:
For the strategy yop once you have time is ok for me

MalloZup:
After some thinking,  I believe this feature does not belong here, in term of design, vision. We solve the problem of deploy, here. If we mix installation and deploy, we deviate from Unix Philosophy. Indeed we could help some users with this, but it will only bring the illusion to solve all problems with one centralized tool. in Linux we have the heritage of solving with small, separate tools big problems, and I will continue with it. Kiwi or others devops tools can solve the problem of installation. Anyway this will open the port of complexity bugs unknwon at moment. I spotted this already (e. G we break feature like wait for lease, it will ackward to wait for an ip address of a domain, which is installing in codebase etc).  It was a funny prototype to learn some internals and also libvirt events, so I don't consider a failure closing this, either the other way. Thank you @dmacvicar for being accurate, open and showing constructive criticism, this is why I love to contribute here :) closing

PR #472, opened 2018-11-07
Support GET-only HTTP servers for image sources

moio:
Some HTTP services, notably GitHub Releases, do not support HEAD requests to their endpoints.

`terraform-provider-libvirt` will thus fail on something along the lines of:

```hcl
resource "libvirt_volume" "centos7_volume" {
  name = "centos7"
  source = "https://github.com/moio/sumaform-images/releases/download/4.0.0/centos7.qcow2"
}
```

With:

```
libvirt_volume.centos7_volume: Error accessing remote resource: https://github.com/moio/sumaform-images/releases/download/4.0.0/centos7.qcow2 - 403 Forbidden
```

I observed this phenomenon in [sumaform](https://github.com/moio/sumaform) when trying to use images built with Packer in CircleCI ([sumaform-images](https://github.com/moio/sumaform-images) project).

This PR removes the assumption a HEAD request will succeed. If it doesn't, it will defer getting the `Content-Length` we need for the libvirt volume definition at GET time.

Test code has been adapted too, namely, the internal Web Server also sets `Content-Length` now.

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -143,11 +143,12 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 
 		// update the image in the description, even if the file has not changed
 		size, err := img.Size()
-		if err == nil {
-			log.Printf("Image %s image is: %d bytes", img, size)
-			volumeDef.Capacity.Unit = "B"
-			volumeDef.Capacity.Value = size
+		if err != nil {
+			return err
 		}
+		log.Printf("Image %s image is: %d bytes", img, size)
+		volumeDef.Capacity.Unit = "B"
+		volumeDef.Capacity.Value = size
 	} else {
 
 		// the volume does not have a source image to upload, first handle
@@ -234,7 +235,7 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 
 	// upload source if present
 	if _, ok := d.GetOk("source"); ok {
-		err = img.Import(newCopier(client.libvirt, volume), volumeDef)
+		err = img.Import(newCopier(client.libvirt, volume, volumeDef.Capacity.Value), volumeDef)
 		if err != nil {
 			return fmt.Errorf("Error while uploading source %s: %s", img.String(), err)
 		}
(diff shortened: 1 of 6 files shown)


libvirt/resource_libvirt_volume.go
```diff
@@ -143,12 +143,11 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 
 		// update the image in the description, even if the file has not changed
 		size, err := img.Size()
-		if err != nil {
-			return err
+		if err == nil {
```
MalloZup: if we do so, we are not checking and returning the error of the above function 
or i am missing something here?  :thinking:

moio: At the moment there's not many other failure scenarios but I see the general point. If this is the only thing holding back merging just +1 this and I will fix it of course.

MalloZup: @moio thanks, yop otherwise looks clean to me :guitar:

wking: I like how you've pushed the size information down into the copier (most of a2e28a3).  But I'd rather leave this particular hunk as it stands.  To have `httpImage.Size()` support endpoints that 403 `HEAD`:

```console
$ curl -sLI https://github.com/moio/sumaform-images/releases/download/4.0.0/centos7.qcow2 | grep 'HTTP\|Location'
HTTP/1.1 302 Found
Location: https://github-production-release-asset-2e65be.s3.amazonaws.com/...
HTTP/1.1 403 Forbidden
```

how about a "you can't make me read the body" `GET` ;) :

```go
func (i *httpImage) Size() (uint64, error) {
  response, err := http.Head(i.url.String())
  if err != nil {
    return 0, err
  }
  if response.StatusCode == 403 {
    response, err = http.Get(i.url.String())
    if err != nil {
      return 0, err
    }
    response.Body.Close()
  }
  if response.StatusCode != 200 {
    ...
  }
  ...
}
```

moio:
Oooh how much more I like my patch now!

Thank you very much @wking!

MalloZup:
For curiosity how @wking helped here? :) i am walking at moment i will merge all pr after including the wking one. Thx for prs

moio:
@MalloZup https://github.com/dmacvicar/terraform-provider-libvirt/pull/472#discussion_r232153699

libvirt/resource_libvirt_volume.go
MalloZup: nice thx @wking  and @moio for this :japan:

Issue #473, opened 2018-11-07
Cannot use cloud-init on RancherOS


tommyknows:
#  Version Reports:

### Distro  version of host:

``` Ubuntu 18.04```

### Terraform Version Report

```sh
terraform -v
Terraform v0.11.10
+ provider.libvirt (unversioned)
+ provider.template v1.0.0
```
(libvirt built from source a few days ago)

### Provider and libvirt versions

```sh
terraform-provider-libvirt -version
This binary is a plugin. These are not meant to be executed directly.
Please execute the program that consumes these plugins, which will
load any plugins automatically
```
(? 😄)

___
# Description of Issue/Question
I cannot use cloud-init for configuring a RancherOS Instance.
According to [this](https://github.com/rancher/os/issues/1501), the drive should be called "config-2", although I'm neither an expert on cloud-init nor RancherOS. 
The instance boots, but does not apply the cloud-init provided through the terraform module. From the RancherOS Host's cloud-init log:
```
time="2018-11-07T15:11:41Z" level=info msg="datasources that will be consided: []string{"configdrive:/media/config-2"}"
time="2018-11-07T15:11:41Z" level=info msg="cloud-init: Checking availability of "cloud-drive""
time="2018-11-07T15:11:41Z" level=error msg="no such file or directory"
time="2018-11-07T15:11:41Z" level=info msg="cloud-init: Datasource unavailable, skipping: cloud-drive: /media/config-2 (lastError: no such file or directory)"
time="2018-11-07T15:11:41Z" level=info msg="Cloud-init datasource that was used: %!s(<nil>)"
```

It should be the same as with CoreOS, so I guess cloud-init wouldn't work with CoreOS too. However, with CoreOS, you'd use ignition, so maybe no one noticed.

### Setup
```
provider "libvirt" {
    uri = "qemu:///system"
}

data "template_file" "user_data" {
  template = "${file("cloud_init.cfg")}"
}

resource "libvirt_cloudinit_disk" "commoninit" {
  name           = "commoninit.iso"
  user_data      = "${data.template_file.user_data.rendered}"
}

resource "libvirt_domain" "kubernetes" {
  name   = "kubernetes-terraform"
  memory = "2048"
  vcpu   = 2

  cloudinit = "${libvirt_cloudinit_disk.commoninit.id}"

  network_interface {
    network_name = "default"
  }

  boot_device {
    dev = [ "hd", "network"]
  }

  console {
    type        = "pty"
[...]

PR #476, opened 2018-11-11
Implement DataSourceTypes to allow for different CloudInit Implementations

tommyknows:
This PR addresses #473 .
I have investigated why RancherOS (and CoreOS afaik, but haven't tested that) does not read the CloudInit File. 
There were two "issues" with the current implementation:

- The drive should have a volume label called `config-2`
- RancherOS expects the config drive to have an "openstack/latest" folder, in which the CloudConfig Files are located. I have called this "dataSourceType", which in this case would be `openstack`.

This is why I have implemented a field in the cloudInit resource called `data_source_type`. (if you have any naming suggestions, I am open for it).

The data source type influences the creation (and reading) of the cloudconfig drive and solves the above issues.

I have also added two test resources to go with the code, however, i do not know if they were really necessary tbh.

### DIFF
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -280,7 +280,7 @@ func (ci *defCloudInit) setCloudInitDataFromExistingCloudInitDisk(virConn *libvi
 		if err == io.EOF {
 			break
 		}
-		log.Printf("[DEBUG]: Processing file %+v from cloudInit disk")
+		log.Printf("[DEBUG]: Processing file %+v from cloudInit disk", file.Name())
 
 		if err != nil {
 			return err


MalloZup:
@tommyknows  hi ! thank for this pr.

For coreos we support here alredy the `ignition` https://coreos.com/ignition/docs/latest/what-is-ignition.html 

https://github.com/dmacvicar/terraform-provider-libvirt/blob/b11e5b79ff40dc8f03d762d92226fabe315955dc/website/docs/r/coreos_ignition.html.markdown

I think you pr looks nice and correct, thank you for the effort and time invested.

However to me this fix seems less generic then `upstream`.
E.g could you argument more on the 2 specific points you mentioned like:


1)  The drive should have a volume label called config-2  (why and do we have link info?) 
2)   RancherOS expects the config drive to have an "openstack/latest" folder, in which the CloudConfig Files are located. I have called this "dataSourceType", which in this case would be openstack.
 ( why and do we have more info/links?)


Thank you for your time and efforts. 
 i am trying to make the provider with an opensource and  long-term vision as possible, so this mean we accept every OS but we need also to be generic ( in term of the architecture design we want to take also )
TIA

@dmacvicar jfyi

tommyknows:
Thanks for the comments.

Yes, basically it fixes something that RancherOS is missing - the `NoCloud` data source.
The DataSources of RancherOS are documented [here](https://rancher.com/docs/os/v1.2/en/running-rancheros/server/pxe/). 
They have [pre-baked images](https://github.com/rancher/os/releases/tag/v1.4.2) that watch for a specific CloudInit DataSource, but the only qcow2-image is the openstack build - thus I had implemented this datasource.

Also, I am unsure why it needs the `config-2` label, but I had acted upon this issue: https://github.com/rancher/os/issues/2544 

I agree with you that we should not need to implement different DataSources than `NoCloud`, as KVM is not a cloud :-)

I guess opening an issue at the RancherOS Repo would be more appropriate, but this fix was easier for me to do & fixes issues with other OSes (CoreOS as another example, although I know that you support Ignition) too.

So it's upon you to decide wether you want to merge the code or if it's not generic enough :-)

nschlewitz:
Hello,

I know this issue is closed but basically, this is not only a rancherOS problem. We are trying to use this terraform libvirt plugging on a Redhat 7.5 host which is working with qemu-kvm 1.5.3. Unfortunately, this package doesn't support the -fw_cfg parameter so this makes impossible a CoreOs VM creation using an ignition file. @tommyknows development seems to solve this issue as CoreOs doesn't support a NoCloud cloudinit file but rather an openstack/latest config-2 iso. With this new argument, is it possible to re-open this issue and reconsider to merge @tommyknows code and allow the support of Cloudinit for CoreOs.

johnfedoruk:
+1

Issue #389, opened 2018-08-31
updating existing networks 


MalloZup:
currently all network that we create, we cannot modify them  via a terraform Resource update api call.

This is due because we don't implement Network.update call from libvirt ( i try it out, it is not really straightforward because we need to consider various flags plus pass the xml part of updating network)

https://libvirt.org/html/libvirt-libvirt-network.html#virNetworkUpdate

To me the easiest solution would be to destroy the network and recreate it  but this has implication that the domains need to be restarted. :rofl: :robot: 

I think we should investiage a little on the libvirt.update or if we don|t implement the functionality we should state this in some documentation.
@inercia @dmacvicar jfyi

MalloZup:
clsoing this since we merged several prs for improving. let's wait if user add more specific issues

PR #477, opened 2018-11-12
Allow DHCP configuration of static ip addresses for network resource

aaannz:
Libvirt DHCP can configure static ip address assignements. This
commit adds support to read resource config with extended dhcp
settings with dhcp host-ip-mac assignements.

Example:

    ```
    resource "libvirt_network" "my_network" {
      name = "my_network"
      mode = "nat"
      domain = "example.org"
      addresses = [ "192.168.1.0/24" ]
      dhcp = {
          enabled = true
          hosts = [
            {
              host = "host1"
              ip = "192.168.1.2"
              mac = "00:11:22:33:44:55"
            },
            {
              host = "host2"
              ip = "192.168.1.3"
              mac = "00:11:22:33:44:56"
            },
            {
              host = "host3"
              ip = "192.168.1.4"
              mac = "00:11:22:33:44:57"
            }
          ]
      }
    }
    ```

### DIFF
--- a/libvirt/data_source_libvirt_network.go
+++ b/libvirt/data_source_libvirt_network.go
@@ -156,3 +156,76 @@ func resourceLibvirtNetworkDNSSRVRead(d *schema.ResourceData, meta interface{}) 
 
 	return nil
 }
+
+//// DHCP
+
+// a libvirt network DHCP host template datasource
+//
+// Datasource example:
+//
+// data "libvirt_network_dhcp_host_template" "k8smasters" {
+//   count = "${var.master_count}"
+//   ip = "${var.master_ips[count.index]}"
+//   mac = "${var.master_mac[count.index]}"
+//   name = "master-${count.index}"
+// }
+//
+// resource "libvirt_network" "k8snet" {
+//   ...
+//   dhcp = [{
+//     hosts = [ "${flatten(data.libvirt_network_dhcp_host_template.k8smasters.*.rendered)}" ]
+//   }]
+//   ...
+// }
+//
+func datasourceLibvirtNetworkDHCPHostTemplate() *schema.Resource {
+	return &schema.Resource{
+		Read: resourceLibvirtNetworkDHCPHostRead,
+		Schema: map[string]*schema.Schema{
+			"ip": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"mac": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"rendered": {
+				Type: schema.TypeMap,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceLibvirtNetworkDHCPHostRead(d *schema.ResourceData, meta interface{}) error {
+	dhcpHost := map[string]interface{}{}
+	if address, ok := d.GetOk("ip"); ok {
+		ip := net.ParseIP(address.(string))
+		if ip == nil {
+			return fmt.Errorf("Could not parse address '%s'", address)
+		}
+		dhcpHost["ip"] = ip.String()
+	}
+	if address, ok := d.GetOk("mac"); ok {
+		mac, _ := net.ParseMAC(address.(string))
+		if mac == nil {
+			return fmt.Errorf("Could not parse MAC address '%s'", address)
+		}
+		dhcpHost["mac"] = mac.String()
+	}
+	if name, ok := d.GetOk("name"); ok {
+		dhcpHost["name"] = name.(string)
+	}
+	d.Set("rendered", dhcpHost)
+	d.SetId(strconv.Itoa(hashcode.String(fmt.Sprintf("%v", dhcpHost))))
+
+	return nil
+}
--- a/libvirt/provider.go
+++ b/libvirt/provider.go
@@ -28,8 +28,9 @@ func Provider() terraform.ResourceProvider {
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
-			"libvirt_network_dns_host_template": datasourceLibvirtNetworkDNSHostTemplate(),
-			"libvirt_network_dns_srv_template":  datasourceLibvirtNetworkDNSSRVTemplate(),
+			"libvirt_network_dns_host_template":  datasourceLibvirtNetworkDNSHostTemplate(),
+			"libvirt_network_dns_srv_template":   datasourceLibvirtNetworkDNSSRVTemplate(),
+			"libvirt_network_dhcp_host_template": datasourceLibvirtNetworkDHCPHostTemplate(),
 		},
 
 		ConfigureFunc: providerConfigure,
--- a/libvirt/network.go
+++ b/libvirt/network.go
@@ -176,6 +176,24 @@ func setDhcpByCIDRAdressesSubnets(d *schema.ResourceData, networkDef *libvirtxml
 				return err
 			}
 			if d.Get("dhcp.0.enabled").(bool) {
+				// prepare host static ip assignement if present
+				var staticHosts []libvirtxml.NetworkDHCPHost
+				if hostsCount, ok := d.GetOk("dhcp.0.hosts.#"); ok {
+					for i := 0; i < hostsCount.(int); i++ {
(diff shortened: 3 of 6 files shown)


MalloZup:
@dmacvicar good catch. Yop we implemented the `datasource` here.

https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/data_source_libvirt_network.go

I think the `data_source` here make more sense.

aaannz:
Sure, I will redo it as a datasource then. Thanks for feedback.

Just to be clear, both datasource and network schema approach is desirable, right? As I see DNSSRV parsing in resource network schema as well.

aaannz:
Added documentation and possibility to use datasource (I hope I understood the concept correctly). Also fixed the testsuite since apparently `11:22:33:44:55:66` MAC address is reserved multicast MAC.

@MalloZup `make testacc` passes my tests, though I have failure for `TestAccLibvirtDomain_NetworkInterface` and sometimes `TestAccLibvirtDomain_Consoles`. But I have these failures even without my changes. Can be due to a test server being SLE12SP3 with older libvirt?
About the update. This version forces resource recreation, I did not yet implement update mechanism. So modification of static entries and subsequent `terraform apply` destroys the network and recreates it.

MalloZup:
@aaannz thank much for this PR! :guitar: :guitar:  i don't have time today for having deep look, i will do it during the week.

For datasource, take inspiration from this commit: https://github.com/dmacvicar/terraform-provider-libvirt/commit/a4d0ba6a319d8728cb5d6c10aae593bdd27da516

I think you are missing one step on `provider.go`.

For `TestAccLibvirtDomain_Consoles.` failure this is kind a issues we have from time to time, depending on which `pts` we are running the test.

for `TestAccLibvirtDomain_NetworkInterface` if don't remember to have seen it recently,  feel free to add the failure here. :sunflower: cu

aaannz:
> I think you are missing one step on provider.go.

I am not sure what I missed. I already added datasource parsing there and for dhcp I have only one datasource (instead of two options for DNS)
> For the `update` mechanism i know for sure we have issues on the whole networking part, see issue #389.
>
> We could create a follow-up issue on this, for fixing it separately, since i think it might need some works.
>
> Of course if you implement the update mechanism would be extra +1

I prefer to tackle `update` mechanism in separate issue.

Is there anything else to do?

MalloZup:
The update mechanism need to be handled in this pr for not introducing bugs

libvirt/data_source_libvirt_network.go
```diff
@@ -156,3 +156,76 @@ func resourceLibvirtNetworkDNSSRVRead(d *schema.ResourceData, meta interface{})
...
+	return &schema.Resource{
+		Read: resourceLibvirtNetworkDHCPHostRead,
+		Schema: map[string]*schema.Schema{
+			"ip": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"mac": {
+				Type:     schema.TypeString,
+				Required: true,
```
inercia: Is this really `Required`? I guess users will use `ip` with `mac`/`name`/`mac`-and-`name`,  so I gues both `mac` and `name` should be `Optional: true`...

libvirt/provider.go
```diff
@@ -28,8 +28,9 @@ func Provider() terraform.ResourceProvider {
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
-			"libvirt_network_dns_host_template": datasourceLibvirtNetworkDNSHostTemplate(),
-			"libvirt_network_dns_srv_template":  datasourceLibvirtNetworkDNSSRVTemplate(),
+			"libvirt_network_dns_host_template":  datasourceLibvirtNetworkDNSHostTemplate(),
+			"libvirt_network_dns_srv_template":   datasourceLibvirtNetworkDNSSRVTemplate(),
+			"libvirt_network_dhcp_host_template": datasourceLibvirtNetworkDHCPHostTemplate(),
```
inercia: Don't you guys find all these templates a bit confusing? Shouldn't all be the same? Just a template where some IPs (with an optional MAC) should be mapped to some names?

libvirt/data_source_libvirt_network.go
```diff
@@ -156,3 +156,76 @@ func resourceLibvirtNetworkDNSSRVRead(d *schema.ResourceData, meta interface{})
...
+			},
+			"name": {
```
inercia: Is this a hostname? If that is the case, maybe we should use `hostname` (as we do in the `libvirt_network_dns_host_template`)

libvirt/data_source_libvirt_network.go
aaannz: In case of standalone dhcp datasource, mac is indeed a required one, as is the ip.

libvirt/data_source_libvirt_network.go
aaannz: Indeed a hostname, I used `name` as used in libvirt - https://github.com/libvirt/libvirt-go-xml/blob/master/network.go#L145

libvirt/provider.go
aaannz: True, I did not think about using just one template for both dns and dhcp. Once I'll have time to work on this again, I'll try it.

abbbi:
hi,

i have been playing around with your branch as i have the demand of fixed ip addresses and hostnames via DHCP and it works quite well.

There is however one thing im not quite sure if its a issue others will hit aswell: if one updates the network, all running clients will loose connectivity because the MAC address of the virtual bridge changes.

Lets assume you have a host list which creates a subset of hosts and dns entries for the network ressource, and you add one, your terraform plan ends up like this:

> -/+ libvirt_network.qa_network (new resource required)
>       id:                                 "652dfe06-79e3-4ad2-9862-341a581a6cb6" => <computed> (forces new resource)
>       addresses.#:                        "1" => "1"
>       addresses.0:                        "10.17.1.0/24" => "10.17.1.0/24"
>       bridge:                             "virbr3" => <computed>
>       dhcp.#:                             "1" => "1"
>       dhcp.0.enabled:                     "true" => "true"
>       dhcp.0.hosts.#:                     "3" => "4" (forces new resource)
>       dhcp.0.hosts.0.ip:                  "10.17.1.50" => "10.17.1.50"
>       dhcp.0.hosts.0.mac:                 "52:54:01:6c:3c:50" => "52:54:01:6c:3c:50"
>       dhcp.0.hosts.0.name:                "ma-setup-rds1" => "ma-setup-rds1"
>       dhcp.0.hosts.1.ip:                  "10.17.1.80" => "10.17.1.80"
>       dhcp.0.hosts.1.mac:                 "52:54:01:6c:3c:80" => "52:54:01:6c:3c:80"
>       dhcp.0.hosts.1.name:                "ma-setup-server1" => "ma-setup-server1"
>       dhcp.0.hosts.2.ip:                  "10.17.1.2" => "10.17.1.2"
>       dhcp.0.hosts.2.mac:                 "52:54:01:6c:3c:01" => "52:54:01:6c:3c:01"
>       dhcp.0.hosts.2.name:                "ma-setup-client1" => "ma-setup-client1"
>       dhcp.0.hosts.3.ip:                  "" => "10.17.1.3" (forces new resource)
>       dhcp.0.hosts.3.mac:                 "" => "52:54:01:6c:3c:02" (forces new resource)
>       dhcp.0.hosts.3.name:                "" => "ma-setup-client2" (forces new resource)
> 

the last hosts entry (dhcp.0.hosts.3) is the client i have added to the resource.
After appyling it, the underlying libvirt network changes its mac address, the bridge however stays the same:

before apply the libvirt network definition looks like:

>   <bridge name='virbr3' stp='on' delay='0'/>
>   <mac address='52:54:00:42:06:49'/>

after apply:

>   <bridge name='virbr3' stp='on' delay='0'/>
>   <mac address='52:54:00:69:12:78'/>

this means that all already existing clients will loose their network connectivity, only the host
which was added to the resource will be reachable.

aaannz:
@abbbi Yes, that is a known issue I have to fix. Currently any change forces resource recreation instead of update. I do not think I'll have time this week, but next one I should have a slot to look into it.

MalloZup:
Closing fee' free to reopen in case you want to prove

abbbi:
@aaannz any chance you are going to look at this again? it would be really helpful if the plugin would support this.

MalloZup:
@abbbi open an issue referring to this pr and we might finish this

MalloZup:
I m reopening this pr but I think we need to signale that somebldy need to take over on this pr

MalloZup:
i have opened an issue with a link to this pr. work can be continued, discussed ther with a new pr. closing as stale

PR #478, opened 2018-11-13
close stream, check size, and fix backingstore  and other minor size assignement

MalloZup:
this is a logical backport of what we are doing in `utils.volume.go` file already

The history behind this is a tentative to fix an issue we have currently when updating the `cloudinit.iso` and creating a vol on remote-kvm server.

### DIFF
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -341,7 +341,7 @@ func downloadISO(virConn *libvirt.Connect, volume libvirt.StorageVol) (*os.File,
 
 	if uint64(bytesCopied) != info.Capacity {
 		stream.Abort()
-		return tmpFile, fmt.Errorf("Error while copying remote volume to local disk, bytesCopied %d !=  %d volume.size:", bytesCopied, info.Capacity)
+		return tmpFile, fmt.Errorf("Error while copying remote volume to local disk, bytesCopied %d !=  %d volume.size", bytesCopied, info.Capacity)
 	}
 
 	err = stream.Finish()
--- a/libvirt/cloudinit_def.go
+++ b/libvirt/cloudinit_def.go
@@ -304,6 +304,8 @@ func readIso9660File(file os.FileInfo) ([]byte, error) {
 // pointer when you are done.
 func downloadISO(virConn *libvirt.Connect, volume libvirt.StorageVol) (*os.File, error) {
 	// get Volume info (required to get size later)
+	var bytesCopied int64
+
 	info, err := volume.GetInfo()
 	if err != nil {
 		return nil, fmt.Errorf("Error retrieving info for volume: %s", err)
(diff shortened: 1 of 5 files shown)


MalloZup:
@Bischoff  feel free to build from here

MalloZup:
@dmacvicar  this is a tentative fix for fixing some issue we have currently.

In deep we are uploading the cloudinit disk but this stuck forever when uploading it to a remote-kvm server.

This pr make per-se already sense, see https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/utils_volume.go#L220

MalloZup:
@juliogonzalez jfyi, i have seen 3 builds without any errors should be fine :pray:  , and make sense to me that we had that issue,  (since the stream was not free), let's see tmw

MalloZup:
@dmacvicar  this has not fixed the issue but is to me a good small nice to have. feel free to review once you have time

dmacvicar:
Please hold until I merge https://github.com/dmacvicar/terraform-provider-libvirt/pull/479 which is more important and touches also the same file. Then we rebase this one.

MalloZup:
ok no pb

Bischoff:
At least tests have shown this does not break anything. LGTM.

MalloZup:
@dmacvicar  i have added more things to this which i am currenlty seeing if they works as expected :dancing_men:

MalloZup:
@dmacvicar #479  merged, we can merge this. feel free to merge once you have time

dmacvicar:
I was reviewing the code again, and I am not sure this is the right thing (nor the current code).

https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/stream.go implements _Close_:

```golang
// Close closes the stream
func (sio *StreamIO) Close() error {
	return sio.Stream.Finish()
}
```

However, this is not part of _Reader_ nor _Writer_, and it is not used. We actually defer a _Finish()_ just after creating the stream and before wrapping it:

```golang
// download ISO file
stream, err := virConn.NewStream(0)
if err != nil {
	return tmpFile, err
}
defer stream.Finish()
```

If we look at the wrapper of stream in go:

```golang
// See also https://libvirt.org/html/libvirt-libvirt-stream.html#virStreamRecv
func (v *Stream) Recv(p []byte) (int, error) {
	var err C.virError
	n := C.virStreamRecvWrapper(v.ptr, (*C.char)(unsafe.Pointer(&p[0])), C.size_t(len(p)), &err)
	if n < 0 {
		return 0, makeError(&err)
	}
	if n == 0 {
		return 0, io.EOF
	}

	return int(n), nil
}
```

The C documentation for [virStreamFinish](https://libvirt.org/html/libvirt-libvirt-stream.html#virStreamFinish) states:

>Indicate that there is no further data to be transmitted on the stream. For output streams this should be called once all data has been written. For input streams this should be called once virStreamRecv returns end-of-file.
>
>This method is a synchronization point for all asynchronous errors, so if this returns a success code the application can be sure that all data has been successfully processed.

One can confirm the workflow in [virStreamRecv](https://libvirt.org/html/libvirt-libvirt-stream.html#virStreamRecv) docs example:

* Loop
  * Recv
  * if err is EOF, Finish stream, and break
  * If err is other either when reading or writing to our sink, Abort stream
* If the stream is not aborted, call Finish again and see if an error is reported (sync point for error)
* in all cases, Free the stream

The alternative code that we have for volumes is:

```golang
defer func() {
	if uint64(bytesCopied) != info.Capacity {
		stream.Abort()
	} else {
		stream.Finish()
	}
	stream.Free()
}()
```

I can't see the full logic in that defer. For example, the last _stream.Finish()_ may hide an error.

I suspect that code was modified to check for bytesCopied vs info.Capacity because the copy failed, but returned no error, because the libvirt docs were ignored and the finish to get the error was not called. The caller noticed in this failure the bytes copied were not full and therefore compared this.

If I had to copy the logic, and given that _io.Copy_ stops at the _EOF_, I think we should:

* wrap the stream
  * defer Free
* io.Copy
  * if error either in copy or elsewhere, Abort the stream
  * otherwise, Finish the stream
    * if finishing the stream returns error, report error (this error may not had shown in the io.Copy loop)

And probably I would, unless we find a way to encapsulate this logic in the stream wrapper itself, remove the _Close_ function.

MalloZup:
thx for report.I agree with you on that point, i have already expirimented it that the finish stream was erroring locally and we should call it outside defer. I might have deep lookk tomorrow on this. Thx for pointers is really interessanting :)

Bischoff:
Yes, thanks a lot. @MalloZup, do you think there a chance this is the root cause for our "timeout" problem?

MalloZup:
@Bischoff it can be but from your side you should upgrade libvirt this will not harm :)

MalloZup:
# tests coverage after running make testacc
copier

MalloZup:
@dmacvicar  @inercia i have repushed things according to @dmacvicar  comments.

(The close call was not needed.) i added the test coverage for showing that now the finish is called.

MalloZup:
# cloudinit stream
cloduinit

dmacvicar:
Did this help with the issues you were facing?

MalloZup:
It was not tested yet but i think is better to have it ( it is more clear)@moio @ebischoff we might give a try with this one, let.me know when we could see it

Bischoff:
I tested the initial version from Dario (not the latest version from here).

I don't really want to test the latest version as we gave up this cloudinit thing, and I don't think it would solve the cloudinit problem anyway.

But the code looks good to me and I'm still in favour of integrating it (once it's tested of course).

MalloZup:
to me we can merge this as standalone improvement for better handling error on stream and also the mechanism finish.stream is better then before. If the sumaform community has different Prios now, i can understand, however i would have enjoyed to see this applied because at least we could detect the root cause of the error or even fix it. I think with this we could print the `error` where before was not printed but yop, i can understand. On the kubic side we will continue to use  cloudinit and so far we had not issues with it. tia :sunflower:

Issue #455, opened 2018-10-19
Retry on failures downloading images


juliogonzalez:
This is a feature request.

Today we are having HTTP 503 problems for a few days while downloading images from download.opensuse.org (one of the mirros returns 503 from time to time, see https://progress.opensuse.org/issues/42725 for details).

For the time being I asked openSUSE Heroes to disable the affected mirror, but it would be good if 

```
func (i *httpImage) Import(copier func(io.Reader) error, vol libvirtxml.StorageVolume) error {
```

Could retry for a few times before giving up.

Not sure if it should retry on all errors, but at least maybe on 403, 503 and network errors, as that could help avoid problems when using mirroring such as download.opensuse.org, and would not have really impact if using URLs that are not mirrors (other than the delay for the retries).

dmacvicar:
I will give it a shot using something like https://github.com/Rican7/retry

MalloZup:
https://github.com/cenkalti/backoff is also nice

MalloZup:
@dmacvicar  one question, i was wondering why we introduce a new dep/library.

Is this not solvable via terraform timeouts/refresh functions?

Or you proposed that library because of the `backoff` functionality? tia

dmacvicar:
I haven't decided whether to use the library. Ma be just a for loop that retries for 5xx and fails for 4xx is enough.

PR #479, opened 2018-11-13
WIP: Retry HTTP downloads for 5xx errors

dmacvicar:
This PR implements what is suggested on https://github.com/dmacvicar/terraform-provider-libvirt/issues/455

Most of the work was to do the proper testcase.

I also discovered a bug: we where not erroring at all for non-200 codes. We ignored [this part of the documentation](https://golang.org/pkg/net/http/#Client.Do):

>An error is returned if caused by client policy (such as CheckRedirect), or failure to speak HTTP (such as a network connectivity problem). A non-2xx status code doesn't cause an error. 

So _Import_ did not error, and the error we spit to the user was probably some garbage when trying to read an empty response. This is now fixed in 5fa3124f87600e6b5a87b692660fbe03566bc23d

### TODO

- [ ] wait/backoff between retries

### DIFF
--- a/libvirt/utils_volume.go
+++ b/libvirt/utils_volume.go
@@ -186,10 +186,15 @@ func (i *httpImage) Import(copier func(io.Reader) error, vol libvirtxml.StorageV
 	}
 
 	defer response.Body.Close()
+	log.Printf("[DEBUG]: url resp status code %s\n", response.Status)
 	if response.StatusCode == http.StatusNotModified {
 		return nil
 	}
-	log.Printf("[DEBUG]: url resp status code %s\n", response.Status)
+
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("Error while downloading %s: %d", i.url.String(), response.StatusCode)
+
+	}
 	return copier(response.Body)
 }
 
--- a/libvirt/utils_volume.go
+++ b/libvirt/utils_volume.go
@@ -169,6 +169,7 @@ func isQCOW2Header(buf []byte) (bool, error) {
 
 // number of download retries on non client errors (eg. 5xx)
 const maxHTTPRetries int = 3
+const retryWaitSec = 2
 
 func (i *httpImage) Import(copier func(io.Reader) error, vol libvirtxml.StorageVolume) error {
 	client := &http.Client{}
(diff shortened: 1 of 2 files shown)


libvirt/utils_volume.go
```diff
@@ -167,6 +167,10 @@ func isQCOW2Header(buf []byte) (bool, error) {
 	return false, nil
 }
 
+// number of download retries on non client errors (eg. 5xx)
+const maxHTTPRetries int = 3
+const retryWaitSec = 2
```
MalloZup: `super small nitpick` why the 2nd var we omit the type and in the first we have the type?  Can we have some coeherence along the vars ? :cat: 

( i think also we could move this variables *inside the the `scope` of  method `Import`, since seeing them here i would think they are global and shared along others methods/functions

libvirt/utils_volume.go
```diff
@@ -179,18 +183,30 @@ func (i *httpImage) Import(copier func(io.Reader) error, vol libvirtxml.StorageV
...
-		return nil
+		log.Printf("[DEBUG]: url resp status code %s (retry #%d)\n", response.Status, retryCount)
+		if response.StatusCode == http.StatusNotModified {
+			return nil
+		} else if response.StatusCode == http.StatusOK {
+			return copier(response.Body)
+		} else if response.StatusCode < 500 {
+			break
+		} else {
+			// retry
```
MalloZup: for this comment here i would expect that either we add more value to it, saying dunno why we are retrying and which behaviour this can happen like `Retry HTTP downloads for 5xx errors` or we can just omit ( because a programmer could read the `retrycount` var) and directly thinking there is a retry

libvirt/utils_volume.go
```diff
@@ -179,18 +183,30 @@ func (i *httpImage) Import(copier func(io.Reader) error, vol libvirtxml.StorageV
...
+			if retryCount < maxHTTPRetries {
+				time.Sleep(retryWaitSec * time.Second)
+			}
+		}
 	}
-	log.Printf("[DEBUG]: url resp status code %s\n", response.Status)
-	return copier(response.Body)
+	return fmt.Errorf("Error while downloading %s: %v", i.url.String(), response)
```
MalloZup: we could maybe say `unexpected Error while downloading ..` and differentiate the log,  because we are returning when we don't catch the error

dmacvicar: I don't think any of these errors is "expected".

dmacvicar:
Incorporated latest feedback from @MalloZup. Waiting for tests.

PR #516, opened 2019-01-03

abbbi:
Small patch that allows to define a disks cache mode as documented here:

 https://doc.opensuse.org/documentation/leap/virtualization/html/book.virt/cha.cachemodes.html


abbbi:
-> some FMT errors .. will resubmit.

PR #517, opened 2019-01-03

abbbi:
Small patch to allow setting of disk cache mode as described here: https://doc.opensuse.org/documentation/leap/virtualization/html/book.virt/cha.cachemodes.html

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -158,6 +158,10 @@ func resourceLibvirtDomain() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
+						"cache": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
 					},
 				},
 			},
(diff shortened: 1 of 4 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -158,6 +158,10 @@ func resourceLibvirtDomain() *schema.Resource {
 							Type:     schema.TypeString,
 							Optional: true,
 						},
+						"cache": {
```
MalloZup: Add default value here

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/CONTRIBUTING.md

libvirt/resource_libvirt_domain.go
abbbi: As i understand it, a default value is not needed here. The default cache mode from the underlying components will be used if the setting is not available in the domains disk xml:

"If no cache mode is specified, an appropriate default cache mode is used. "

that mode may be chosen by qemu then.

MalloZup: OK I will look later thx for clarification, was on mobile phone

MalloZup:
@abbbi I will relook it later on, thx

MalloZup:
@abbbi can you remove the 2nd commit? thx. I will have look next week.

abbbi:
Shall be gone by now.

abbbi:
will resubmit with testacc.

PR #518, opened 2019-01-03
Static routes for networks

inercia:
Static routes for networks. Users could now specify static routes in a network like this:

```hcl
    ...
    # (Optional) one or more static routes.
    # "cidr" and "gateway" must be specified. The format is:
    routes = [
    {
        cidr = "10.17.0.0/16"
        gateway = "10.18.0.2"
    },
    ...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
 LDFLAGS += -X main.version=$$(git describe --always --abbrev=40 --dirty)
 
+# default  args for tests
+TEST_ARGS_DEF := -covermode=count -coverprofile=profile.cov
+
 default: build
 
 build: gofmtcheck golint vet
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -8,12 +11,29 @@ build: gofmtcheck golint vet
...
+# unit tests
+# usage:
+# - run all the unit tests: make test
+# - run some particular test: make test TEST_ARGS="-run TestAccLibvirtDomain_Cpu"
 test:
-	go test -v -covermode=count -coverprofile=profile.cov ./libvirt
+	go test -v $(TEST_ARGS_DEF) $(TEST_ARGS) ./libvirt
 	go test -v .
+
+# acceptance tests
```
MalloZup: nice one here. can you update also this doc: https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/CONTRIBUTING.md#running-the-tests-testacc

inercia: Done.

inercia:
@MalloZup I have updated the PR, implementing the `Read` part so Terraform can know about the current static routes. I don't really know much about how Terraform deals with `Update`, but I guess Terraform will know if things need to be updated when it can `Read` stuff, right?

MalloZup:
@inercia thx i will have look later on this.and let u know :dog: :national_park:

inercia:
It seems routes cannot be `Update`d, so we can skip our concerns about not implementing that part...

dmacvicar:
So, given that we use data sources for other network configurations, was that evaluated for this feature?

MalloZup:
Thank you for remark, 🦄 I opened an issue regarding this https://github.com/dmacvicar/terraform-provider-libvirt/issues/539

Issue #349, opened 2018-08-02
Configure VNC listen address


congaframe:
#  Version Reports:

### Distro  version of host:
Debian 8

### Terraform Version Report
(Provided by running `terraform -v`.)
0.11.7
### Libvirt version

```virsh --version```
1.2.9
### terraform-provider-libvirt plugin version (git-hash)
0.4.1
``` git log```
___
# Description of Issue/Question
Hi,
How can I specify the listen address for VNC?
It is currently listen to 127.0.0.1

My config is: 
```
  graphics {
    type = "vnc"
    listen_type = "address"
    autoport = "true"
  }
```
Thank you
### Setup
(Please provide the full **main.tf** file for reproducing the issue (Be sure to remove sensitive info)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

___
# Additional Infos:

Do you have SELinux or Apparmor/Firewall enabled? Some special configuration?
Have you tried to reproduce the issue without them enabled?

MalloZup:
@congaframe ciao thx for issue.
yop we just didn't implement that.  ( i added some details later)
What woud be your use-case just for curiosity?


for contributors interested look here:
starting from here: 
https://godoc.org/github.com/libvirt/libvirt-go-xml#DomainGraphicListener

and here: https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_domain.go#L1020

brgalloway:
> @congaframe ciao thx for issue.
> yop we just didn't implement that. ( i added some details later)
> What woud be your use-case just for curiosity?
> 
A specific use case for me is that the KVM hypervisor is hosted on a Ubuntu Server with no GUI. In order to get a console session you would have to find the new VM's listening port, and then ssh into the host server to forward the ports over an ssh tunnel. If you configure a specific listing ip address you can then just open a VNC Viewer session. All of my sessions for console are saved in VNC so I automatically know what the next available port will be for any new VM's.

congaframe:
hey, 
Our use-case is at the moment we need to login to the console in order to configure the VM, such as networking configuration for SSH. or sometimes we need to troubleshoot the VMs if ssh/network isn't working properly .

MalloZup:
fixed

PR #519, opened 2019-01-03
#349 Allow Configure VNC listen address

abbbi:
As statet in #349 it shall be possible to define a listener address for the VNC type listener.

This is handy in case you want  users allow to access the virtual machines you deployed via VNC connection to a special IP address on the host system. Defaults to 127.0.0.1 like before.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,19 @@
 # How to contribute to terraform-libvirt-plugin:
 
+## Workflow/requirements for a PR:
+
+- Undocumented feature do not exist.
+Document each new feature.
+- Untested features do not exist.
+To ensure that what we code really works, relevant flows should be covered via acceptance tests.
+So when thinking about a contribution, also think about testability. All tests can be run local without the need of CI. Have a look at the Testing section (later on this page)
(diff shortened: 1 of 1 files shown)


abbbi:
Example:

  graphics {
    type = "vnc"
    listen_type = "address"
    autoport = "true",
    listen_address = "192.168.20.2"
  }

MalloZup:
@abbbi  thx for PR. i will try this out next week. sofar looks good, some tests-acc for this is needed for longterm maintainance.

website/docs/r/domain.html.markdown
```diff
@@ -408,10 +408,15 @@ The block supports:
 * `type` - the type of graphics emulation (default is "spice")
 * `autoport` - defaults to "yes"
 * `listen_type` - "listen type", defaults to "none"
+* `listen_address` - (Optional) IP Address where the VNC listener should be started if
+`listen_type` is set to `address`. Defaults to 127.0.0.1
```
inercia: I was wondering if it wouldn't be better to group these `listen_*` parameters in a

```hcl
   ...
   listen {
      type = "address"
      address = "127.0.0.1"
   }
   ...
```

PR #521, opened 2019-01-04
Allow setting of network card model

abbbi:
I might have the need to deploy some systems that do not bring virtio specific implementations and as such need other network card emulation models such as e1000.

Patch allows to set model type for network card.


MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/pull/525

MalloZup:
@abbbi  i have reverted this PR because it was introducing regression on the testsacc.

Feel free to reopen a PR with testacc for this and fix what is failing on that PR . thx for your time and reviw to you and @inercia

abbbi:
i have absolutely no idea why the testacc is failing currently.
It sets default to virtio, but it seems during testacc, this default is not used:

```
--- FAIL: TestAccLibvirtDomain_NetworkInterface (11.70s)
    testing.go:538: Step 0 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        UPDATE: libvirt_domain.xfsdcxrp4d
          network_interface.0.model: "" => "virtio"
          network_interface.1.model: "" => "virtio"
 [..]
          network_interface.0.hostname = 
          network_interface.0.mac = A6:E7:6F:96:70:81
          network_interface.0.macvtap = 
          network_interface.0.model = 
          network_interface.0.network_id = 6723b489-ce88-46e6-8a88-f8f65dded26c
          network_interface.0.network_name = default

```

abbbi:
OK, stuff was missing in the read function.

=== RUN   TestAccLibvirtDomain_NetworkInterface
--- PASS: TestAccLibvirtDomain_NetworkInterface (11.90s)

dmacvicar:
The whole reason we implemented XSLT support is because of #350 , which we decided against. So why was this merged?

dmacvicar:
I think schema changes require a good discussion, because they have implications.

May be I am biased, but I don't think both the codebase and the schema will scale if we add to the schema every tweak or exotic feature some user needs is added. Feel free to argue against, but until know I haven't seen the _cons_ outweighing the benefits.

Meanwhile we should document the general vision for the schema and the provider and establish a more thoughtful process for schema changes. We do need maintainers to share and be able to review based on that vision. The vision can be discussed and changed, but right now PR are not being merged in a consistent way.

PR #524, opened 2019-01-07
fix documentation, scsi needs to be boolean

abbbi:
hi,

scsi = "yes" wont work with my terraform version, it complains:

`Error: libvirt_domain.domain: disk.0.scsi: cannot parse '' as bool: strconv.ParseBool: parsing "yes": invalid syntax`

value needs to be boolean.
Fix documentation.


MalloZup:
thx

PR #530, opened 2019-01-07

abbbi:
hi,

see #517 , now comes with testacc that checks driver cache set.

### DIFF
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -142,7 +142,7 @@ func TestAccLibvirtDomain_VolumeTwoDisks(t *testing.T) {
 		}
 
 		disk {
-			volume_id = "${libvirt_volume.%s.id}"
+		    volume_id = "${libvirt_volume.%s.id}",
 		}
 	}`, randomVolumeName, randomVolumeName, randomVolumeName2, randomVolumeName2, randomDomainName, randomDomainName, randomVolumeName, randomVolumeName2)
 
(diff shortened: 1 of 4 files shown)


libvirt/resource_libvirt_domain_test.go
```diff
@@ -176,6 +176,49 @@ func TestAccLibvirtDomain_VolumeTwoDisks(t *testing.T) {
...
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckLibvirtDomainDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckLibvirtDomainExists("libvirt_domain."+randomDomainName, &domain),
+					testAccCheckLibvirtVolumeExists("libvirt_volume."+randomVolumeQCOW2, &volumeQCOW2),
+					// Check that each disk has the appropriate cache
```
MalloZup: with this comment we state `each` disk but we are just testing one disk.

We should test that `cache` is nil for the normal or remove the comment because is contradictory

abbbi:
Depending on the feature set, i would expect something like this to be settable via terraform config file. After call the cache setting may have several impact on performance on the systems that iam deploying.

Lets take for example a nested ESXi i want to deploy:

1) needs video type vga, otherwise it wont boot
2) needs network card e1000 for networking
3) needs an SATA disk instead of default virtio 

This means i would have to prepare an XSL definition to change those things, and even then,
im not sure how an XSLT looks like that would change several settings for several components,
lets say i want the first disk to have disk type SATA and disk Cache mode writeback, instead of
the second disk which is on LVM and should not use any cache...

I think using XSLT is fine here, but makes things really complicated.

Thats why iam looking for those things to work out in the plain terraform configuration, so other people
might get an glimpse of how everything works, instead of having to look at several xslt stuff then..

Maybe it would help or make things easier if someome provides more examples of how to change Vm
attributes via XSLT.

MalloZup:
Closing, thx @abbbi for the awesome work

PR #532, opened 2019-01-08
allow setting of video card type

abbbi:
Usually this is something one might not think of, but there are certain systems that have trouble with the default video card set by libvirt (cirrus). For example ESXi will be stuck in a endless boot loop shortly after GRUB loading... It however works with video card type VGA.

Simple PR introduces posiblity to set video card type to VGA. There might be more options like ram/vram which are good with default values, so i ignored them. Maybe its good to add more options like 3d acceleration if needed. Comes with test case but does not re-read upon VM change, which is sufficient for my needs.

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -251,6 +251,19 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
 	return nil
 }
 
+func setVideo(d *schema.ResourceData, domainDef *libvirtxml.Domain) error {
+	prefix := "video.0"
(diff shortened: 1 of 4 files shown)


dmacvicar:
I missed this, but I regret we merged this. We discussed at length with @MalloZup did not merged setting NIC models because we can't see where this will stop.

We can't support every single hardware setting in the schema. It will not work unless we do a 1:1 mapping between HCL. Otherwise we will pollute our schema.

I want to highlight again, we don't want to merge every setting that is contributed unless there is a very common use-case, until we decide as a group that we will consistently just go that direction.

PR #533, opened 2019-01-08
#349 Allow Configure VNC listen address #519

abbbi:
as last PR, now with test ACC. Changing syntax  to

```
listen {
 address ..
}
```
will probably break already existing configurations, so i left the syntax in place.


MalloZup:
I will post some comment on this later. Abbi thx for pr good job but of course can be improved :)

inercia:
@abbbi Maybe you could have continued work using your previous PR (and `git push -f`ing) instead of opening a new one (that way we could keep the discussion history), but anyway thanks for your PRs. Merging.

MalloZup:
before merging it i will test it. @abbbi thx. I opened a follow-up issue, if you have motivation to look at it.

I think in such cases we can say that is not a mandatory for this.

PR #553, opened 2019-02-17
Switch to use Go modules

dmacvicar:
Until now, we have been using glide to manage dependencies.

I don't think Go modules need an introduction, but [here is a good place to start](https://blog.golang.org/modules2019).

This PR moves the dependency handling to use Go 1.11 modules.

Worth mentioning, other providers like AWS already moved. It should not affect the end contributor, as we also vendor dependencies (`go mod vendor`), but it will give the ability to develop outside of `GOPATH`, which is very convenient.

The PR also refactors the Makefile/travis chain in order to build using modules with the vendor directory inside travis. I also moved some of the lint/fmt checks from the travis scripts into the main Makefile after realizing the mess was so big having useless scripts inside travis/ that we were *NOT* actually running unit tests inside Travis :angry:. We do now.

There are still a bunch of other scripts there to cleanup, so in a later PR I may end removing the travis directory completely.


vendor/modules.txt
```diff
@@ -242,6 +243,7 @@ golang.org/x/text/transform
 golang.org/x/text/secure/bidirule
 golang.org/x/text/unicode/bidi
 # golang.org/x/tools v0.0.0-20180917221912-90fa682c2a6e
+golang.org/x/tools/cover
```
MalloZup: I think we could remove this dép since we don't use it anymore

dmacvicar: Looks like this is an indirect dependency

dmacvicar:
I think it is ready for review now.

dmacvicar:
Tested inside GOPATH with GO111MODULES=off

dmacvicar:
Packages are also building correctly. They build with classical GOPATH + vendor/ by using
`GO111MODULES=off` and `GOPROXY=off`

Issue #502, opened 2018-12-17
Not possible to leave network card without ip address


dmacvicar:
## Full example
```hcl
provider "libvirt" {
    uri = "qemu:///system"
}

resource "libvirt_volume" "my2_qcow2" {
  name = "my2_qcow2-${count.index}"
  count = 2
  pool = "default"
  source = "/home/some-image.qcow2"
  format = "qcow2"
}

resource "libvirt_network" "vm_network" {
   name = "inet"
   addresses = ["10.0.1.0/24"]
   dhcp {
		enabled = false
	}
}

resource "libvirt_domain" "domain-sle" {
  name = "sle-terraform-${count.index}"
  memory = "512"
  vcpu = 1
  count = 2 

  network_interface {
    network_id = "${libvirt_network.vm_network.id}"
  }

  network_interface {
    network_id = "${libvirt_network.vm_network.id}"
  }
    

  console {
    type        = "pty"
    target_port = "0"
    target_type = "serial"
  }

  console {
      type        = "pty"
      target_type = "virtio"
      target_port = "1"
  }
 

  disk {
   volume_id = "${libvirt_volume.my2_qcow2.*.id[count.index]}"
  }

  graphics {
    type = "spice"
    listen_type = "address"
    autoport = "true"
  }
}
```

### Details
Given network:

```hcl
resource "libvirt_network" "vm_network" {
   name = "inet"
   addresses = ["10.0.1.0/24"]
   dhcp {
		enabled = false
	}
}
```

And network interface:

```hcl
  network_interface {
    network_id = "${libvirt_network.vm_network.id}"
  }
```

It is not possible to leave the IP address unconfigured, which is a common case for QA teams.

````
libvirt_domain.domain-leap15: Cannot map 'leap15-terraform': we are not waiting for DHCP lease and no IP has been provided
```

This could be a check/leftover from the initial times, where the provider tried to emulate a cloud-like environment and always have a way to access the domain, however, this may not be necessary.

asmorodskyi:
this is problematic code from `libvirt/domain.go` :
```
// no IPs provided: if the hostname has been provided, wait until we get an IP
					wait := false
					for _, iface := range *waitForLeases {
						if iface == &netIface {
							wait = true
							break
						}
					}
					if !wait {
						return fmt.Errorf("Cannot map '%s': we are not waiting for DHCP lease and no IP has been provided", hostname)
					}
```

getting to `waitForLeases` is controlled by 

```
[...]

PR #556, opened 2019-02-22
Enable to Leave network card without ip address

MalloZup:
fix issue #502

Testacc OK and tested With Ubuntu domain we Don't bail out with error

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -676,17 +676,16 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
 							break
 						}
 					}
-					if !wait {
-						return fmt.Errorf("Cannot map '%s': we are not waiting for DHCP lease and no IP has been provided", hostname)
-					}
-					// the resource specifies a hostname but not an IP, so we must wait until we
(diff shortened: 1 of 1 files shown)


go.sum
```diff
@@ -1,20 +1,16 @@
-cloud.google.com/go v0.26.0 h1:e0WKqKTd5BnrG8aKH3J3h+QvEIQtSUcf2n5UZ5ZgLtQ=
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
MalloZup: Regarding go.sum it can make prs unclean. Maybe we could ignore the file

dmacvicar: No. You should set your `-mod` to `vendor` or `readonly` manually or in `GOFLAGS` and not checkin go.sum changes unless you are updating  dependencies.

https://github.com/golang/go/wiki/Modules

MalloZup: ok i will update it

MalloZup: thx

MalloZup: done

PR #567, opened 2019-03-19
Switch back to using File paths for disk volumes

### LINKED ISSUE #126: Using disk type file instead of type volume would fix apparmor permissions

electrofelix:
Prefer resolving disk definitions to file paths for libvirt domain
creation to allow libvirt security helper utilities such as
virt-aa-helper correctly determine the files upon which to apply the
necessary security policies required for non-root user usage.

Fixes #126

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -428,19 +428,9 @@ func setDisks(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *lib
 				return fmt.Errorf("Can't retrieve volume %s: %v", volumeKey.(string), err)
 			}
 
-			diskVolumeName, err := diskVolume.GetName()
+			diskVolumeFile, err := diskVolume.GetPath()
 			if err != nil {
-				return fmt.Errorf("Can't retrieve name for volume %s", volumeKey.(string))
-			}
-
-			diskPool, err := diskVolume.LookupPoolByVolume()
-			if err != nil {
-				return fmt.Errorf("Can't retrieve pool for volume %s", volumeKey.(string))
-			}
-
-			diskPoolName, err := diskPool.GetName()
-			if err != nil {
-				return fmt.Errorf("Can't retrieve name for pool of volume %s", volumeKey.(string))
+				return fmt.Errorf("Can't retrieve volume file %s", volumeKey.(string))
 			}
 
 			// find out the format of the volume in order to set the appropriate
@@ -469,9 +459,8 @@ func setDisks(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *lib
 			}
 
 			disk.Source = &libvirtxml.DomainDiskSource{
-				Volume: &libvirtxml.DomainDiskSourceVolume{
-					Pool:   diskPoolName,
-					Volume: diskVolumeName,
+				File: &libvirtxml.DomainDiskSourceFile{
+					File: diskVolumeFile,
 				},
 			}
 		} else if rawURL, ok := d.GetOk(prefix + ".url"); ok {
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -768,9 +768,9 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
 				"file": diskDef.Source.File,
 			}
 		} else if diskDef.Source.File != nil {
-			// LEGACY way of handling volumes using "file", which we replaced
-			// by the diskdef.Source.Volume once we realized it existed.
-			// This code will be removed in future versions of the provider.
+			// Volumes need to be defined in libvirt XML using File paths
+			// to allow libvirt security helper utilities correctly identify the
+			// storage to apply the correct access polices
 			virVol, err := virConn.LookupStorageVolByPath(diskDef.Source.File.File)
 			if err != nil {
 				return fmt.Errorf("Error retrieving volume for disk: %s", err)
@@ -786,6 +786,9 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
 				"volume_id": virVolKey,
 			}
 		} else {
+			// Compatibility way of handling volumes from XML, which is replaced
+			// by switching to use file paths to ensure libvirt security helper
+			// utilities can apply the correct access policies.
 			pool, err := virConn.LookupStoragePoolByName(diskDef.Source.Volume.Pool)
 			if err != nil {
 				return fmt.Errorf("Error retrieving pool for disk: %s", err)


electrofelix:
So this is just a very rudimentary pass at trying to help with #126, I expect that some effort may be needed around testing/validating and maybe an acceptance test.

Would need some guidance

MalloZup:
try out to run `make testacc` if this works, imho we should have a test for automating recognition already.

electrofelix:
@MalloZup it should be transparent, in that this focuses on only changing how the XML is written out, and the existing code has support for reading in XML containing either File or Volume definitions, so it should mean only when domains are updated should the disk info change to file paths instead of volumes.

I'll look spin up a test VM to run through the full set of tests to be sure, it seems to pass the important ones when running via a go docker image, but anything that needs to access various files blows up and I'd like to make it's isolated from my workstation to be run in a repeatable manner

dmacvicar:
How does this work with other volume backends? I understand this way would only work with pools of type directory, right?

blaggacao:
I believe this is related to https://github.com/dmacvicar/terraform-provider-libvirt/issues/561

blaggacao:
I'm using this fix in a custom build, thanks @electrofelix ! It would be nice to soon see a verified version to fix #126 incorporated in master...

dmacvicar:
> I'm using this fix in a custom build, thanks @electrofelix ! It would be nice to soon see a verified version to fix #126 incorporated in master...
> 
> EDIT: https://gist.github.com/blaggacao/25d08af66b9678f07a35a95e6b71644f There is too much hacking to get this working, still.

This is not a fix. See https://github.com/dmacvicar/terraform-provider-libvirt/issues/126#issuecomment-480597050

MalloZup:
Although we might want to fix apparmor issue in future (if we can fix it ourside, need to be researched), I agree that merging this would compromise the possibility to the provider to support other types of pool, and this is in term of  design a big limitation which we shouldn't take

blaggacao:
As to politics: Is a branched build with hacks included a good combination of the _wise_ and the _kind_? I understand this reduces desired pressure on upstream, but also takes users out of the quenching zone.

dmacvicar:
> As to politics: Is a branched build with hacks included a good combination of the _wise_ and the _kind_? I understand this reduces desired pressure on upstream, but also takes users out of the quenching zone.

See https://github.com/dmacvicar/terraform-provider-libvirt/issues/126#issuecomment-480629345

If someone is going to maintain that branch/build, and do all the work keeping up to date, then go ahead. I can even help you setting it up in OBS.

However, there will be no hacks in master, at least not until we understand the situation, and there will be no rush for that. At first glance it looks like it is not that hard to fix the helper to do the right thing.

I am also confused why internal functions like:

```c
const char *
virDomainDiskGetSource(virDomainDiskDef const *def)
{
    return def->src->path;
}
```

Return a single path, when source has attributes like:

```xml
<source pool='default' volume='volume-test'/>
```

But I still think this is just API leftovers from old times. Checking in the helper code for disk type volume, then attempting to get a `virStorageSourcePoolDef`from there, get the volume definition from the pool, and then get the _target_ from that _virStorageVolDef_, which has the same type _virStorageSource_ the helper deals with right now (as the volumes end also mapped to files).

```c
struct _virStorageVolDef {
    char *name;
    char *key;
    int type; /* virStorageVolType */

    bool building;
    unsigned int in_use;

    virStorageVolSource source;
    virStorageSource target;
};
```

```c
security/virt-aa-helper.c
967:        if (!virStorageSourceHasBacking(disk->src))
```

So it looks like everything is about adding the ability, for disk of type _volume_, to the helper to do the same it does on `disk->src` to do it in the `vol->target` of the volume retrieved using the `virStorageSourcePoolDef` pool/name pair.

blaggacao:
Here would be a search query containing `virt-aa-helper`: https://bugzilla.redhat.com/buglist.cgi?component=libvirt&longdesc=virt-aa-helper&longdesc_type=allwordssubstr&product=Virtualization%20Tools&query_format=advanced

It looks (to me) like your observation is not yet reported. But I can be as quite be wrong.

zeenix:
@MalloZup It's been almost a month you marked this as stale, maybe time to close it then?

MalloZup:
@zeenix  thx I think this PR as such can be closed. We can conserve the issue perhaps to conserve log.

Allthough I agree with the points raised by @dmacvicar . 

Closing as stale and because it lacks general implementation , it is a small-case fix which will impact the whole provider.

PR #571, opened 2019-03-21
Add storage pool resource

### LINKED ISSUE #435: support for storage pool resource
#  Version Reports:

### Distro  version of host:

Ubuntu Bionic 18.04

### Terraform Version Report

v0.11.8

### Libvirt version

4.7.0

df6aed4d9af0493e09454746549a5e95df9a2505

___

There doesn't seem to be any support for managing storage pools or any mentions regarding it. I've hacked together some very basic support in this provider for creating and deleting a default pool but I don't want to spend too much time on it if there is maybe a reason for it not being supported?

Would be happy to contribute the support for it, at least for basic dir and filesystem pool types.

zeenix:
This patch adds a resource type for libvirt storage pool.

Known issue:

The tests `TestAccLibvirtPool_ManuallyDestroyed` and 'TestAccLibvirtPool_UniqueName' currently fail. I failed to figure out why yet.

Fixes #435.
---
As obvious from the title, this is still **WIP**. I just wanted to publish it already to get possible early feedback and maybe help with figuring out why exact the tests are not passing. Next steps:

- [x] Add `update` implementation.
- [x] Add docs.
- [x] Fix the failing tests
  - [x] UniqueName
  - [x] ManuallyDestroyed

### DIFF
--- a/libvirt/pool.go
+++ b/libvirt/pool.go
@@ -1,0 +1,115 @@
+package libvirt
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/resource"
+	libvirt "github.com/libvirt/libvirt-go"
+)
+
+const (
+	poolExistsID    = "EXISTS"
+	poolNotExistsID = "NOT-EXISTS"
+)
+
+// poolExists returns "EXISTS" or "NOT-EXISTS" depending on the current pool existence
+func poolExists(virConn *libvirt.Connect, uuid string) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		pool, err := virConn.LookupStoragePoolByUUIDString(uuid)
+		if err != nil {
+			if err.(libvirt.Error).Code == libvirt.ERR_NO_STORAGE_POOL {
+				log.Printf("Pool %s does not exist", uuid)
+				return virConn, "NOT-EXISTS", nil
+			}
+			log.Printf("Pool %s: error: %s", uuid, err.(libvirt.Error).Message)
+		}
+		if pool != nil {
+			defer pool.Free()
+		}
+		return virConn, poolExistsID, err
+	}
+}
+
+// poolWaitForExists waits for a storage pool to be up and timeout after 5 minutes.
+func poolWaitForExists(virConn *libvirt.Connect, uuid string) error {
+	log.Printf("Waiting for pool %s to be active...", uuid)
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{poolNotExistsID},
+		Target:     []string{poolExistsID},
+		Refresh:    poolExists(virConn, uuid),
+		Timeout:    1 * time.Minute,
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		log.Printf("%s", err)
+		return fmt.Errorf("unexpected error during pool creation operation. The operation did not complete successfully")
+	}
+	return nil
+}
+
+// poolWaitDeleted waits for a storage pool to be removed
+func poolWaitDeleted(virConn *libvirt.Connect, uuid string) error {
+	log.Printf("Waiting for pool %s to be deleted...", uuid)
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{poolExistsID},
+		Target:     []string{poolNotExistsID},
+		Refresh:    poolExists(virConn, uuid),
+		Timeout:    1 * time.Minute,
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		log.Printf("%s", err)
+		return fmt.Errorf("unexpected error during pool destroy operation. The pool was not deleted")
+	}
+	return nil
+}
+
+// deletePool deletes the pool identified by `uuid` from libvirt
+func deletePool(client *Client, uuid string) error {
+	virConn := client.libvirt
+	if virConn == nil {
+		return fmt.Errorf(LibVirtConIsNil)
+	}
+
+	pool, err := virConn.LookupStoragePoolByUUIDString(uuid)
+	if err != nil {
+		return fmt.Errorf("error retrieving storage pool info: %s", err)
+	}
+
+	poolName, err := pool.GetName()
+	if err != nil {
+		return fmt.Errorf("error retrieving storage pool name: %s", err)
+	}
+	client.poolMutexKV.Lock(poolName)
+	defer client.poolMutexKV.Unlock(poolName)
+
+	info, err := pool.GetInfo()
+	if err != nil {
+		return fmt.Errorf("error retrieving storage pool info: %s", err)
+	}
+
+	if info.State != libvirt.STORAGE_POOL_INACTIVE {
+		err := pool.Destroy()
+		if err != nil {
+			return fmt.Errorf("error deleting storage pool: %s", err)
+		}
+	}
+
+	err = pool.Delete(0)
+	if err != nil {
+		return fmt.Errorf("error deleting storage pool: %s", err)
+	}
+
+	err = pool.Undefine()
+	if err != nil {
+		return fmt.Errorf("error deleting storage pool: %s", err)
+	}
+
+	return poolWaitDeleted(client.libvirt, uuid)
+}
--- a/website/docs/r/pool.html.markdown
+++ b/website/docs/r/pool.html.markdown
@@ -1,0 +1,56 @@
+---
+layout: "libvirt"
+page_title: "Libvirt: libvirt_pool"
+sidebar_current: "docs-libvirt-pool"
+description: |-
+  Manages a storage pool in libvirt
+---
+
+# libvirt\_pool
+
+Manages a storage pool in libvirt. Currently only directory-based storage pool are supported. For more information on
+storage pools in libvirt, see [the official documentation](https://libvirt.org/formatstorage.html).
+
+**WARNING:** This is experimental API and may change in the future.
+
+## Example Usage
+
+```hcl
+# A pool for all cluster volumes
+resource "libvirt_pool" "cluster" {
+  name = "cluster"
+  type = "dir"
+  path = "/home/user/cluster_storage"
+}
+
+resource "libvirt_volume" "opensuse_leap" {
+  name = "opensuse_leap"
+  pool = "${libvirt_pool.cluster.name}"
+  source = "http://download.opensuse.org/repositories/Cloud:/Images:/Leap_42.1/images/openSUSE-Leap-42.1-OpenStack.x86_64.qcow2"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) A unique name for the resource, required by libvirt.
+* `type` - (Required) The type of the pool. Currently, only "dir" supported.
+* `path` - (Optional) The directory where the pool will keep all its volumes. This is only relevant to (and required by)
+                      the "dir" type pools.
+
+### Altering libvirt's generated pool XML definition
+
+The optional `xml` block relates to the generated pool XML.
+
+Currently the following attributes are supported:
+
+* `xslt`: specifies a XSLT stylesheet to transform the generated XML definition before creating the pool. This is used
+  to support features the provider does not allow to set from the schema. It is not recommended to alter properties and
+  settings that are exposed to the schema, as terraform will insist in changing them back to the known state.
+
+See the domain option with the same name for more information and examples.
+
+## Attributes Reference
+
+* `id` - a unique identifier for the resource
--- a/examples/ubuntu/ubuntu-example.tf
+++ b/examples/ubuntu/ubuntu-example.tf
@@ -3,10 +3,16 @@ provider "libvirt" {
   uri = "qemu:///system"
 }
 
+resource "libvirt_pool" "ubuntu" {
+  name = "ubuntu"
+  type = "dir"
+  path = "/tmp/terraform-provider-libvirt-pool-ubuntu"
+}
+
 # We fetch the latest ubuntu release image from their mirrors
 resource "libvirt_volume" "ubuntu-qcow2" {
   name   = "ubuntu-qcow2"
-  pool   = "default"
+  pool   = "${libvirt_pool.ubuntu.name}"
   source = "https://cloud-images.ubuntu.com/releases/xenial/release/ubuntu-16.04-server-cloudimg-amd64-disk1.img"
   format = "qcow2"
 }
@@ -27,6 +33,7 @@ resource "libvirt_cloudinit_disk" "commoninit" {
   name           = "commoninit.iso"
   user_data      = "${data.template_file.user_data.rendered}"
   network_config = "${data.template_file.network_config.rendered}"
+  pool           = "${libvirt_pool.ubuntu.name}"
 }
 
 # Create the machine
--- a/libvirt/helpers_test.go
+++ b/libvirt/helpers_test.go
@@ -71,6 +71,21 @@ func getResourceFromTerraformState(resourceName string, state *terraform.State) 
 
 // ** resource specifics helpers **
 
+// getPoolFromTerraformState lookup pool by name and return the libvirt pool from a terraform state
+func getPoolFromTerraformState(name string, state *terraform.State, virConn libvirt.Connect) (*libvirt.StoragePool, error) {
+	rs, err := getResourceFromTerraformState(name, state)
+	if err != nil {
+		return nil, err
+	}
+
+	pool, err := virConn.LookupStoragePoolByUUIDString(rs.Primary.ID)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("[DEBUG]:The ID is %s", rs.Primary.ID)
+	return pool, nil
+}
+
 // getVolumeFromTerraformState lookup volume by name and return the libvirt volume from a terraform state
 func getVolumeFromTerraformState(name string, state *terraform.State, virConn libvirt.Connect) (*libvirt.StorageVol, error) {
 	rs, err := getResourceFromTerraformState(name, state)
--- a/libvirt/provider.go
+++ b/libvirt/provider.go
@@ -23,6 +23,7 @@ func Provider() terraform.ResourceProvider {
 			"libvirt_domain":         resourceLibvirtDomain(),
 			"libvirt_volume":         resourceLibvirtVolume(),
 			"libvirt_network":        resourceLibvirtNetwork(),
+			"libvirt_pool":           resourceLibvirtPool(),
 			"libvirt_cloudinit_disk": resourceCloudInitDisk(),
 			"libvirt_ignition":       resourceIgnition(),
 		},
--- a/libvirt/resource_libvirt_pool.go
+++ b/libvirt/resource_libvirt_pool.go
@@ -1,0 +1,255 @@
+package libvirt
+
+import (
+	"encoding/xml"
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	libvirt "github.com/libvirt/libvirt-go"
+	"github.com/libvirt/libvirt-go-xml"
+)
+
+func resourceLibvirtPool() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceLibvirtPoolCreate,
+		Read:   resourceLibvirtPoolRead,
+		Delete: resourceLibvirtPoolDelete,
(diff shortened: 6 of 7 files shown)


zeenix:
Actually, after implementing `Update`, I found out that libvirt currently does not support even renaming storage pools so `Update` would fail with an error. So I'm wondering if we want `Update` at all? If we do, it shouldn't be a lot of work to implement it in libvirt but I'm wondering if it's worth it, given that nobody every asked for this feature from libvirt before.

MalloZup:
if `libvirt` doesn't support update we should document it in the doc as limitation imho. thx @zeenix for working on this

zeenix:
@MalloZup Thanks. Will add to docs, which is next.

zeenix:
So I added the docs and I fixed the testcase for unique names. From my side the only thing left is the failing `TestAccLibvirtPool_ManuallyDestroyed` test. After spending a lot of hours on this, I at least figured out why it's failing. The reason is simple: This implementation is not able to cope with pool deletion behind Terrform's back, as it should and that's what the testcase is supposed to test.

What is a big mystery to me still is that `terroform destroy` runs successfully for volumes even if you delete the volume before that but the same is not true for the pool implementation despite the fact that I'm doing exactly the same as Volume implementation is doing (since I copied and updated that code): error out if resource being deleted can not be found.

So I'd really like a second pair of eyes to see what I'm missing here.

MalloZup:
thx @zeenix for your time. we will have look

MalloZup:
I have looked little on the pr code and found out only difference is in the exists volume implementation, there is a try hard look-up function this is spread on all resources. Otherwise you are right. I might to check next week better on it

dmacvicar:
>What is a big mystery to me still is that terroform destroy runs successfully for volumes even if you delete the volume before that but the same is not true for the pool implementation despite the fact that I'm doing exactly the same as Volume implementation is doing (since I copied and updated that code): error out if resource being deleted can not be found.

Best way to triage this is to create a test-case.

dmacvicar:
>I have looked little on the pr code and found out only difference is in the exists volume implementation, there is a try hard look-up function this is spread on all resources.(not in pool) Otherwise you are right. I might to check next week better on it. AFAIK the exist function, should prevent error out, if an user remove a volume manually , and we do terraform apply we shouldn't error out but only update the resources files data in tf state

Those "hard lookups" were, IIRC added very early to workaround race issues in libvirt.

zeenix:
Thanks @MalloZup and @dmacvicar for reviewing.

>     * This is not a _pool resource_, but a _dir pool resource_. There is no way to create something different, and this will not be immediately obvious to the users.

I understood that we can later add support for other types of pools by adding more attributes (e.g `type`) when there is a demand for it, without breaking anything. Is this understanding not correct?

>     * This limitation is hardcoded in the design. _type_ is hardcoded in the XML created. The _dir_ specific setting _path_ leaks to the schema.

If you are talking about `xml` attribute in resource schema, I actually didn't touch that when i copied volume resource code over as the starting point. I can have a look at what that really does/is and fix it. :) If you're not talking about that, I would need more context/pointers. :)

>     * The first issue we will get after merging this, is somebody asking for a different pool type

As long as new attributes/types can be added later, that's not an issue. Otherwise, I can add attributes already for `type` so other types of pools can later be added.

> Those "hard lookups" were, IIRC added very early to workaround race issues in libvirt.

Hm.. do you have any idea why manual deletion doesn't break destroy operation for volume but only for pool?

MalloZup:
@zeenix @dmacvicar :

Imho what we need here is to create a generic interface with 2 layers abstraction, so we can enable plug-gable pools.

As i understand also `libvirt` upstream, some `attributes` varies depending on the `type` of the pool


### the 1st layer would be :

```mandatory```
```hcl
	"name": {
				Type:     schema.TypeString,
				Required: true,
				ForceNew: true,
			},
         "type": {
				Type:     schema.Map,
				Required: true,
				ForceNew: true,
                             ... ( see 2nd layer)
			},

```

`name` is the name of pool.

the `type` would be the type of pool  which will contain other maps, which will be the implementations e.g `directory`, `disk` pools etc , which they will be other maps.

### 2nd layer

```hcl
			"type": {
				Type:     schema.TypeMap,
				Required: true,
				ForceNew: true,
				Elem: &schema.Resource{
					Schema: map[string]*schema.Schema{
						"dir": {
							Type:     schema.TypeMap,
                                                            // implementation here  which will be the fields for the map
							},
						"disk": {
							Type:     schema.TypeString,
                                                           // implementation here
						},
						"scsi": {... 
```



https://www.terraform.io/docs/extend/schemas/schema-types.html#typemap

zeenix:
> Imho what we need here is to create a generic interface with 2 layers abstraction, so we can enable plug-gable pools.

@MalloZup Thanks. Sounds good to me but I could easily be missing something important here. :) So if this sounds good to @dmacvicar as well, I can change my PR accordingly.

zeenix:
@MalloZup I tried with

```
            "type": {     
                Type:     schema.TypeMap,
                Required: true,
                ForceNew: true,
                Elem: &schema.Resource{
                    map[string]*schema.Schema{
                        "dir": {
                            Type: schema.TypeMap,
                            Required: true,
                            Elem: &schema.Resource{
                                Schema: map[string]*schema.Schema{
                                    "path": { 
                                        Type: schema.TypeString,
                                        Required: true,
                                        ForceNew: true,
                                    },
                                },
                            },
                        },
                    },
                },
            },
```

as schema and test case:

```
                resource "libvirt_pool" "%s" {
                    name = "%s"
                    type = {                                                                                                                                                                                                                                                              
                        dir = {
                            path = "%s"
                        }
                    }
                }
```

but i get error: `type (dir): '' expected type 'string', got unconvertible type '[]map[string]interface {}'`. Which part is wrong? The spec or the testcase?

MalloZup:
https://github.com/dmacvicar/terraform-provider-libvirt/blob/6536f78a60151315f3e281d88de8fc127bfbbaa2/libvirt/resource_libvirt_domain_test.go#L684

zeenix:
> afaik you don't need the `=`

Yeah i tried without as well. Still the same.



> take this examples:

I did, since it's the only example of Map in the resource implementations I could find. However, that's more basic since it's a map rather than map of a map.

MalloZup:
@zeenix  Take inspiration from this https://github.com/terraform-providers/terraform-provider-nutanix/blob/03625014f48382e934bda2e4648ff097262e6161/future/volume_group/data_source_nutanix_volume_groups.go.future#L131.

We could use a TypeList/.TypeSet of TypeMaps ( which should be not mandatory or should exclude themselfs).

It might be a terraform limitation/design to don't allow nested maps in this way

zeenix:
@MalloZup I actually thought of List+Map but didn't want to go for that because then we can't mandate one of the types in the schema, i-e people could provide multiple types or even worse, no types under 'types' and it'll be valid spec as per the schema.

Other solutions that come to mind:

1. Separate resource types for each of the pool type. This would be easiest as I just need to rename basically but there will be a bit of code duplication.
1. Separate backend-specific resources corresponding to pool's `target` node (`libvirt_pool_dir_target` in this case) and a `target` field under pool resource to connect them both. This will be a bit of a work, also more complicated for users.

Unless we're ok with the problem with using of List+Map I mentioned above, I'd say we should just bite the bullet and go for idea 1.

MalloZup:
I think that 1 is valid option allthough we should more investigate the maps of maps approach, e. G find out if it is a terraform bug limitation and at least open an issue if it that the case (I did not researched lot on that)

zeenix:
> I think that 1 is valid option

Cool, I'll go for that one then.

> allthough we should more investigate the maps of maps approach, e. G find out if it is a terraform bug limitation and at least open an issue if it that the case (I did not researched lot on that)

Actually I did quite a bit of searching yesterday and the only thing that I found that came close was [this](https://github.com/hashicorp/terraform/issues/2114) (and similar articles/posts) but it's about nested maps in variables and not about how custom resource providers can do this.

zeenix:
Solution 1 implemented. PTALA

MalloZup:
thx @zeenix i have also opened an issue upstream ( hashicorp/terraform#20927).

So recapitulating, ` ManuallyDestroyed` still fail.  Besides the `schema` pb.

To me , i agree it is a trade-off but could be ok. Let's wait for @dmacvicar  opinion on the `schema`. Meanwhile i will try to help you once i have some time on the failing tests.

MalloZup:
https://github.com/hashicorp/terraform/issues/6215
https://github.com/hashicorp/terraform/issues/20927#issuecomment-479953877

:(  I think the only way for us is to have a `libvirt_poo_dir` and `libvirt_pool_type` as we already implemented.

zeenix:
> :( I think the only way for us is to have a `libvirt_poo_dir` and `libvirt_pool_type` as we already implemented.

Oh well, I think it's not too bad actually. When we add another pool resource, much of the code can be reused and moved to a separate file but I don't see the point of doing that right now.

So that brings us to the failing testcase. So nobody has any clue why the manual pool deletion works for volume and not for pool?

Also the CI here doesn't run the added test, since it's passing?

MalloZup:
@zeenix  for the test i actually didn't have the time  for diving it.

about the `ci` on PRs the problems is that we need `libvirt`, which in `travis/circleci` is really problematic to setup. We had in the past anhack in travis but it was not easy to maintain. If you have any idea/improvents in that direction feel free to experiment :grimacing:

zeenix:
> @zeenix for the test i actually didn't have the time for diving it.

Oh ok. I was hoping @dmacvicar might have some insight. I'll ask some other folks around if they could take a look as well.

> about the `ci` on PRs the problems is that we need `libvirt`, which in `travis/circleci` is really problematic to setup.

I'll add to my TODO to look into that. :) Any specific reason it was problematic? If it was using system libvirt, I can imagine how that can be problematic to sandbox. We should use session libvirt.

zeenix:
All test pass now. :)

About the CI, I had a brief look and yes, it does make use of system libvirt, which is problematic to setup on CI env. AFAIK, most tests should work fine on session libvirt connection except for the network creation bits so next week i'll look into reviving the CI, using session libvirt this time around.

libvirt/resource_libvirt_dir_pool.go
```diff
@@ -0,0 +1,240 @@
...
+		return false, fmt.Errorf(LibVirtConIsNil)
+	}
+
+	pool, err := virConn.LookupStoragePoolByUUIDString(d.Id())
+	if err != nil {
+		virErr := err.(libvirt.Error)
+		if virErr.Code != libvirt.ERR_NO_STORAGE_POOL {
+			return false, fmt.Errorf("Can't retrieve pool %s", d.Id())
+		}
+		// does not exist, but no error
```
MalloZup: can you give more info about this context?

libvirt/resource_libvirt_dir_pool.go
```diff
@@ -0,0 +1,240 @@
...
+	// Check whether the storage pool already exists. Its name needs to be
+	// unique.
+	if _, err := client.libvirt.LookupStoragePoolByName(poolName); err == nil {
+		return fmt.Errorf("storage pool '%s' already exists", poolName)
+	}
+	log.Printf("[DEBUG] Pool with name '%s' does not exist yet", poolName)
+
+	poolPath := d.Get("path").(string)
+	poolDef := libvirtxml.StoragePool{
+		// We only support "dir" type at the moment
```
MalloZup: is this comment still relevant in our new format? :thinking:

libvirt/pool.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/hashicorp/terraform/helper/resource"
+	libvirt "github.com/libvirt/libvirt-go"
+)
+
+const (
+	poolExistsID    = "EXISTS"
+	poolNotExistsID = "NOT-EXISTS"
+)
+
+// poolExists returns "EXISTS" or "NOT-EXISTS" depending on the current pool existence
```
MalloZup: question here: are all this methods specific for the `libvirt` type? are we sure for another type they works? or required other flags.

libvirt/resource_libvirt_dir_pool.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	err = pool.Refresh(0)
+	if err != nil {
+		return fmt.Errorf("Error refreshing libvirt storage pool: %s", err)
+	}
+
+	id, err := pool.GetUUIDString()
+	if err != nil {
+		return fmt.Errorf("Error retrieving libvirt pool id: %s", err)
```
MalloZup: i was thinking it since we hardcode the `dir` type withing the implementation, if also the error-msg should be adapted

MalloZup: with "Error retrieving libvirt directory pool id: %s", err)

MalloZup:
@zeenix  regarding the CI can you open an issue and share your insight about this? tia :+1:

libvirt/resource_libvirt_dir_pool.go
zeenix: This is the case where you handle the manual deletion. The comment was C&P with the code from volume code.

libvirt/resource_libvirt_dir_pool.go
zeenix: not :)

libvirt/pool.go
zeenix: I'm not sure I understand the question. This method is for libvirt pool, yes.

libvirt/resource_libvirt_dir_pool.go
zeenix: No, i don't think so. As I said before, quite a bit of this code can be moved to a common pool module, when someone adds another pool implementation.

zeenix: Also i don't think the exact type of the pool is relevant here.

zeenix:
> @zeenix regarding the CI can you open an issue and share your insight about this? tia +1

Done: https://github.com/dmacvicar/terraform-provider-libvirt/issues/579

dmacvicar:
@zeenix I will need some time to review this. Please stay tuned.

libvirt/resource_libvirt_dir_pool.go
MalloZup: yop,  i think we can mark as resolved.

libvirt/pool.go
MalloZup: ok thx

libvirt/resource_libvirt_dir_pool.go
MalloZup: ok thx for clarification

zeenix:
@dmacvicar Sure thing but I hope it's going to be soon. :)

libvirt/pool.go
```diff
@@ -0,0 +1,113 @@
...
+		Pending:    []string{poolExistsID},
+		Target:     []string{poolNotExistsID},
+		Refresh:    poolExists(virConn, uuid),
+		Timeout:    1 * time.Minute,
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		return fmt.Errorf("Error waiting for pool to reach NOT-EXISTS state: %s", err)
```
MalloZup: i was thinking if we could improve this Error message. 

This error is something that an User will read if something goes wrong and it seems more for a devel perspective.

I agree that we can keep it but imho we could improve the UX.

From experience we had in the past for the network IP the same `devel` message and at the end we had lot of issues open because the message was not user friendly enough, so an user could not take an action on it

`` Unexpected error during pool destroy operation. The pool was not deleted

libvirt/pool.go
```diff
@@ -0,0 +1,113 @@
...
+		Pending:    []string{poolNotExistsID},
+		Target:     []string{poolExistsID},
+		Refresh:    poolExists(virConn, uuid),
+		Timeout:    1 * time.Minute,
+		Delay:      5 * time.Second,
+		MinTimeout: 3 * time.Second,
+	}
+
+	if _, err := stateConf.WaitForState(); err != nil {
+		return fmt.Errorf("Error waiting for pool to reach EXISTS state: %s", err)
```
MalloZup: same

libvirt/pool.go
zeenix: Sure, I'll change. Just FYI, this is copy&paste from the similar code for volume resource.

MalloZup: ok thank you zeenix. I think we didn't spot it out. imho we can improve also the other errMsg. But not worth with PR imho. thank you for this :+1:

zeenix: I updated the messages and also sent the exact error to the log. PTALA.

zeenix:
Thanks for reviews, tests and very encouraging words in general @MalloZup. :)

zeenix:
@dmacvicar I understand you're very busy but just wanted to remind about this issue since it's been waiting for more than 2 weeks now.

zeenix:
@dmacvicar Thanks for the review. I'll make the changes. :)

zeenix:
@dmacvicar PTALA.

> Otherwise, apart of that, and trivial stuff like `make test` failing on me while testing the code because of trivial stuff like `vet`, I think this can be merged.

Strange. Test succeed here:

```
$ make testacc TEST_ARGS="-run TestAccLibvirtPool"
./travis/run-tests-acceptance -run TestAccLibvirtPool
+ go test -v -covermode=count -coverprofile=profile.cov -timeout=1200s -run TestAccLibvirtPool ./libvirt
=== RUN   TestAccLibvirtPool_Basic
--- PASS: TestAccLibvirtPool_Basic (13.97s)
=== RUN   TestAccLibvirtPool_ManuallyDestroyed
--- PASS: TestAccLibvirtPool_ManuallyDestroyed (10.10s)
=== RUN   TestAccLibvirtPool_UniqueName
--- PASS: TestAccLibvirtPool_UniqueName (10.04s)
=== RUN   TestAccLibvirtPool_NoDirPath
--- PASS: TestAccLibvirtPool_NoDirPath (0.01s)
PASS
coverage: 6.3% of statements
ok  	github.com/dmacvicar/terraform-provider-libvirt/libvirt	34.140s	coverage: 6.3% of statements
```

website/docs/r/pool.html.markdown
```diff
@@ -0,0 +1,56 @@
...
+# libvirt\_pool
+
+Manages a storage pool in libvirt. Currently only directory-based storage pool are supported. For more information on
+storage pools in libvirt, see [the official documentation](https://libvirt.org/formatstorage.html).
+
+**WARNING:** This is experimental API and may change in the future.
+
+## Example Usage
+
+```hcl
```
MalloZup: feel free to create an example https://github.com/dmacvicar/terraform-provider-libvirt/tree/master/examples

MalloZup:
I will re-review this PR this week and I think we can merge it if nobody is against it. Thx @zeenix . I will check if I have some issues locally with go vet

MalloZup:
so I didn't have any problem with go-vet. However golint is failing and that is why also travis is faling.
```
make lint-check 
go run golang.org/x/lint/golint -set_exit_status ./libvirt .
libvirt/pool.go:49:21: error strings should not be capitalized or end with punctuation or a newline
libvirt/pool.go:68:21: error strings should not be capitalized or end with punctuation or a newline
Found 2 lint suggestions; failing.
```

website/docs/r/pool.html.markdown
zeenix: Instead of creating a separate example for just pool, I think we should add pool to existing examples as to discourage the assumption of "default" pool's existence. As I'm sure I wrote before, there is no "default" pool guaranteed by libvirt on any of it's daemons. The system daemon usually has this pool available, because virt-manager and virt-install (and other tools) create it.

MalloZup: ok we can add the pool creation to the ubuntu.  I can agree with your POV , but we also need to think that each example should showcase a feature and if we add maybe to much new infos, in term of cognition load can be to much.

I'm personally fine with both examples.. Feel free to pick one thx

zeenix: Right. I agree with your POV too and it would be best to keep examples as simple as possible. However, it's pretty annoying for newcomers to not be able to run the examples as is and without pool setup, that will be the experience and IMO that's worse than a bit more of cognitive load.

zeenix: but yeah, for now I'll just add to Ubuntu example. :)

MalloZup: sounds good!

zeenix: Done

zeenix:
@MalloZup Getting closer to end of the week btw. :)

zeenix:
Thanks @MalloZup and @dmacvicar for your reviews and patience with me. :) I'm looking into #579 now. Fingers crossed if i succeed. :)

MalloZup:
@zeenix thank you for this. :+1: welcome

PR #573, opened 2019-03-31
When volume size is not given, set the backing store size as the size

### LINKED ISSUE #561: with version 0.5.1+git.1550869862.4283c564-10.1 (from yesterday) can't boot a VM with a non-base image volume
### Linux distribution

```Debian testing```

### Terraform version

```sh
terraform -v

Terraform v0.11.13
```

```sh
terraform-provider-libvirt -version

terraform-provider-libvirt 0.5.1
Compiled against library: libvirt 4.0.0
Using library: libvirt 5.0.0
Running hypervisor: QEMU 3.1.0
Running against daemon: 5.0.0
```

I have a simple setup with:

- one base volume created from a local file image
- one volume based on the previous one
- one domain which disk is the previous one

When using version 0.5.1 it works like expected. Yesterday I installed 0.5.1+git.1550869862.4283c564-10.1 from binary, and now when applying terraform plan, there is no error but my domain can't boot on its disk with the error "could not read the boot disk".
Without changes in the tf files.

### Setup

(Please provide the full _main.tf_ file for reproducing the issue (Be sure to remove sensitive information)

```
provider "libvirt" {
        uri = "qemu+ssh://user@kvm.example.com/system"
}

resource "libvirt_volume" "debian_stretch_image" {
  name = "debian_stretch_image"
  source = "/tmp/debian-stretch-qemu"
  format = "qcow2"
}

resource "libvirt_volume" "debian_stretch_volume-bastion" {
  name = "debian_stretch_volume-bastion"
  base_volume_id = "${libvirt_volume.debian_stretch_image.id}"
}

resource "libvirt_domain" "terraform_bastion" {
  name = "terraform_bastion"
  vcpu = 1
  memory = 1024
  autostart = true
  disk {
    volume_id = "${libvirt_volume.debian_stretch_volume-bastion.id}"
  }
  network_interface {
    network_name = "default"
    hostname = "bastion"
  }
  graphics {
    type = "vnc"
    listen_type = "address"
  }
}
```

### Steps to Reproduce Issue

```
$ terraform apply
```

Then look at the state of the machine (I use virt-manager for that), I see in the console "Booting from Hard Disk... Boot failed: could not read the boot disk"

dmacvicar:
Fix regression introduced with f86539c365b5bd12374318070894c504a1636c4b

Rework the logic to:

* First set the volume size to the given size, if something was given
* If the size is not given, and there is backing store, use the size of the backing store
* Only then check if the size of the volume is smaller than the backing store

Also, cleanup the testcase and add a check for the size when this one is not specified, but a backing store is given.

Fixes #561

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -164,13 +164,16 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 		volumeDef.Capacity.Unit = "B"
 		volumeDef.Capacity.Value = size
 	} else {
+		// the volume does not have a source image to upload
 
-		// the volume does not have a source image to upload, first handle
-		// whether it has a backing image
-		//
+		// if size is given, set it to the specified value
+		if _, ok := d.GetOk("size"); ok {
+			volumeDef.Capacity.Value = uint64(d.Get("size").(int))
+		}
+
+		//first handle whether it has a backing image
 		// backing images can be specified by either (id), or by (name, pool)
 		var baseVolume *libvirt.StorageVol
-
 		if baseVolumeID, ok := d.GetOk("base_volume_id"); ok {
 			if _, ok := d.GetOk("base_volume_name"); ok {
 				return fmt.Errorf("'base_volume_name' can't be specified when also 'base_volume_id' is given")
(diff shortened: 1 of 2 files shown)


libvirt/resource_libvirt_volume.go
```diff
@@ -196,29 +197,32 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
...
 
-				if backingStoreVolumeDef.Capacity != nil && volumeDef.Capacity.Value < backingStoreVolumeDef.Capacity.Value {
-					return fmt.Errorf("When 'size' is specified, it shouldn't be smaller than the backing store specified with 'base_volume_id' or 'base_volume_name/base_volume_pool'")
-				}
+			backingStoreVolumeDef, err := newDefVolumeFromLibvirt(baseVolume)
+			if err != nil {
+				return err
+			}
+
+			// if the volumen does not specify size, set it to the size of the backing store
```
MalloZup: `volumen` --> `volume`

MalloZup: i think this is a german typo :grin:

PR #582, opened 2019-04-09
Provide a scalable CoreOS example

lhw:
I tried to combine two uses-cases in one more in-depth example using the new support for Ignition.

### DIFF
--- a/examples/coreos/README.md
+++ b/examples/coreos/README.md
@@ -1,0 +1,33 @@
+# CoreOS multi-machine example setup
+
+### Requirements
+* Linux Kernel ~> 4.14
+* Libvirt ~> 3.0
+* QEMU ~> 2.6
+
+This is a relatively simple scalable example using CoreOS as operating system.
+By modifying the `hosts` variable you can kickstart any number of virtual machines
+with their own ignition configuration
+
+
+### Using the QEMU Guest Agent
+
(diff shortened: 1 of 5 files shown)


MalloZup:
Thx @lhw for this, I will read it later on

examples/coreos/README.md
```diff
@@ -0,0 +1,33 @@
+# CoreOS multi-machine example setup
+
+### Requirements
+* Linux ~> 4.14
```
MalloZup: this requirements here are for the `kvm-host`?

MalloZup: e.g i would state `linux-kernel` since `Linux` is generic word imho.

MalloZup:
@lhw i'm ok with this. Thank you for your work. I assume the example you posted here are working so you have my trust there :+1:

examples/coreos/README.md
lhw: That is indeed the linux-kernel requirement. It should be a general requirement for this provider if you create more than one machine with graphics {} at a time it will cause issues.
I mentioned it on the gitter channel but that seems to be mostly dead:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=863266
https://bugzilla.redhat.com/show_bug.cgi?id=1432684

MalloZup:
the travis failure is unrelated to this pr, is something that is travis specific.

thx @lhw merging

PR #587, opened 2019-04-11
[wip]: fix travis internal errors with golang

MalloZup:
using latest `golang` devel version work but with 1.** not. 

I will let the PR open to see if i found better solution


MalloZup:
Travis has fixed it is own issue closing

Issue #583, opened 2019-04-09
Enhancement: Support for tags in libvirt_domain resources


lhw:
Some of the more popular providers offer the tags block in VM resources (e.g. AWS, Azure, vSphere). Quite a few tools make use of these to categorize the machines in internal databases etc.
I thought it would be a good feature to add. I think we could use the `libvirt.DomainMetadata` to add a metadata structure for tf-libvirt. Like

```terraform
resource "libvirt_domain" "example" {
  tags {
    Type = "master"
  }
}
```
```xml
<domain>
  <metadata>
    <terraform:instance xmlns="https://terraform.io">
      <terraform:tag key="Type">master</terraform:tag>
    </terraform:instance>
  </metadata>
</domain
```
Some additional ideas for metadata from [OpenStack Nova](https://specs.openstack.org/openstack/nova-specs/specs/juno/implemented/libvirt-driver-domain-metadata.html)

I wouldn't mind implementing this feature but wanted to request input first in case somebody is already working on something similar etc.

MalloZup:
@lhw thx for this proposal. :+1:  

I think in the `kvm` context which use-case could we have for the metadata field? E.g which info whoul we store in a concrete use case

thx for you time

lhw:
I don't know what you mean about a kvm context but one common use-case in the wild is to tag certain machines as "master" or "worker" in a kubernetes environment for easier export i.e. to an ansible inventory (That would be our use-case). Others use the tags for staging by tagging certain machines for different stages with different configurations. Again then used by other software using the terraform state.

Some advanced usage can also be seen by the [terraform-label project ](https://github.com/cloudposse/terraform-terraform-label)

MalloZup:
thx for clarification! i will have a look. @dmacvicar  cc

MalloZup:
@lhw i'm ok with this feature and to me you could come up with a PR for this.

To me you can start thinking and coming out with a valid proposal about the `schema` of the resource and we can integrate with it.

e.g the tag definitions. etc .

e.g how will look like it in golang ( https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_volume.go#L40)

lhw:
The default implementation is pretty boring:
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_instance.go#L309
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/tags.go#L22
[...]

PR #590, opened 2019-04-30
[Feature] Support for libvirt_domain tags

lhw:
## Reason
Quite a few external tools make use of the virtual machine `tags` to sort them for further use. Currently there is at least support in AWS and Azure instances as well as vSphere machines. An example usage would be creating an ansible inventory.

## Changes
This merge request adds an optional paramater that does not affect existing configurations. Based on the input a custom xml is generated and added to the standard domain metadata. Updates to the tags are done live and to the configuration of the machine.
An existing but unused parameter `metadata` was removed from the `libvirt_domain` but could be kept and be integrated in the future.

A minimal change was made to existing domain unit test. No additional test was added.

## Breaking Changes
None

## Discussion
Currently the custom xml uses the namespace "terraform.io" which might not be appropriate for an unofficial provider. The full github path just seemed a bit too long.

  (Closes: GH-583)

### DIFF
--- a/examples/v0.11/coreos/main.tf
+++ b/examples/v0.11/coreos/main.tf
@@ -70,6 +70,7 @@ resource "libvirt_domain" "coreos-machine" {
     # Requires qemu-agent container if network is not native to libvirt
     wait_for_lease = true
   }
+
   ## mounts filesystem local to the kvm host. used to patch in the
   ## qemu-guest-agent as docker container
   #filesystem {
@@ -78,7 +79,7 @@ resource "libvirt_domain" "coreos-machine" {
   #  readonly = true
   #}
   tags {
-    OS       = "CoreOS"
+    OS = "CoreOS"
     NodeType = "${count.index <= 3 ? "master" : "worker"}"
   }
 }
--- a/examples/v0.11/coreos/main.tf
+++ b/examples/v0.11/coreos/main.tf
@@ -5,7 +5,7 @@ provider "libvirt" {
 
 # -[Variables]-------------------------------------------------------------
 variable "hosts" {
-  default = 1
+  default = 7
 }
 
 variable "hostname_format" {
@@ -38,8 +38,8 @@ resource "libvirt_ignition" "ignition" {
 resource "libvirt_domain" "coreos-machine" {
   count  = "${var.hosts}"
   name   = "${format(var.hostname_format, count.index + 1)}"
-  vcpu   = "1"
-  memory = "2048"
+  vcpu   = "${count.index <= 3 ? 4 : 8}"
+  memory = "${count.index <= 3 ? 4096 : 16384}"
 
   ## Use qemu-agent in conjunction with the container
   #qemu_agent = true
@@ -78,6 +78,10 @@ resource "libvirt_domain" "coreos-machine" {
   #  target = "qemu_docker_images"
   #  readonly = true
   #}
+  tags {
+    OS = "CoreOS"
+    NodeType = "${count.index <= 3 ? "master" : "worker"}"
+  }
 }
 
 # -[Output]-------------------------------------------------------------
--- a/examples/v0.11/coreos/main.tf
+++ b/examples/v0.11/coreos/main.tf
@@ -70,7 +70,6 @@ resource "libvirt_domain" "coreos-machine" {
     # Requires qemu-agent container if network is not native to libvirt
     wait_for_lease = true
   }
-
   ## mounts filesystem local to the kvm host. used to patch in the
   ## qemu-guest-agent as docker container
   #filesystem {
@@ -79,7 +78,7 @@ resource "libvirt_domain" "coreos-machine" {
   #  readonly = true
   #}
   tags {
-    OS = "CoreOS"
+    OS       = "CoreOS"
     NodeType = "${count.index <= 3 ? "master" : "worker"}"
   }
 }
--- a/examples/v0.11/coreos/main.tf
+++ b/examples/v0.11/coreos/main.tf
@@ -78,7 +78,7 @@ resource "libvirt_domain" "coreos-machine" {
   #  target = "qemu_docker_images"
   #  readonly = true
   #}
-  tags {
+  tags = {
     OS = "CoreOS"
     NodeType = "${count.index <= 3 ? "master" : "worker"}"
   }
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -1,6 +1,7 @@
 package libvirt
 
 import (
+	"encoding/xml"
 	"errors"
 	"fmt"
 	"log"
@@ -28,6 +29,18 @@ const domWaitLeaseStillWaiting = "waiting-addresses"
 const domWaitLeaseDone = "all-addresses-obtained"
 
 var errDomainInvalidState = errors.New("invalid state for domain")
+
+// TerraformInstanceXML type
+type TerraformInstanceXML struct {
+	XMLName xml.Name          `xml:"https://terraform.io instance"`
+	Tags    []TerraformTagXML `xml:"tag"`
+}
+
+// TerraformTagXML type
+type TerraformTagXML struct {
+	Key   string `xml:"name,attr"`
+	Value string `xml:",chardata"`
+}
 
 func domainWaitForLeases(domain *libvirt.Domain, waitForLeases []*libvirtxml.DomainInterface,
 	timeout time.Duration, rd *schema.ResourceData) error {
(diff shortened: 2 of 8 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -629,6 +633,20 @@ func resourceLibvirtDomainUpdate(d *schema.ResourceData, meta interface{}) error
 		d.SetPartial("cloudinit")
 	}
 
+	if d.HasChange("tags") {
+		out := TerraformInstanceXML{}
```
lhw: This should also not be empty. domain.GetMetadata() should grab the existing information and merge the tags on top

lhw:
Added the test and made sure it also kept custom metadata when updating.

libvirt/domain.go
```diff
@@ -737,6 +750,57 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
...
+}
+
+func setMetadata(d *schema.ResourceData, domain *libvirtxml.Domain) (err error) {
+	metadata := TerraformInstanceXML{}
+	if err = xml.Unmarshal([]byte(domain.Metadata.XML), &metadata); domain.Metadata.XML != "" && err != nil {
+		return fmt.Errorf("XML Unmarshal Error: %s", err)
+	}
+	if _, ok := d.GetOk("tags"); ok {
+		tags := d.Get("tags").(map[string]interface{})
+		var out string
```
MalloZup: this afaik you can omit it by using `:=` next line

lhw: I use `out` after the error check. with `:=` it would only available inside the if{} context.

MalloZup:
@lhw  we will review your PR soonish thx.

dmacvicar:
This is on my queue, but I haven't yet found proper time. Please be patient.

libvirt/domain.go
```diff
@@ -29,6 +30,18 @@ const domWaitLeaseDone = "all-addresses-obtained"
 
 var errDomainInvalidState = errors.New("invalid state for domain")
 
+// TerraformInstanceXML type
+type TerraformInstanceXML struct {
```
MalloZup: Regarding this and all functions, I'm thinking that it could make sense to create a domain_tags.go file even if small to don't overload the domaingo imho

website/docs/r/domain.html.markdown
```diff
@@ -551,6 +552,17 @@ boot_device {
 }
 ```
 
+### Add tags to the domain
+
+By setting tags for the domain you can more easily filter and export the terraform state to external tools:
+
+```hcl
+tags {
```
MalloZup: It would improve the Ux if we add an example to our example dir with a domain use case, create a dir https://github.com/dmacvicar/terraform-provider-libvirt/tree/master/examples

libvirt/domain.go
lhw: It could make more sense to maybe call it metadata instead and add to it later maybe? Right now I only added tags to the metadata but in the future other options that need to be saved could also be added to the metadata instance xml.

website/docs/r/domain.html.markdown
lhw: I think it would make more sense to just expand upon the coreos example with extra tags. Its not such a wide reaching feature that it would require an extra example in my opinion.

i.e. expand the example to add 1-2 master tags and the rest worker or something to show how it can be used.

lhw:
> I have a question regarding the Domain Metadata element: since when, which version is this used in Libvirt? we have around lot of libvirt users which use like 1.** version, so we need to take care of Backwards compatibility or at least maybe state that this feature is only for XX version. Thx for your patience

> metadata
    The metadata node can be used by applications to store custom metadata in the form of XML nodes/trees. Applications must use custom namespaces on their XML nodes/trees, with only one top-level element per namespace (if the application needs structure, they should have sub-elements to their namespace element). Since **0.9.10**

Should be fine.

website/docs/r/domain.html.markdown
lhw: Amended it to the commit with the CoreOS example

libvirt/domain.go
lhw: I added a commit with my suggestion.

website/docs/r/domain.html.markdown
MalloZup: thx!

libvirt/domain_metadata.go
```diff
@@ -0,0 +1,73 @@
...
+		})
+	}
+	var bytesOut []byte
+	if bytesOut, err = xml.Marshal(metadata); err != nil {
+		return "", fmt.Errorf("Failed to marshal metadata XML: %s", err)
+	}
+	return string(bytesOut), nil
+}
+
+func setMetadata(d *schema.ResourceData, domain *libvirtxml.Domain) (err error) {
```
MalloZup: I think we should name this `setMetadataFromXML`, because only `setMetadata` is from an extnernal reviw kind the same of `domain.setMetadata` function.

Also `getMetadataFromXML` make more sense imho

lhw: Don't really care about the precise function name. I renamed it.

MalloZup: thx!

zeenix:
@lhw Could you please rebase on master?

dmacvicar:
I am not against this, but I am not terribly exited either. I don't see the concrete use case very well, given that metadata in Terraform is not standardized. eg. there is not a standard way for a tool to ask for tags for a generic terraform resource, in a provider independent manner.

* You mention AWS and Azure supporting tags. Can you explain how those are used by applications interacting with Terraform? (eg. not for cost control and reporting).

* You mention Ansible inventories as a use-case. I am myself the author of Salt inventory support. I based my work on the Ansible inventory support, and tags are not used. Instead, a custom terraform provider is used that connects Terraform resources with Ansible entries in the inventory.
  Can you explain this use case better?

lhw:
In our case we don't intend to use the [ansible inventory provider](https://github.com/nbering/terraform-provider-ansible) I'm guessing you are referring to but just the [ansible inventory script](https://github.com/adammck/terraform-inventory) that directly reads from the terraform state. This one does make use of the tags (only for the instances where its available ofcourse) and directly groups them. Which we use to provision the kubernetes cluster accordingly. You can see an example of how would tag the machines in the first commit of the PR.

I think that answers both of your points. And while there currently is no way to generally access metadata on terraform resources I could imagine that a future version of terraform will provide/require this feature.  And tags would be a good starting point for this provider.

MalloZup:
I agree with  last-comment @dmacvicar in a way that this kind of metadata per se in terraform is not really a super-awesome feature. (personally I had also this feeling of lacking use-case at begin).

Although I support the implementation of this feature and it is really clean from technical part, I also see the major point to be minimalistic and don't add a feature which use-cases aren't clear or providing a benift.

zeenix:
@MalloZup I believe the specific use case for this was presented and defended well by @lhw. There is a concrete implementation and I can think of good uses for this feature so I'm sold. However, i think it'd be nice to have the rationale/usecase for adding this into the commit message(s) themselves, although that's not a show-stopper and you can merge this already.

zeenix:
CI failed though so @lhw has to at least update the tests. :)

MalloZup:
adding `need_info` for retriggering and tests_failures

examples/v0.11/coreos/main.tf
```diff
@@ -70,7 +70,6 @@ resource "libvirt_domain" "coreos-machine" {
     # Requires qemu-agent container if network is not native to libvirt
     wait_for_lease = true
   }
-
```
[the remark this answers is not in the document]
lhw: Just means that my example wasn't terraform formatted or the format has changed.

examples/v0.11/coreos/main.tf
```diff
@@ -79,7 +78,7 @@ resource "libvirt_domain" "coreos-machine" {
   #  readonly = true
   #}
   tags {
-    OS = "CoreOS"
+    OS       = "CoreOS"
```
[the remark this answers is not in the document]
lhw: Just means that my example wasn't terraform formatted or the format has changed.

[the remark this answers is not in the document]
zeenix: Yeah but it's a separate logical change so should go in it's own commit (along with any other related fixes).

zeenix:
@lhw seems your patches need some fixing. Check the [CI failure log](https://travis-ci.org/dmacvicar/terraform-provider-libvirt/builds/557224277?utm_source=github_status&utm_medium=notification).

zeenix:
@lhw Thanks for your quick actions. Much appreciated!

dmacvicar:
> I think that answers both of your points. And while there currently is no way to generally access metadata on terraform resources I could imagine that a future version of terraform will provide/require this feature. And tags would be a good starting point for this provider.

After reading the code of the ansible dynamic inventory provider, I am in favour of merging this if this is used.

Once a PR to that project is approved, implementing what is missing in the [inventory code that uses the tags](https://github.com/adammck/terraform-inventory/blob/3a1f43306133e8fc981801ea3e0aa31a537edbba/resource.go#L98), we can then merge the PRs in both projects.

I don't like the idea to merge the code *just in case* it is useful for a project in the future.

lhw:
> Once a PR to that project is approved, implementing what is missing in the [inventory code that uses the tags](https://github.com/adammck/terraform-inventory/blob/3a1f43306133e8fc981801ea3e0aa31a537edbba/resource.go#L98), we can then merge the PRs in both projects.

Again, a different project. And that project does not yet support terraform 0.12 and might be on a back burner for the author anyhow.

> I don't like the idea to merge the code _just in case_ it is useful for a project in the future.

It is in use in our case and the *just in case* part was just for future extensibility not the already introduced feature set.

dmacvicar:
> Again, a different project. And that project does not yet support terraform 0.12 and might be on a back burner for the author anyhow.
> 
> > I don't like the idea to merge the code _just in case_ it is useful for a project in the future.
> 
> It is in use in our case and the _just in case_ part was just for future extensibility not the already introduced feature set.

Sorry. I can't follow. Can you point me to the code that uses our would use this tag support if merged?

lhw:
We are currently not using the terraform-libvirt-provider anymore and I don't have the time to work on this anymore.

PR #593, opened 2019-05-09
domain: telnet console support

enool:
Add a new console connection type - telnet over tcp. This feature is
useful when LAN access to the domain is required but the guest OS is
not provisioned with networking.

Treat all other console types as they were pty based. This is needed
to avoid regression with unknown types since no value bounds check
was in place.

As the telnet default port is a system port, use cisco-sccp service
in the tests. This service name is IANA assigned, is not likely to
be seen on a Linux and resolves on a wide set of distros.

### DIFF
--- a/examples/ubuntu/ubuntu-example.tf
+++ b/examples/ubuntu/ubuntu-example.tf
@@ -51,7 +51,7 @@ resource "libvirt_domain" "domain-ubuntu" {
   }
 
   console {
-    type        = "pty"
+    type        = "tcp"
     target_type = "virtio"
     target_port = "1"
   }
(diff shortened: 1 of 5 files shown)


MalloZup:
@enool  thx for your PR. We will review it soonish

enool:
Just realized that type on console block is required. I'll simplify the code further.

enool:
skytelnet does not resolve on debian based distros. This might be a problem if the acceptance tests are run on them.

I'll change the service to something that works on wider set of distros.

dmacvicar:
Sorry @enool I have this on my queue. I am not familiar with the console configuration, so I need to make up some time to understand well this PR.

enool:
Fair point, I tried to be consistent with the graphics block schema but being consistent with libvirt might be more useful.

Should it be source_host/source_service then to be in line with source_path?

Issue #585, opened 2019-04-10
Enhancement: Support for network storage volumes


lhw:
One common use-case for libvirt_volumes would be using network devices. There is a whole list of different network storage options [available in libvirt](https://libvirt.org/formatdomain.html#elementsDisks).
One example of network devices as backing store for the VM disks would be Ceph RBDs or iSCSI. Which is as far as I can see right now not currently not natively supported. 

There currently is a way to do it with the included XSLT parameter for the volume which I assume was intended for this purpose but that is still a crude way to do it in the end. But I get it as due to the slew of different options available this is sadly not an easy one to implement.

MalloZup:
thx . yes as you stated the `xslt` feature should cover the missing GAP between this provider and the libvirt API. ( https://github.com/dmacvicar/terraform-provider-libvirt#introduction--goals)

lhw:
Doing xslt again after so many years was entertaining. I'll probably make an update to the existing xslt example. But for now here is an example with rbd. I'll look into how this could be implemented in the future. Still have other things to work on currently.

```terraform
data "template_file" "cephdisk" {
  template = "${file("${path.module}/templates/ceph_rbd.xml")}"
  vars {
    diskname = "${format(var.hostname_format, count.index)}.disk"
    ceph_user = "${var.libvirt_ceph_user}"
    ceph_uuid = "${var.libvirt_ceph_uuid}"
    ceph_pool = "${var.libvirt_ceph_pool}"
    ceph_mons = "${join("\n", formatlist("<host name='%s' port='6789' />", var.ceph_mons))}"
  }

  count = "${var.hosts}"
}

resource "libvirt_domain" "coreos-machine" {
  [...]
  xml {
    xslt = "${data.template_file.cephdisk.rendered}"
  }
}
```
```xml
<?xml version="1.0" ?>
<!-- templates/ceph_rbd.xml -->
<xsl:stylesheet version="1.0"
                xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
  <xsl:output omit-xml-declaration="yes" indent="yes"/>
  <xsl:template match="node()|@*">
     <xsl:copy>
       <xsl:apply-templates select="node()|@*"/>
     </xsl:copy>
  </xsl:template>

  <xsl:template match="/domain/devices/disk[source/@volume='${diskname}']">
    <disk type='network' device='disk'>
      <xsl:copy-of select="driver" />
      <auth username='${ceph_user}'>
[...]

PR #604, opened 2019-06-03
[Feature] Domain disks configuration for Ceph network disks

lhw:
## Reason
As mentioned in #585 support for some of the common network drives would be useful in the provider. Also as mentioned in the issue it would be somewhat simpler to separate different implementations by just simply providing them when requested/implemented. I went ahead and got the Ceph version here.

## Changes
When provided with a Ceph configuration in the `libvirt_domain.disks` resource the provider will assume that the linked volume is provided by a network storage. If the Ceph config is not provided no further action will be taken.
If the config does exist the required set of XML domain configurations will be added to the VM.

## Breaking Changes
None I can think of but to use the clone functionality e.g. for OS base images, as given in the example, the PR #591 needs to be merged first. Normal empty disks work just fine without.

(Closes: GH-585)

### DIFF
--- a/examples/ceph/ceph-ubuntu-example.tf
+++ b/examples/ceph/ceph-ubuntu-example.tf
@@ -1,0 +1,102 @@
+# instance the provider
+provider "libvirt" {
+  uri = "qemu:///system"
+}
+
+locals {
+  ceph_user = "libvirt"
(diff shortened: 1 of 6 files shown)


zeenix:
@lhw Could you please rebase on master? There is a good chance the checks pass now. :)

lhw:
We are currently not using the terraform-libvirt-provider anymore and I don't have the time to work on this anymore.

PR #607, opened 2019-06-06
Use only randomChars in testAcc helpers

MalloZup:
- breaking change from Terraform v12: it doesn't allow other chars. Numbers and non letter will
   cause failures


MalloZup:
@dmacvicar the TestACC pass locally so I will merge this soonish so we can remove lot of failures we have with tf12

MalloZup:
merging  since it is for tf12 and not design change

PR #625, opened 2019-08-15
Updating build instructions to adhere to the terraform provider standards

### LINKED ISSUE #620: Confusion on pulling specific Version
### Linux distribution

``` Docker golang:alpine```

### Terraform version

```sh
N/A
```

```
Multiple Versions
```

### Setup

I am working on getting the docker examples, #617,  running but hitting some issues with GO. I am not an expert in GO and it might be a bug in my usage rather then a bug with the provider.

### Steps to Reproduce Issue

```
# Building Libvirt Plugin Docker
FROM golang:alpine

ARG VERSION
ENV VERSION=$VERSION

# Install Needed Packages
RUN apk update \
    && apk upgrade \
    && apk add --no-cache git make pkgconfig libvirt gcc libvirt-dev libc-dev

# Pull Down Needed Projects
RUN go get -u golang.org/x/lint/golint && env GO111MODULE=on go get github.com/dmacvicar/terraform-provider-libvirt@$VERSION

# Set Workdir for bin build
WORKDIR $GOPATH/src/github.com/dmacvicar/terraform-provider-libvirt

# Build
RUN make build
```

In this case version can be set to any release version you guys produced. In my testing I been trying to install `v0.5.2` as its the last version before terraform `0.12`. When setting the version to `v0.5.2`, the module seems not to download. Usually I can find the module in `/go/src/github.com/......`. But when I run with `v0.5.2` nothing seems to download. 

I passed in `env GO111MODULE=on` to support the versioning of `go get`. When I remove `env GO111MODULE=on` and the `@v0.5.2`. the module seems to install find, but master is built for `0.12` of terraform. I wanted to be able to select different versions of this provider. 

Maybe I am mixing things up with GO, if so please correct me. Like I stated I am fairly new to GO and the whole module setup. 

I am not sure if this would work, but could  I cd into the providers folder and do a get checkout `v0.5.2`? I am leaning towards no, since the modules this module relies on might be different versions between your releases.

james-crowley:
Fixes #620 

I updated the README to adhere to the "standards" set by the official Terraform provider READMEs. Theses changes will add more clarity when a user wants to build a specific version. Additionally, these build instructions are how I will be completing the Docker examples.

A couple of questions for @MalloZup, is the `cgo` requirement still a thing? I was able to build the plugin without setting `export CGO_ENABLED="1"`. If so, I think we should remove it from the `Building from source` section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,13 @@ Follow the instructions for your distribution:
 
 ## Building from source
 
-Before building, you will need the following
+### Requirements
 
-* libvirt 1.2.14 or newer development headers
-* latest [golang](https://golang.org/dl/) version
-* `cgo` is required by the [libvirt-go](https://github.com/libvirt/libvirt-go) package. `export CGO_ENABLED="1"`
+-	[Terraform](https://www.terraform.io/downloads.html) 0.11.x
+-	[Go](https://golang.org/doc/install) (to build the provider plugin)
+-   [libvirt](https://libvirt.org/downloads.html) 1.2.14 or newer development headers
+-   `cgo` is required by the [libvirt-go](https://github.com/libvirt/libvirt-go) package. `export CGO_ENABLED="1"`
+
 
 This project uses [go modules](https://github.com/golang/go/wiki/Modules) to vendor all its
 dependencies.
@@ -74,8 +76,18 @@ takes advantage of features available only inside of the latest stable release.
 
 You need also need libvirt-dev(el) package installed.
 
+###Building The Provider
+
+Clone repository to: `$GOPATH/src/github.com/dmacvicar/terraform-provider-libvirt`
+
 ```console
-go get github.com/dmacvicar/terraform-provider-libvirt
+mkdir -p $GOPATH/src/github.com/dmacvicar; cd $GOPATH/src/github.com/dmacvicar
+git clone git@github.com:dmacvicar/terraform-provider-libvirt
+```
+
+Enter the provider directory and build the provider
+
+```console
 cd $GOPATH/src/github.com/dmacvicar/terraform-provider-libvirt
 make install
 ```
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Clone repository to: `$GOPATH/src/github.com/dmacvicar/terraform-provider-libvir
 
 ```console
 mkdir -p $GOPATH/src/github.com/dmacvicar; cd $GOPATH/src/github.com/dmacvicar
-git clone git@github.com:dmacvicar/terraform-provider-libvirt
+git clone https://github.com/dmacvicar/terraform-provider-libvirt.git
 ```
 
 Enter the provider directory and build the provider
--- a/README.md
+++ b/README.md
@@ -60,10 +60,10 @@ Follow the instructions for your distribution:
 
 ### Requirements
 
--	[Terraform](https://www.terraform.io/downloads.html) 0.11.x
+-	[Terraform](https://www.terraform.io/downloads.html)
 -	[Go](https://golang.org/doc/install) (to build the provider plugin)
--   [libvirt](https://libvirt.org/downloads.html) 1.2.14 or newer development headers
--   `cgo` is required by the [libvirt-go](https://github.com/libvirt/libvirt-go) package. `export CGO_ENABLED="1"`
+-	[libvirt](https://libvirt.org/downloads.html) 1.2.14 or newer development headers
+-	`cgo` is required by the [libvirt-go](https://github.com/libvirt/libvirt-go) package. `export CGO_ENABLED="1"`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,8 +76,18 @@ takes advantage of features available only inside of the latest stable release.
 
 You need also need libvirt-dev(el) package installed.
 
+###Building The Provider
```
MalloZup: #### Building The Provider

README.md
```diff
@@ -74,8 +76,18 @@ takes advantage of features available only inside of the latest stable release.
...
+
+Clone repository to: `$GOPATH/src/github.com/dmacvicar/terraform-provider-libvirt`
+
+```console
```
MalloZup: I'm not sure of this doc.

The `go get` way is the standard way of golang, see for example

https://github.com/kubernetes/kubernetes#you-have-a-working-go-environment

README.md
```diff
@@ -58,11 +58,13 @@ Follow the instructions for your distribution:
 
 ## Building from source
 
-Before building, you will need the following
+### Requirements
+
+-	[Terraform](https://www.terraform.io/downloads.html) 0.11.x
```
MalloZup: we have `0.12` in master so the 0.11 is wrong. I wouldn't add any version here since an user could also decide to use terraform 11 if he pick up the older releases etc.

README.md
MalloZup: we can have a `git clone` approach but a Git clone approach would be needed in case we are **outside** gopath, since modules golang where implemented for that reason.

But imho if we decided to document the `module` approach with git clone, we should add it a second method  as the `gopath` one, which was historically what  golang offered first and still offer.

So imho I don't know if it is worth the effort because it complicated a little the doc providing 1, 2 approaches but if you feel to do it  ok. Imho we should just state to use `go get` as was before. What was your problem that led you to do so? In theory go get create automatically directory without need to create and git clone

README.md
```diff
@@ -58,11 +58,13 @@ Follow the instructions for your distribution:
...
+-	[Go](https://golang.org/doc/install) (to build the provider plugin)
+-   [libvirt](https://libvirt.org/downloads.html) 1.2.14 or newer development headers
```
MalloZup: space identattion is somehow weird here or is the Github interface :grin?

james-crowley:
@MalloZup I actually ran into issues with `cgo` on alpine. I needed to install `libc-dev` on alpine which in turned installed `musl-dev`. But I still did not need to set the environment var of `export CGO_ENABLED="1"`. 

Let me fix up some of the other issues you addressed up above.

README.md
james-crowley: The problem comes when I want to build a specific version of this plugin. If you uses `go get` it always grabs the master. Which as of right now is build around using Terraform `0.12`, which my team is not using just yet. 

With `git clone`, you can pull down the repo. Then use a `git checkout` to pull down the correct tag version, aka `v0.5.2`. 

While you can do a similar thing with `go get` then a `git checkout` the go dependencies are not correct.

The only other way I know to do it is to use GO's new syntax, `go get <package>@<version>`. This requires `GO111MODULE=ON` and its not 100% stable in my experience.  

Hence, why I think the `git clone` is the best approach. Plus its what the official modules suggest, https://github.com/terraform-providers/terraform-provider-random#building-the-provider.

README.md
james-crowley: Got it removing it!

README.md
```diff
@@ -58,11 +58,13 @@ Follow the instructions for your distribution:
 
 ## Building from source
 
-Before building, you will need the following
+### Requirements
+
+-	[Terraform](https://www.terraform.io/downloads.html)
```
james-crowley: @MalloZup Should I add a line for `golint`, `go get -u golang.org/x/lint/golint`? It is not install when downloading the go dependencies, but it is required to build the project.

MalloZup: I think `golint` isn't really a requirement. 

In theory we should move it away from the build target to make it more granular. https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/Makefile#L11

README.md
MalloZup: I'm ok to add it

README.md
MalloZup: we can do it other prs if needed but  I think should be ok without

MalloZup:
@james-crowley  I will let this PR a little pending to let @dmacvicar  the time to review and respect his own time and privacy and if he is in vacation or has other things todo. 

I will let this pr like 1/2 weeks open waiting for review, if he will not answer I will assume everything is ok. So just for your info.

james-crowley:
@MalloZup It is all good. No rush. I'll continue to work on the Docker examples and submit a PR for that soon.

Issue #626, opened 2019-08-17
flatcar linux: fw_cfg has been updated


tormath1:
### Linux distribution

```
arch linux
```

### Terraform version

```sh
$ terraform -v
Terraform v0.12.6
```

### Provider and libvirt versions

```sh
$ terraform-provider-libvirt -version
5c01d18431ea2b9df4870990be6a1736ec6f322a
Compiled against library: libvirt 5.5.0
Using library: libvirt 5.5.0
Running hypervisor: QEMU 4.0.0
Running against daemon: 5.5.0
```
___
## Description of Issue/Question

Hi ! I am using this plugin in order to deploy flatcar linux cluster on libvirt. FYI flatcar linux is a fork of coreOS container linux (https://www.flatcar-linux.org/).

Back in the days, the ignition config file was given to the domain through fw_cfg, but the key value has changed from `opt/com.coreos/config` to `opt/com.flatcar-linux/config`. (https://github.com/flatcar-linux/ignition/issues/2)

It could be nice to specify the value of the key in terraform config file ? 
https://github.com/dmacvicar/terraform-provider-libvirt/blob/5c01d18431ea2b9df4870990be6a1736ec6f322a/libvirt/domain.go#L240-L249

MalloZup:
@tormath1  thx. I think it should be not really difficult to convert it to a variable. thx for issue

tormath1:
Ok, I am on it. :rocket:

MalloZup:
@tormath1 considering this closed by your pr. Feel free to squash commits

PR #627, opened 2019-08-17
feat(domain): add fw_cfg_name option

tormath1:
Hi, 

This PR implements #626 (and this is my first PR on a Terraform plugin :rocket: ). I already squashed my commits. 

Goal of this PR is to allow the user to give the firmware configuration path (`opt/something/something`) when using an ignition file.

For the unit test, I simply copied this one:
https://github.com/dmacvicar/terraform-provider-libvirt/blob/5c01d18431ea2b9df4870990be6a1736ec6f322a/libvirt/resource_libvirt_domain_test.go#L714 and I updated the `testAccCheckIgnitionXML` because `opt/com.coreos/config` was hardcoded. 

Thanks !

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -236,16 +236,20 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
 		if err != nil {
 			return err
 		}
+		// `fw_cfg_name` stands for firmware config is defined by a key and a value
+		// credits for this cryptic name: https://github.com/qemu/qemu/commit/81b2b81062612ebeac4cd5333a3b15c7d79a5a3d
+		if fwCfg, ok := d.GetOk("fw_cfg_name"); ok {
 
-		domainDef.QEMUCommandline = &libvirtxml.DomainQEMUCommandline{
-			Args: []libvirtxml.DomainQEMUCommandlineArg{
-				{
-					Value: "-fw_cfg",
+			domainDef.QEMUCommandline = &libvirtxml.DomainQEMUCommandline{
+				Args: []libvirtxml.DomainQEMUCommandlineArg{
+					{
+						Value: "-fw_cfg",
+					},
+					{
+						Value: fmt.Sprintf("name=%s,file=%s", fwCfg, ignitionKey),
+					},
 				},
-				{
-					Value: fmt.Sprintf("name=opt/com.coreos/config,file=%s", ignitionKey),
-				},
-			},
+			}
 		}
 	}
 
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -758,7 +758,7 @@ func TestAccLibvirtDomain_IgnitionObject(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckLibvirtDomainExists("libvirt_domain."+randomDomainName, &domain),
 					testAccCheckIgnitionVolumeExists("libvirt_ignition."+randomIgnitionName, &volume),
-					testAccCheckIgnitionXML(&domain, &volume),
+					testAccCheckIgnitionXML(&domain, &volume, "opt/com.coreos/config"),
 				),
 			},
 		},
@@ -996,7 +996,7 @@ func testAccCheckLibvirtDomainExists(name string, domain *libvirt.Domain) resour
 	}
 }
 
-func testAccCheckIgnitionXML(domain *libvirt.Domain, volume *libvirt.StorageVol) resource.TestCheckFunc {
+func testAccCheckIgnitionXML(domain *libvirt.Domain, volume *libvirt.StorageVol, fwCfg string) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 
 		domainDef, err := getXMLDomainDefFromLibvirt(domain)
@@ -1008,7 +1008,7 @@ func testAccCheckIgnitionXML(domain *libvirt.Domain, volume *libvirt.StorageVol)
 		if err != nil {
 			return err
 		}
-		ignStr := fmt.Sprintf("name=opt/com.coreos/config,file=%s", ignitionKey)
+		ignStr := fmt.Sprintf("name=%s,file=%s", fwCfg, ignitionKey)
 
 		cmdLine := domainDef.QEMUCommandline.Args
 		for i, cmd := range cmdLine {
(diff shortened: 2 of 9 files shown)


MalloZup:
@tormath1 thx for pr I will review during the week

libvirt/resource_libvirt_domain_test.go
```diff
@@ -1767,3 +1767,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
 	}
 	return nil
 }
+func TestAccLibvirtDomain_FwCfgName(t *testing.T) {
```
MalloZup: can we have a comment here after this line specifying the `why` of this test exist? I would just state it in high manner that we want to support other things then coreos and for that we need to test that fwcfname (what a cryptic name ! ) . So i would like some comment which will help us to maintain in future

MalloZup: I just want at the end to have a highlevel comment, to say what we are testing and why.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -1767,3 +1767,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
...
+	var domain libvirt.Domain
+	var volume libvirt.StorageVol
+	randomDomainName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	randomFwCfgName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	randomPoolName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	randomPoolPath := "/tmp/terraform-provider-libvirt-pool-" + randomPoolName
+	randomIgnitionName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	var config = fmt.Sprintf(`
+	data "ignition_systemd_unit" "acceptance-test-systemd" {
+		name    = "example.service"
```
MalloZup: randomize this name with a number since you copied pasted.

`example.service02` or whatever

libvirt/resource_libvirt_domain_test.go
```diff
@@ -1767,3 +1767,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
...
+		content = "[Service]\nType=oneshot\nExecStart=/usr/bin/echo Hello World\n\n[Install]\nWantedBy=multi-user.target"
+	}
+
+	data "ignition_config" "acceptance-test-config" {
```
MalloZup: change name `acceptance-test-config` to  a unique name, `acceptance-test-config-fcqf` or make it more appropriate.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -1767,3 +1767,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
...
+		systemd = [
+		"${data.ignition_systemd_unit.acceptance-test-systemd.id}",
+		]
+	}
+
+    resource "libvirt_pool" "%s" {
```
MalloZup: for this test, I don't think we need the `libvirt_pool` Just remove this resource and adapt it accordingly.

Having the libvirt_pool here creation would add a pointer of failure of the test.

libvirt/resource_libvirt_domain_test.go
```diff
@@ -1767,3 +1767,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
...
+        name = "%s"
+        type = "dir"
+        path = "%s"
+    }
+
+	resource "libvirt_ignition" "%s" {
+		name    = "ignition"
+		content = "${data.ignition_config.acceptance-test-config.rendered}"
```
MalloZup: adapt accordingly

libvirt/resource_libvirt_domain_test.go
```diff
@@ -1767,3 +1767,58 @@ func testAccCheckLibvirtDomainDestroy(s *terraform.State) error {
...
+        pool    = "${libvirt_pool.%s.name}"
+	}
+
+	resource "libvirt_domain" "%s" {
+		name            = "terraform-test-domain"
```
MalloZup: ditto

MalloZup: change this name with a unique name. for moment this is coped pasted. we can randomize it.

Good would be that your randomize it also in coreos but is not releated to this

tormath1:
Hey @MalloZup, 

Thanks for your quick review ! 
Yes, I tested with the files provided in the example: 
```shell
$ cat main.tf | grep fw 
  fw_cfg_name = "opt/org.flatcar-linux/config"
$ terraform apply
$ virsh dumpxml --domain node-01 | grep qemu:arg
    <qemu:arg value='-fw_cfg'/>
    <qemu:arg value='name=opt/org.flatcar-linux/config,file=/var/lib/libvirt/images/container-linux/node-01-ignition'/>
```
Now, we remove the `fw_cfg_name`

```shell
$ cat main.tf| grep fw
$ virsh dumpxml --domain node-01 | grep qemu:arg
    <qemu:arg value='-fw_cfg'/>
    <qemu:arg value='name=opt/com.coreos/config,file=/var/lib/libvirt/images/container-linux/node-01-ignition'/>
```

We have the default value to `opt/com.coreos/config`. :tada: 

I pushed my commit unsquashed to make review easier. I will squash them before merging. :)

libvirt/resource_libvirt_domain_test.go
tormath1: If we remove the `livbirt_pool` libvirt will looking for the `default` pool which leads to CI failure. :thinking:

MalloZup: ok somehow I think is related to the fact we don't have a default. I think in that case we can restore the default pool

libvirt/domain.go
```diff
@@ -237,15 +237,18 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
...
 		}
 
-		domainDef.QEMUCommandline = &libvirtxml.DomainQEMUCommandline{
-			Args: []libvirtxml.DomainQEMUCommandlineArg{
-				{
-					Value: "-fw_cfg",
-				},
-				{
-					Value: fmt.Sprintf("name=opt/com.coreos/config,file=%s", ignitionKey),
+		if fwCfg, ok := d.GetOk("fw_cfg_name"); ok {
```
MalloZup: we can add also here a comment to the cryptic name.
``fw_cfg` stands for firmware config is defined by a key and a value.`

Another thing, but this will be out of scope: consider that your create the disk and so on with the example.

Now we need to think about the `update` operation. the update operation is when you have a terraform infra already deployed (`terraform apply`, and we change the tfstate file, e.g one change the name.

To handle this operation is kind tricky for many reasons, but I would like that we could think about this.. In some resources this update can't be implemented and we have already in codebase some incongruence with this. However we might think.

tormath1: IMO it was handled by the Terraform Option 
```go
ForceNew: true
```
If `fw_cfg_name` is updated, actions after will be created from scratch no ?

libvirt/resource_libvirt_domain_test.go
tormath1: Ok, I revert the commit so.

libvirt/domain.go
MalloZup: ah right anyways it is more tricky in other places but for this looks good resolving thx

tormath1:
Alright. I will rebase / squash if needed. Thanks for your review!

tormath1:
Hi @MalloZup , 

Rebased and squashed :heavy_check_mark:

MalloZup:
nice thx! If you find a way for fixing travis would be awesome! Iknow that is not your pr causing the failure, but someting with travis.Otherwise i will merge it tomorrow

tormath1:
Yeah I've seen that earlier. Apache git repo seems down?

MalloZup:
lgtm thx!

Issue #624, opened 2019-08-13
Add system z (s390x) support for coreos ignition 


jichenjc:
### Linux distribution
N/A

### Terraform version
N/A (master)

### Provider and libvirt versions
N/A

## Description of Issue/Question
currently the provider of libvirt is using fw_cfg anyway while from coreos perspective, which means in order to create and add ingition files , it uses fw_cfg unconditionally
https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/domain.go#L240

however coreos it supports guestfish and fw_cfg
https://github.com/coreos/coreos-assembler/blob/master/src/cmd-run#L240

so the proposal here is  to have a new method to pass ignition to coreos by 
referring to following implementation
https://github.com/coreos/coreos-assembler/blob/master/src/libguestfish.sh
which means, you can either select fw_cfg or files based on different arch (power and s390x)
### Setup

(Please provide the full _main.tf_ file for reproducing the issue (Be sure to remove sensitive information)

### Steps to Reproduce Issue

(Include debug logs if possible and relevant).

___
## Additional information:

Do you have SELinux or Apparmor/Firewall enabled? Some special configuration?
Have you tried to reproduce the issue without them enabled?

No

jichenjc:
@bjhuangr FYI

MalloZup:
@jichenjc  hi thx! can you create a better description and removing/adapting the template? thx

jichenjc:
@MalloZup ,thanks, I updated with some info I think helpful just now, let me know you need further info ,thanks

PR #629, opened 2019-08-19
Add coreos ignition through guestfs on s390 platform

jichenjc:
Fixes: #624 

Add support for s390x on coreos ignition

now x86 platform use fw_cfg to process coreos ingition
but fw_cfg is not usable in s390 platform.

so we need use libguestfs to update the coreos image by
injecting the ignition into it.

we have some comments like following to try but that's not for coreos
https://gitlab.gnome.org/GNOME/gnome-boxes/blob/master/src/unattended-file.vala

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -1,11 +1,16 @@
 package libvirt
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"log"
 	"net"
 	"net/url"
+	"os"
+	"os/exec"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
@@ -230,7 +235,143 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
 	return disk, nil
 }
 
-func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) error {
+func getFirstDiskPath(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *libvirt.Connect) (string, error) {
+	prefix := fmt.Sprintf("disk.%d", 0)
+
+	if volumeKey, ok := d.GetOk(prefix + ".volume_id"); ok {
+		diskVolume, err := virConn.LookupStorageVolByKey(volumeKey.(string))
+		if err != nil {
+			return "", fmt.Errorf("Error lookup storage disk volume: %s", err)
+		}
+		diskVolumeFile, err := diskVolume.GetPath()
+		if err != nil {
+			return "", fmt.Errorf("Error get path of disk volume: %s", err)
+		}
+		return diskVolumeFile, nil
+	}
+	return "", fmt.Errorf("failed to find disk 0")
+}
+
+func getStderr(stderr io.ReadCloser) string {
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(stderr)
+	return buf.String()
+}
+
+func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *libvirt.Connect) error {
+	if ignition, ok := d.GetOk("coreos_ignition"); ok {
+
+		// Get the root disk (first disk) volume location
+		rootPath, err := getFirstDiskPath(d, domainDef, virConn)
+		if err != nil {
+			return err
+		}
+
+		// Get the ingition file to be uploaded
+		ignitionKey, err := getIgnitionVolumeKeyFromTerraformID(ignition.(string))
+		if err != nil {
+			return err
+		}
+		diskVolume, err := virConn.LookupStorageVolByKey(ignitionKey)
+		if err != nil {
+			return err
+		}
+		fileToUpload, err := diskVolume.GetPath()
+		if err != nil {
+			return err
+		}
+		err = guestfishExecution(rootPath, fileToUpload)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+var execCommand = exec.Command
+
+func guestfishExecution(rootPath string, fileToUpload string) error {
+	// This is the file in coreos image going to be copied from local
+	fileRemoteLocation := "/ignition/config.ign"
+
+	// eval  $(guestfish --listen -a $1)
+	cmd := execCommand("guestfish", "--listen", "-a", rootPath)
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("Get stdout failed: %v", err)
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("Get stderr failed: %v", err)
+	}
+
+	err = cmd.Start()
+	if err != nil {
+		return fmt.Errorf("Start command failed: %v", err)
+	}
+
+	// if the command start successfully, the output will be following format
+	// GUESTFISH_PID=xxxx; export GUESTFISH_PID, so we need get the PID and set by using os.Setenv
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(stdout)
+	guestfishPidLabel := strings.Split(buf.String(), ";")
+	guestfishPidValue := strings.Split(guestfishPidLabel[0], "=")
+	os.Setenv(guestfishPidValue[0], guestfishPidValue[1])
+
+	// guestfish --remote -- run
+	cmd = execCommand("guestfish", "--remote", "--", "run")
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("Launch failed: %v, %s", err, getStderr(stderr))
+	}
+
+	// guestfish --remote -- exit
+	defer func() {
+		cmd := execCommand("guestfish", "--remote", "--", "exit")
+		cmd.Run()
+	}()
+
+	// guestfish --remote -- findfs-label
+	cmd = execCommand("guestfish", "--remote", "--", "findfs-label", "boot")
+	output, err := cmd.Output()
+
+	if err != nil {
+		return fmt.Errorf("Failed to find fs label: %v", err)
+	}
+
+	bootDisk := strings.TrimSpace(string(output))
+	if len(bootDisk) == 0 {
+		return fmt.Errorf("failed to get the boot filesystem")
+	}
+
+	// guestfish --remote -- mount $2 /
+	cmd = execCommand("guestfish", "--remote", "--", "mount", bootDisk, "/")
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("Mount failed: %v, %s", err, getStderr(stderr))
+	}
+
+	// This is the real command that upload the file from current location (the local file system)
+	// to remote file location (the coreos file system)
+	// guestfish --remote -- upload $4 $3/$4
+	cmd = execCommand("guestfish", "--remote", "--", "upload", fileToUpload, fileRemoteLocation)
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("Upload failed: %v, %s", err, getStderr(stderr))
+	}
+
+	// guestfish --remote -- umount-all
+	cmd = execCommand("guestfish", "--remote", "--", "umount-all")
+	err = cmd.Run()
+	if err != nil {
+		return fmt.Errorf("Unmount failed: %v, %s", err, getStderr(stderr))
+	}
+
+	return nil
+}
+
+// Use fw_cfg device to update coresos
+func fwcfgUpdateCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) error {
 	if ignition, ok := d.GetOk("coreos_ignition"); ok {
 		ignitionKey, err := getIgnitionVolumeKeyFromTerraformID(ignition.(string))
 		if err != nil {
@@ -250,6 +391,15 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
 	}
 
 	return nil
+}
+
+func updateCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *libvirt.Connect) error {
+	if runtime.GOARCH == "s390x" || runtime.GOARCH == "s390" {
+		// There is no support for fw_cfg in s390x, so instead of doing fw_cfg
+		// we have to inline update the coreos ignition file
+		return inlineUpdateCoreOsIgnition(d, domainDef, virConn)
+	}
+	return fwcfgUpdateCoreOSIgnition(d, domainDef)
 }
 
 func setVideo(d *schema.ResourceData, domainDef *libvirtxml.Domain) error {
--- a/libvirt/guestfish_test.go
+++ b/libvirt/guestfish_test.go
@@ -1,0 +1,248 @@
+package libvirt
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"regexp"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+type GuestfsStatus int
+
+const (
+	Stopped GuestfsStatus = iota
+	Added
+	Started
+	Mounted
+)
+
+var currentGuestfishStatus = Stopped
+
+func (s GuestfsStatus) String() string {
+	return []string{"Stopped", "Added", "Started", "Mounted"}[s]
+}
+
+/*
+var supportedCommands = [][]string{
+	{
+		"guestfish", "--listen", "-a", "*",
+	},
+	{
+		"guestfish", "--remote", "--", "run",
+	},
+	{
+		"guestfish", "--remote", "--", "findfs-label", "boot",
+	},
+	{
+		"guestfish", "--remote", "--", "mount", "*", "/",
+	},
+	{
+		"guestfish", "--remote", "--", "upload", "*.ign", "/ignition/config.ign",
+	},
+	{
+		"guestfish", "--remote", "--", "umount-all",
+	},
+	{
+		"guestfish", "--remote", "--", "exit",
+	},
+} */
+
+var supportedCommandTree = map[string]interface{}{
+	"guestfish": map[string]interface{}{
+		"--listen": map[string]interface{}{
+			"-a": map[string]interface{}{
+				".*": nil,
+			},
+		},
+		"--remote": map[string]interface{}{
+			"--": map[string]interface{}{
+				"run": nil,
+				"findfs-label": map[string]interface{}{
+					"boot": nil,
+				},
+				"mount": map[string]interface{}{
+					".*": map[string]interface{}{
+						"/": nil,
+					},
+				},
+				"upload": map[string]interface{}{
+					".*.ign": map[string]interface{}{
+						"/ignition/config.ign": nil,
+					},
+				},
+				"umount-all": nil,
+				"exit":       nil,
+			},
+		},
+	},
+}
+
+// fakeExecCommand is used to replace the real command in unit test
+func fakeExecCommand(executable string, args ...string) *exec.Cmd {
+	newArgs := []string{"-test.run=TestHelperCommand", "--", executable}
+	newArgs = append(newArgs, args...)
+
+	cmd := exec.Command(os.Args[0], newArgs...)
+	cmd.Env = []string{"USE_HELPER_COMMAND=1"}
+
+	cmdToValidate := []string{executable}
+	cmdToValidate = append(cmdToValidate, args...)
+
+	fmt.Printf("### Executing command: %s\n", strings.Join(cmd.Args, " "))
+
+	if valid := validateCommandSyntax(cmdToValidate...); !valid {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("STDERR=%s", fmt.Sprintf("invalid command: %s", strings.Join(cmdToValidate, " "))))
+		cmd.Env = append(cmd.Env, "EXIT_STATUS=1")
+	} else {
+		if valid, msg := validateCommandSemanticsAndGenerateOutput(cmdToValidate...); valid {
+			cmd.Env = append(cmd.Env, fmt.Sprintf("STDOUT=%s", msg))
+			cmd.Env = append(cmd.Env, "EXIT_STATUS=0")
+		} else {
+			cmd.Env = append(cmd.Env, fmt.Sprintf("STDERR=%s", msg))
+			cmd.Env = append(cmd.Env, "EXIT_STATUS=2")
+		}
+	}
+	return cmd
+}
+
+// validateCommand checks whether the command's syntax is valid or not
+func validateCommandSyntax(args ...string) bool {
+	if len(args) < 4 {
+		return false
+	}
+
+	newArgs := args
+	if newArgs[0] == "sudo" && newArgs[1] == "--preserve-env" {
+		newArgs = newArgs[2:]
+	}
+
+	cmdTree := supportedCommandTree
+	// check each argument one by one
+	for index, arg := range newArgs {
+		if v, ok := cmdTree[arg]; ok {
+			if index+1 == len(newArgs) {
+				return v == nil
+			}
+			cmdTree = v.(map[string]interface{})
+		} else {
+			matched := false
+			for k, v := range cmdTree {
+				if matched, _ = regexp.MatchString(k, arg); matched {
+					if index+1 == len(newArgs) {
+						return v == nil
+					}
+					cmdTree = v.(map[string]interface{})
+					break
+				}
+			}
+			if !matched {
+				return false
+			}
+		}
+	}
+
+	return true
+}
+
+// validateCommandSemantic checks whether the command's semantics is valid or not.
+// if return true, then the retunred string is the specific output message for the command;
+// if return false, then the returned string is an error message.
+func validateCommandSemanticsAndGenerateOutput(args ...string) (bool, string) {
+	newArgs := args
+	if newArgs[0] == "sudo" && newArgs[1] == "--preserve-env" {
+		newArgs = newArgs[2:]
+	}
+
+	errMsg := fmt.Sprintf("Can't execute command '%v' when the guestfish status is %s", strings.Join(args, " "), currentGuestfishStatus)
+
+	if newArgs[1] == "--listen" {
+		if Stopped == currentGuestfishStatus {
+			currentGuestfishStatus = Added
+			return true, "GUESTFISH_PID=4513; export GUESTFISH_PID"
+		}
+		return false, errMsg
+	}
+
+	if newArgs[3] == "run" {
+		if Added == currentGuestfishStatus {
+			currentGuestfishStatus = Started
+			return true, ""
+		}
+		return false, errMsg
+	}
+
+	if newArgs[3] == "findfs-label" {
+		if Started == currentGuestfishStatus {
+			return true, "/dev/sda1"
+		}
+		return false, errMsg
+	}
+
+	if newArgs[3] == "mount" {
+		if Started == currentGuestfishStatus {
+			currentGuestfishStatus = Mounted
+			return true, ""
+		}
+		return false, errMsg
+	}
+
+	if newArgs[3] == "upload" {
+		if Mounted == currentGuestfishStatus {
+			return true, ""
+		}
+		return false, errMsg
+	}
+
+	if newArgs[3] == "umount-all" {
+		if Mounted == currentGuestfishStatus {
+			currentGuestfishStatus = Started
+			return true, ""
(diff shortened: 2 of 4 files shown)


jichenjc:
@MalloZup @zeenix can you help to check this?  Basically, s390x and power arch doesn't support fw_cfg, so instead of create a disk that contains ignition file,  I want to create a script that update *in place* about the coreos disk volume directly by using the guestfish and then put the ignition by using https://github.com/coreos/ignition/blob/master/internal/providers/file/file.go 

looks like the libguestfs need 'generate' the go file here https://github.com/libguestfs/libguestfs , so I think it will be easier to just put some scripts and add some document about how it can be used, can you provide some comments whether this is the right direction? Thanks

MalloZup:
@jichenjc thx for this. Sofar, I must admit that 1) I'm not familiar with the COREOS env. and I don't understand the problematic with mainframe right now, or which problem are we trying to solve that could not be fixed for example here: https://www.terraform.io/docs/providers/ignition/index.html

Regarding the current status of PR, I also have concern of how we will use some bash scripts and adding them to the codebase lifecycle isn't a good idea.

Can you help me understand better the problem without any specific solution to guest_fs?

E.g consider that we want to provide the most generic  and stable solution for all users outside here, so this are my first concern as maintainer.

Beeing that sayd thx for this PR and effort. Imho I would need to invest more time on this PR, would be for me also great if you could provide more details on the issue or here

@zeenix  if you have more details on this you can feel free to share thx! :rocket:

jichenjc:
@MalloZup THe prolbem is described in #624 , coreos uses ignition (which is similar to cloud-init but in even early stage of the boot process), the fw_cfg which is a device is used as the way to pass ignition info to the boot machine in x86, however, the device is not supported by power or s390x, so that means we need find another way to pass those information 

https://github.com/coreos/coreos-assembler/blob/2851f8df0aafd575e70b61a788fe4c6698433527/src/cmd-run#L239 has detail on the method used , it use guestfish to inject files *before* OS start, so the ignition file is updated directly in the boot disk and what I am proposing here is to do the same way

please let me know further question, thanks a lot

jichenjc:
Also, I updated the code to remove the bash scripts and use native command way, please let me know whether it's ok to use those method... thanks @MalloZup

jichenjc:
@MalloZup additional info, I can add something like below to operate the qcow2 image file
however, it might (not 100% sure) cause terraform provider to link guestfish so file and in turn request all other users who don't have guestfish need force to install this ...
```
 package main

 import (
        "fmt"
        "os"
        "libguestfs.org/guestfs"
 )

 func main() {
        if len(os.Args) < 2 {
                panic ("usage: inspect-vm disk.img")
        }
        disk := os.Args[1]

        g, errno := guestfs.Create ()
        if errno != nil {
                panic (fmt.Sprintf ("could not create handle: %s", errno))
        }

          if err := g.Shutdown (); err != nil {
            panic (err)
        }
 }
```

if I remove the /usr/lib64/libguestfs.so in my system, then compile above file will have following error, so If I add those into terraform-provier-libvirt it might cause similar issue which I think we might want to avoid ...
```
/usr/bin/ld: cannot find -lguestfs
collect2: error: ld returned 1 exit status
```

zeenix:
From my side, adding a dep to `libguestfish` is not desireable. It requires OCaml devtools on the build system and that was one of the main reasons we tried our best to avoid this dep in GNOME Boxes, even though it could have been useful. Especially here, it's not at all needed/useful for the most common case: x86.

Maybe we could achieve the injection w/o requiring `libguestfs`? Maybe you could use one of the [methods used by Boxes](https://gitlab.gnome.org/GNOME/gnome-boxes/blob/master/src/unattended-file.vala)?

jichenjc:
I don't have too much knowledge on the boxes you provided, might need more time to dig 
and because the os.execute command will not need the guestfish library ,so is it acceptable tradeoff (which is the current code I uploaded)? @zeenix

libvirt/domain.go
```diff
@@ -252,6 +291,15 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
 	return nil
 }
 
+func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) error {
```
zeenix: I don't quite get it. You're adding another function with the same name as above? Who calls it?

jichenjc: sorry, I uploaded the wrong stuff, I will re-upload this soon, thanks for pointing out

zeenix:
>  because the os.execute command will not need the guestfish library ,so is it acceptable tradeoff

Hmm.. it does require the `guestfish` binary, which also comes with the same build-time deps but since it's used dynamically, we don't depend on it for x86. We need to ensure this code path is only used for non-x86 platforms. At the moment, i find it hard to read your patch and it doesn't even build. Could you please fix it?

jichenjc:
yes, I actually had several paths that I want to know whether it's good approach or not.. so the current code is out dated and I will upload a new one soon, thanks and sorry for inconvenience.

MalloZup:
@zeenix  thx for helping here! I will need to allocate time for looking at this.

@jichenjc  thx for extra info :+1:

jichenjc:
I uploaded the latest change, I didn't finish the end to end test now but most functions I did some local test already

jichenjc:
@MalloZup @zeenix can you help to take a look at updated code? I pasted the test result below.. thanks

```
[root@wxkocp59 coreos]# cat main.tf 


provider "libvirt" {
  uri = "qemu:///system"
}


variable "libvirt_provider" {
  type = "string"
}


variable "hostname_format" {
  type    = "string"
  default = "coreos%02d"
}


resource "libvirt_volume" "coreos-disk" {
  name = "coreos02.qcow2"
  source = "/home/qemu/fedora-coreos-30.20190828.dev.2-qemu.qcow2"
  pool             = "qemu"
  format           = "qcow2"
}


resource "libvirt_ignition" "ignition" {
  name    = "${format(var.hostname_format, count.index + 1)}-ignition"  
  pool    = "qemu"
  content = "/home/qemu/config.ign"
}


resource "libvirt_network" "neti"{
  name = "fcoreosneti"
  mode = "nat"
  bridge = "brf2"
  domain = "domainf"
  addresses = ["10.10.3.0/24"]
}


# Create the virtual machines
resource "libvirt_domain" "coreos-machine" {
  name   = "coreos830"
  vcpu   = "1"
  memory = "2048"


  coreos_ignition = "${element(libvirt_ignition.ignition.*.id, count.index)}"


  disk {
    volume_id = "${libvirt_volume.coreos-disk.id}"
  }


  graphics {
    listen_type = "address"
  }


  console {
    type = "pty"
    target_port = "0"
  }


  network_interface {
    network_name = "default"
    network_id = "${libvirt_network.neti.id}"  
  }
}


output "ipv4" {
  value = "${libvirt_domain.coreos-machine.*.network_interface.0.addresses}"
}


[root@wxkocp59 coreos]# cat /home/qemu/config.ign 
{
  "ignition": {
    "config": {},
    "timeouts": {},
    "version": "3.0.0"
  },
  "networkd": {},
  "passwd": {
    "users": [
      {
        "name": "core",
        "sshAuthorizedKeys": [
          "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCZ9F72vS0dCc9qQPVRu9LAWarP2Oe4vGGGvN1ctuB+0UR40PnI7S8TnshAK/RTgTyxRIY6iNQqTW5bTBU/Ys2FEsHvps+ZzTMUDQtv/5pifngON2UZ/rW2m+cqXtjKP7oKZqXUzZd3K0p7iYm2Gk6D8YwqiSkDazmrxAqv+/gcRolNECaA9UywCqaa744HJLqg2S6FRSH1BVARIHglRhIzWbocqtZKq9YJrMLZlE9h2MP2bN5HoiB5TIJxJrfr8X3wSIPRW7MmJTtQpFfXoCsQIH+in2tiLx98JY1gtw1u03ycTOuUsHGZKketR3HYePDlRxgPDKkJIlmLrdfX9zOf root@wxkocp59"
       ]
      }
    ]
  },
  "storage": {},
  "systemd": {}
}


[root@wxkocp59 coreos]# terraform apply
var.libvirt_provider
  Enter a value: yes




  + create


  + libvirt_domain.coreos-machine
      id:                               <computed>
      arch:                             <computed>
      console.#:                        "1"
      console.0.source_host:            "127.0.0.1"
      console.0.source_service:         "0"
      console.0.target_port:            "0"
      console.0.type:                   "pty"
      coreos_ignition:                  "${element(libvirt_ignition.ignition.*.id, count.index)}"
      disk.#:                           "1"
      disk.0.scsi:                      "false"
      disk.0.volume_id:                 "${libvirt_volume.coreos-disk.id}"
      emulator:                         <computed>
      graphics.#:                       "1"
      graphics.0.autoport:              "true"
      graphics.0.listen_address:        "127.0.0.1"
      graphics.0.listen_type:           "address"
      graphics.0.type:                  "spice"
      machine:                          <computed>
      memory:                           "2048"
      name:                             "coreos830"
      network_interface.#:              "1"
      network_interface.0.addresses.#:  <computed>
      network_interface.0.hostname:     <computed>
      network_interface.0.network_id:   "${libvirt_network.neti.id}"
      network_interface.0.network_name: "default"
      qemu_agent:                       "false"
      running:                          "true"
      vcpu:                             "1"


  + libvirt_ignition.ignition
      id:                               <computed>
      content:                          "/home/qemu/config.ign"
      name:                             "coreos01-ignition"
      pool:                             "qemu"


  + libvirt_network.neti
      id:                               <computed>
      addresses.#:                      "1"
      addresses.0:                      "10.10.3.0/24"
      bridge:                           "brf2"
      domain:                           "domainf"
      mode:                             "nat"
      name:                             "fcoreosneti"


  + libvirt_volume.coreos-disk
      id:                               <computed>
      format:                           "qcow2"
      name:                             "coreos02.qcow2"
      pool:                             "qemu"
      size:                             <computed>
      source:                           "/home/qemu/fedora-coreos-30.20190828.dev.2-qemu.qcow2"




Plan: 4 to add, 0 to change, 0 to destroy.


Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.


  Enter a value: yes


libvirt_ignition.ignition: Creating...
  content: "" => "/home/qemu/config.ign"
  name:    "" => "coreos01-ignition"
  pool:    "" => "qemu"
libvirt_network.neti: Creating...
  addresses.#: "" => "1"
  addresses.0: "" => "10.10.3.0/24"
  bridge:      "" => "brf2"
  domain:      "" => "domainf"
  mode:        "" => "nat"
  name:        "" => "fcoreosneti"
libvirt_volume.coreos-disk: Creating...
  format: "" => "qcow2"
  name:   "" => "coreos02.qcow2"
  pool:   "" => "qemu"
  size:   "" => "<computed>"
  source: "" => "/home/qemu/fedora-coreos-30.20190828.dev.2-qemu.qcow2"
libvirt_network.neti: Creation complete after 6s (ID: b431f913-dc60-48f4-b225-a357f562a717)
libvirt_volume.coreos-disk: Creation complete after 8s (ID: /home/qemu/coreos02.qcow2)
libvirt_ignition.ignition: Still creating... (10s elapsed)
libvirt_ignition.ignition: Creation complete after 12s (ID: /home/qemu/coreos01-ignition;5d68c2e4-ac91-9978-654a-7ca18243c084)
libvirt_domain.coreos-machine: Creating...
  arch:                             "" => "<computed>"
  console.#:                        "" => "1"
  console.0.source_host:            "" => "127.0.0.1"
  console.0.source_service:         "" => "0"
  console.0.target_port:            "" => "0"
  console.0.type:                   "" => "pty"
  coreos_ignition:                  "" => "/home/qemu/coreos01-ignition;5d68c2e4-ac91-9978-654a-7ca18243c084"
  disk.#:                           "" => "1"
  disk.0.scsi:                      "" => "false"
  disk.0.volume_id:                 "" => "/home/qemu/coreos02.qcow2"
  emulator:                         "" => "<computed>"
  graphics.#:                       "" => "1"
  graphics.0.autoport:              "" => "true"
  graphics.0.listen_address:        "" => "127.0.0.1"
  graphics.0.listen_type:           "" => "address"
  graphics.0.type:                  "" => "spice"
  machine:                          "" => "<computed>"
  memory:                           "" => "2048"
  name:                             "" => "coreos830"
  network_interface.#:              "" => "1"
  network_interface.0.addresses.#:  "" => "<computed>"
  network_interface.0.hostname:     "" => "<computed>"
  network_interface.0.network_id:   "" => "b431f913-dc60-48f4-b225-a357f562a717"
  network_interface.0.network_name: "" => "default"
  qemu_agent:                       "" => "false"
  running:                          "" => "true"
  vcpu:                             "" => "1"
libvirt_domain.coreos-machine: Creation complete after 2s (ID: beac738f-1d79-4335-a8c9-7e15ddd5f687)


Apply complete! Resources: 4 added, 0 changed, 0 destroyed.


Outputs:


ipv4 = [
    [
    ]
]


[root@wxkocp59 coreos]# ssh core@192.168.122.164
The authenticity of host '192.168.122.164 (192.168.122.164)' can't be established.
ECDSA key fingerprint is SHA256:4IYuOfUg2LGt7cozfre8EnTa8y0TMyRNxuyy8BhqOZI.
ECDSA key fingerprint is MD5:d4:9b:d1:04:f5:09:b8:2f:bf:01:40:23:52:98:dc:d1.
Are you sure you want to continue connecting (yes/no)? yes
Warning: Permanently added '192.168.122.164' (ECDSA) to the list of known hosts.
Fedora 30.20190828.dev.2 (CoreOS preview)
Tracker: https://github.com/coreos/fedora-coreos-tracker
```

jichenjc:
@MalloZup @zeenix can you help to take another look at the PR? the local test passed 


the error here is following so I think it's not related to my PR.. thanks

```
go: git.apache.org/thrift.git@v0.0.0-20180902110319-2566ecd5d999: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /home/travis/gopath/pkg/mod/cache/vcs/83dba939f95a790e497d565fc4418400145a1a514f955fa052f662d56e920c3e: exit status 128:

	fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection timed out

go: error loading module requirements

Makefile:45: recipe for target 'fmt-check' failed

make: *** [fmt-check] Error 1

The command "make fmt-check" exited with 2.
```

MalloZup:
@jichenjc I have not forgotten your PR but I'm pretty busy this week. I will try to have look. stay tuned. thx

MalloZup:
why are not we using the API for this? 
http://libguestfs.org/guestfs-golang.3.html

zeenix:
> why are not we using the API for this?

Probably cause that will make it a build-time dep as well? As I said before, since this is only needed for one specific platform, it'd be best to not make it a build-time dep.

jichenjc:
@MalloZup sorry for this and take your time, let me know when you are free and I am happy to give some more detail if needed , as @zeenix mentioned, actually I have a API version code but I think it will lead to some run time link issue which I don't want to introduce to x86 platform..so I use this one 

I will see whether we can add the link issue only apply to s390 platform then we can switch to API version, but so far I didn't get that yet, thanks

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
 	return disk, nil
 }
 
+func getFirstDiskPath(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *libvirt.Connect) (string, error) {
+	prefix := fmt.Sprintf("disk.%d", 0)
+
+	if volumeKey, ok := d.GetOk(prefix + ".volume_id"); ok {
+		diskVolume, err := virConn.LookupStorageVolByKey(volumeKey.(string))
+		if err != nil {
+			return "", err
```
MalloZup: it is better here to add always a context of error for debugging 

https://github.com/dmacvicar/terraform-provider-libvirt/blob/b7dab08a9194e388a7f043a1ab7fe60d241b50e5/libvirt/cloudinit_def.go#L94

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
...
+		}
+		diskVolumeFile, err := diskVolume.GetPath()
+		if err != nil {
+			return "", err
```
MalloZup: ditto

libvirt/domain.go
jichenjc: sure, thanks~

libvirt/domain.go
jichenjc: ok

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
...
+		defer func() {
+			cmd := exec.Command("guestfish", "--remote", "--", "exit")
+			cmd.Run()
+		}()
+
+		cmd = exec.Command("guestfish", "--remote", "--", "findfs-label", "boot")
+		output, err := cmd.Output()
+
+		if err != nil {
+			return err
```
MalloZup: missing context of error

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
...
+		//guestfish --remote -- run
+		//guestfish --remote -- findfs-label
+		//guestfish --remote -- mount $2 /
+		//guestfish --remote -- mkdir-p $3
+		//guestfish --remote -- upload $4 $3/$4
+		//guestfish --remote -- umount-all
+		//guestfish --remote -- exit
+
+		// Get the root disk (first disk) volume location
+		// FIXME: currently assume first one will be the boot disk
```
MalloZup: Is better to have a Github ISSUE and removing the `fixme` if we can't fix this now

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
...
+		err = cmd.Start()
+		if err != nil {
+			return fmt.Errorf("Start command failed: %v", err)
+		}
+
+		// if the command start successfully, the output will be following format
+		// GUESTFISH_PID=xxxx; export GUESTFISH_PID, so we need get the PID and set by using os.Setenv
+		buf := new(bytes.Buffer)
+		buf.ReadFrom(stdout)
+		s := strings.Split(buf.String(), ";")
```
MalloZup: can we substitute `s` with `GUESTFISH_PID_LABEL`   
and `s1` with `GUESTFISH_PID_VALUE` something in this direction

libvirt/domain.go
jichenjc: I will open an issue after this PR merged

libvirt/domain.go
jichenjc: will update

jichenjc:
Thanks for the review, I will update unit test along with the doc later~

MalloZup:
yw! @zeenix what do you think about this approach? I'm not a coreos expert but I think ok, we can live with that.

Some open question: 

- I think we should consider to check that `guestfs` is at least our required version. Many OS systems can be in a XX situation. We should catch the situation in case the guestfs is not the minimal version we need.

I would do like this  number = `guestfs -version` . if the number < Required_version --> raise error to user saying installing a required one.

libvirt/domain.go
```diff
@@ -252,6 +385,15 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
 	return nil
 }
 
+func updateCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *libvirt.Connect) error {
```
zeenix: I find the function naming strange. Is the difference between `updateCoreOSIgnition` and `setCoreOSIgnition` that former updates while latter sets? I don't think so and hence the naming is not great. `setCoreOSIgnition` should be named to contrast with the new `inlineUpdateCoreOSIgnition` function.

libvirt/domain.go
```diff
@@ -252,6 +385,15 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain) err
...
+	if runtime.GOARCH == "s390x" {
+		// There is no support for fw_cfg in s390x, so instead doing fw_cfg
```
zeenix: grammar: instead -> instead of.

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
...
+	return "", fmt.Errorf("failed to find disk 0")
+}
+
+func getStderr(stderr io.ReadCloser) string {
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(stderr)
+	return buf.String()
+}
+
+func inlineupdateCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *libvirt.Connect) error {
```
zeenix: Let's CameCase properly: inlineUpdateCoreOsIgnition.

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
...
+	if ignition, ok := d.GetOk("coreos_ignition"); ok {
+		//basically we are doing following scripts
```
zeenix: * Space after the `//` please.
* "doing the following script" -> "doing the following".

libvirt/domain.go
```diff
@@ -230,6 +235,134 @@ func newDiskForCloudInit(virConn *libvirt.Connect, volumeKey string) (libvirtxml
...
+		//eval  $(guestfish --listen -a $1)
+		//guestfish --remote -- run
+		//guestfish --remote -- findfs-label
+		//guestfish --remote -- mount $2 /
```
zeenix: This comment is not as useful with the shell variables as reader doesn't know what they are. I'd suggest moving each line above its equivalent code below.

zeenix:
> yw @jichenjc! @zeenix what do you think about this approach? I'm not a coreos expert but I think ok, we can live with that.

Me neither. :)

> I think it is ok to have as workaround for the coreos issue.

Yeah if it's a simple workaround, it's fine.

> I think we should consider to check that `guestfs` is at least our required version. Many OS systems can be in a XX situation. We should catch the situation in case the guestfs is not the minimal version we need.

I could be wrong but I think guestfs doesn't every change it's interface and even if they do, it'll be backwards compatible.

MalloZup:
thx @zeenix for review so far :+1: :100: 

@hexfusion @tripledes @crawford @abhinavdahiya please take look on this. A feedback from your part would be appreciated. THX!

libvirt/domain.go
hexfusion: question: what about `s390` should that also be covered here?

libvirt/domain.go
jichenjc: ok, will do

libvirt/domain.go
jichenjc: we never tested s390, but you are right, technically it should be ok to do so

libvirt/domain.go
jichenjc: I made it to be fwcfgUpdateCoreOSIgnition

zeenix:
@jichenjc Thanks for the quick actions. Please update your patches always, instead of adding new patches so it's easy to tell which new changes are being introduced by the PR. Thanks.

jichenjc:
@zeenix someone told me I need add more commits to help people understand the new changes
are you suggesting I need squash those changes? Thanks

zeenix:
> @zeenix someone told me I need add more commits to help people understand the new changes

I guess some people aren't familiar with the good old way of code reviews coming from mailing-lists and they review all changes at onces, rather than per-commit.

> are you suggesting I need squash those changes?

Yes, if it's a fix to your patch.

MalloZup:
I think if we have some test unit or integration and doc on this it could be OK for me

tripledes:
Sorry, late to the party :smile: 

My only concern/comment would be to favor an API instead of shelling out commands, wouldn't the use of the API help with having tests as well?

If everyone is happy with it, forget I said anything :smile: still I also would like to see some tests and docs.

MalloZup:
Imho perhaps using api with guestfs, a part of making the code cleaner and testable, stable, it could also make sense a general feature. I might need to research if we could even perhaps make guest a resource operation which could modify other imgs. Need to look at

jichenjc:
The question is I hope we can avoid disrupt x86 feature as much as we can , I do have a 'API' version but this need dynamic link which I want to avoid (at least in my s390x env, I install guestfish and then I can run the command but if I use API, I need link the .so file during make stage and this will push a big requirement on x86 ) @tripledes 

how about let's add some unit test and doc first then merge this PR and let's open another PR to track how to add guest resource then after that if x86 need this as well then we can have well design and implement it ? Thanks @MalloZup

zeenix:
> how about let's add some unit test and doc first then merge this PR and let's open another PR to track how to add guest resource then after that if x86 need this as well then we can have well design and implement it ?

Sounds good to me at least.

tripledes:
> The question is I hope we can avoid disrupt x86 feature as much as we can , I do have a 'API' version but this need dynamic link which I want to avoid (at least in my s390x env, I install guestfish and then I can run the command but if I use API, I need link the .so file during make stage and this will push a big requirement on x86 ) @tripledes
> 
> how about let's add some unit test and doc first then merge this PR and let's open another PR to track how to add guest resource then after that if x86 need this as well then we can have well design and implement it ? Thanks @MalloZup

What about Go's [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints)? I've never used them myself but would those be useful in this case? In x86 the dependency would still be pulled but not used if I understand it properly.

MalloZup:
We also need to  consider that this PR can be spared if we use a null-resource `guestfs`. https://www.terraform.io/docs/provisioners/null_resource.html where you literally do all the work we do in code here with a bash thing.

In theory that could be the most easiest and without any coding requirement, unless i'm not seeing things.

Regarding build constraintst I also never used them but could be something worth to search

MalloZup:
just let me resume our talks here, we have following possibilities, (let me order them in costs/efforts)

- [ ] use  a `null-resource` (https://www.terraform.io/docs/provisioners/null_resource.htm): no code

- [ ] use a shelling go PR (the current-approach), need documentation and tests (integrations). 

- [] API approach with guestfs (adding build dep or use build constraints)
    * It can be used only here only on coreos 
    * it can be extended to be a resource/operation so it is generic ( need research/tests/code/doc)

jichenjc:
@MalloZup I went through the null_resource, seems I have to modify the tf file in order to achieve that ? what I want to achieve is old tf files works on x86 should still works on our platform...

so if the tf file need to be kept, I have to stick to our current PR and I will open another issue to track how to move to no dependency way and try it later,  please let me know your comments and suggestions, Thanks a lot for your great input and help~

MalloZup:
yes with the null_resource you need to add the shell script inside it and adapt it to the env and dependencies. If other people are ok with this approach I'm ok also. 

I would be more picky if it was affecting the core functionality of the providers but since it is a special case, I'm ok to have with this approach  but with test + doc is needed.

MalloZup:
I'm fine with this pr approach resuming

MalloZup:
feel free to ping when you have some unit/integration test + doc

MalloZup:
Hi! I honestly add some difficulty to understand at first glance what we are testing in the higher level . Any help in this direction is welcome. I will have look later

jichenjc:
@MalloZup this test comes from https://github.com/openshift/cluster-api-provider-libvirt/pull/174
as those 2 are doing similar things, so I referred to that PR :)

the test itself is trying to mock the execute command and start the test, and make sure every guestfs command got executed correctly with correct param ,as we can't control the guestfs behavior but we need make sure we call the guestfs command with desired param and in correct order ( the state machine introduced in the test)

please let me know whether you need additional info I am happy to provide

libvirt/domain_test.go
```diff
@@ -0,0 +1,248 @@
+package libvirt
+
```
MalloZup: I would rename `domain_test.go` with `guestfish_test.go` or something that doesn't impact a `domain` imho

MalloZup:
@jichenjc  ok thx. Looks good to me. @zeenix @tripledes @hexfusion  feel free to review. Otherwise can merge this imho.

zeenix:
@MalloZup Sounds good to me. I think we can merge this.

MalloZup:
thx @jichenjc @zeenix  for this PR . merging

jichenjc:
thanks a lot for your help~

PR #655, opened 2019-10-08
Initial support for generating local documentation

dmacvicar:
We have the _website_ directory in the tree, using upstream layout to be prepared for inclusion in the _terraform-providers_ project.

This PR adds a _docs_ directory with a Sphinx configuration that allows to generate html documentation from the original layout.

```bash
cd docs
make html
```

Should generate a beautiful website:

Screenshot from 2019-10-09 00-57-19
Screenshot from 2019-10-09 00-57-45

There are minor details that can be improved, but I think it is a good first step.

After this PR, my plan would be to enable CI to automatically publish the documentation, and then move some of it from _README_, and link it appropriately.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,0 +1,2 @@
+build
+__pycache__
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,21 @@
+# Minimal makefile for Sphinx documentation
+#
+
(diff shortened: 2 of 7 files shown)


MalloZup:
@dmacvicar :heart: :heart: :heart_eyes_cat: :dog: :heart_eyes:

PR #661, opened 2019-10-24
Support execute guestfish at remote libvirt host

### LINKED ISSUE #659: [s390x] coreos ignition can not work in remote libvirt mode
##  System Information
Arch: s390x

### Linux distribution
RHEL7

### Terraform version
```Terraform v0.12.11-dev
+ provider.libvirt (unversioned)```
master branch


```terraform-provider-libvirt c611c1dce50e8c6a04aa39938147c6c5acb60622-dirty
Compiled against library: libvirt 4.5.0
Using library: libvirt 4.5.0
Running hypervisor: QEMU 2.12.0
Running against daemon: 4.5.0
```
also master branch

___

## Checklist

This is s390x arch only issue.

fw_cfg is not supported for s390x, instead, using guestfish to inject coreos ignition config file. This works for local libvirt, but but remote node failed due to 
https://github.com/dmacvicar/terraform-provider-libvirt/pull/629/files#diff-f6f8db75203b0db83110f024f21cc4ebR400

In the code,  terraform provider assume the volume can be found locally while actually its in remote node (running libvirtd) 
 
so the root cause is terraform use guestfish method to update ingition volume failed due to they are on different host

### Setup

(Please provide the full _main.tf_ file for reproducing the issue (Be sure to remove sensitive information)

### Steps to Reproduce Issue

(Include debug logs if possible and relevant).

___
## Additional information:

bjhuangr:
In remote libvirt mode, libvirt volumes will be created on remote
host, but guestfish commands are running on terraform host.

In this change, will run those guestfish commands at remote
libvirt host via ssh, And this change assumes
the public key of the user that running terraform has been added
into libvirt host's authorized_keys.

Fixed #659

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"regexp"
 	"runtime"
 	"sort"
 	"strconv"
@@ -17,6 +18,7 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/dmacvicar/terraform-provider-libvirt/libvirt/helper/sshconn"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 	libvirt "github.com/libvirt/libvirt-go"
@@ -260,7 +262,6 @@ func getStderr(stderr io.ReadCloser) string {
 
 func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *libvirt.Connect) error {
 	if ignition, ok := d.GetOk("coreos_ignition"); ok {
-
 		// Get the root disk (first disk) volume location
 		rootPath, err := getFirstDiskPath(d, domainDef, virConn)
 		if err != nil {
@@ -280,7 +281,11 @@ func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Do
 		if err != nil {
 			return err
 		}
-		err = guestfishExecution(rootPath, fileToUpload)
+		connURI, err := virConn.GetURI()
+		if err != nil {
+			return err
+		}
+		err = guestfishExecution(rootPath, fileToUpload, connURI)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


bjhuangr:
WIP, testing this change.

bjhuangr:
Tested in remote libvirt mode, the change works fine, ignition config files can be injected into remote libvirt volume, and the coreos and ignition working correctly.

jichenjc:
@zeenix can you provide some initial feedback on this PR then we can ask more folks to take a look? thanks

libvirt/domain.go
```diff
@@ -280,17 +280,103 @@ func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Do
...
 		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func guestfishExecution(rootPath string, fileToUpload string, connURI string) error {
+	// get remote host from libvirtd remote URI
+	virHost := strings.Split(strings.SplitAfter(connURI, "//")[1], "/")[0]
```
zeenix: Slightly sceptical of this host extraction code but in my limited testing, it works fine: https://play.golang.org/p/Sj1UYFc2LSf . Only thing is that you get the [authority component of the URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax), which could be more than the hostname (in most cases a `username@host`) so please ensure the code doesn't break on that case.

zeenix: Oh and it doesn't account for local URLs with hostname (e.g qemu://localhost/system). You'd want a regex match like we did in Boxes: https://gitlab.gnome.org/GNOME/gnome-boxes/blob/master/src/machine.vala#L66

bjhuangr: Agreed, I'll add a helper function to get libvirt host, by considering more cases.

bjhuangr: Fixed this.

bjhuangr:
Updated the change to fix some comments, tested both local and remote mode, work fine in my env.

bjhuangr:
This change is just a workaround. The long term solution is in discussion.
And the change only affects the code branch of `s390x` and `remote libvirt` .

bjhuangr:
@MalloZup @zeenix Any further comments on this change? Thanks!

MalloZup:
@bjhuangr I will look at this next. Stay tuned

libvirt/domain.go
```diff
@@ -280,17 +281,117 @@ func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Do
...
+		err = guestfishExecution(rootPath, fileToUpload, connURI)
 		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func getLibvirtHostFromURI(connURI string) string {
+	// example uri with all feilds: schema://userinfo@host:port/system
```
zeenix: I'd instead provide [this link](https://libvirt.org/uri.html) here.

libvirt/helper/sshconn/sshconn.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return err
+	} else {
+		s.Client = client
+		return nil
+	}
+}
+
+func (s *SSHConn) CombinedOutput(cmd string) (string, error) {
+	session, err := s.Client.NewSession()
```
zeenix: * can't we keep the session alive so it can be reused for all the calls?
* I know you named the function after the underlying SSH API call but it's not very intuitive. I had to read the code carefully before I realized that you're executing a command rather than combining some output.

libvirt/domain.go
```diff
@@ -280,17 +281,117 @@ func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Do
...
+	if strings.Contains(virHost, "@") {
+		virHost = strings.SplitN(virHost, "@", 2)[1]
+	}
+	virHost = strings.Split(virHost, ":")[0]
+	return virHost
+}
+
+func guestfishExecution(rootPath string, fileToUpload string, connURI string) error {
+	// If the adress is in the 127.0.0.0 block or is localhost, then it is local
+	// Another local case example: qemu+tcp:///system
```
zeenix: not sure how to interpret the second line of the comment. Do you mean this is an example that is not covered? The word "Another" is confusing here.

libvirt/domain.go
```diff
@@ -280,17 +281,117 @@ func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Do
...
+	localReg := regexp.MustCompile(`:\/\/(127\.\d+\.\d+\.\d+|localhost)|///`)
+	isLocal := localReg.MatchString(connURI)
+	if !isLocal {
```
zeenix: isLocal variable is redudant here IMO since it's only used once. I'd go for:

```
If !localReg.Match.. {
   // Remote URI
   ...
} else {
   // Local URI
  ...
}
```

libvirt/domain.go
bjhuangr: Agreed

libvirt/domain.go
bjhuangr: Yes, the second comment line is an example that is not covered.

libvirt/domain.go
bjhuangr: Agreed.

libvirt/helper/sshconn/sshconn.go
bjhuangr: - We can't keep session for reuse. A Session only accepts one call to Run, Start, Shell, Output, or CombinedOutput.  Refer to: https://godoc.org/golang.org/x/crypto/ssh#Session.Run
- Agreed to change the function name

bjhuangr:
> > This change is just a workaround.
> 
> Is this a workaround from the POV of terraform as well? I-e is there any possible solution for the terraform provider (not Openshift)? If so, please mention the possible solution in the commit message. If not, I'd not call it a workaround.

From terraform libvirt provider POV, it is not a workaround. Have removed the statement of `workaround` from commit message.

libvirt/helper/sshconn/sshconn.go
zeenix: >     * We can't keep session for reuse.

Ah ok, strange concept of a "session" they have then. :)

libvirt/domain.go
zeenix: Thanks. I see that you removed it completely now but i don't see any changes in the code. Please add a clearer FIXME comment here. Assuming such URLs are not treated as local, something like `FIXME: URLs of the form qemu+tcp:///system not covered & will be treated as remote`.

bjhuangr: It does not need a FIXME here I think, as you can see `localReg := regexp.MustCompile(`:\/\/(127\.\d+\.\d+\.\d+|localhost)|///`)`, which can identify uri that contains `///` as local. Sorry for the confusing. :)

libvirt/domain.go
```diff
@@ -280,17 +281,115 @@ func inlineUpdateCoreOsIgnition(d *schema.ResourceData, domainDef *libvirtxml.Do
...
 		}
 	}
 	return nil
 }
 
+func getLibvirtHostFromURI(connURI string) string {
+	// Refer to libvirt URI: https://libvirt.org/uri.html
+	virHost := strings.Split(strings.SplitN(connURI, "//", 2)[1], "/")[0]
+	if strings.Contains(virHost, "@") {
+		virHost = strings.SplitN(virHost, "@", 2)[1]
```
zeenix: Coincidently the `username@` URL is mostly likely used with SSH libvirt URI so do we really want to ignore username in the URL?

bjhuangr: In this change, username defaults to `root`, since normally username should be `root` in a libvirt URI, as libvirtd run as user `root`. And the ssh remote call need write access to modify the remote libvirt volume.

zeenix: Not at all. system libvirt demon runs as root but you don't have to be root to connect to it. There is hope we (Openshift) will move away from system demon in the future. Allowing remote root access is not the best idea anyway. So no, this assumption is not correct IMO.

jaypoulz:
Based on input from @crawford today, we should not pursue SSH based guestfs commands as our solution to this problem. It introduces a security vulnerability where the cluster can ssh into the hypervisor it is running on and wreak havoc.

zeenix:
@jaypoulz so the guestfish solutions that were already merged specifically for S390 won't work for S390 either?

jichenjc:
@zeenix that PR works fine locally and without issue, only the ssh way is not a good way as it requires ssh logon, now the proposal to use remote guestfish which won't have ssh connection at all

zeenix:
> @zeenix that PR works fine locally and without issue, only the ssh way is not a good way as it requires ssh logon, now the proposal to use remote guestfish which won't have ssh connection at all

Thanks. That's great to hear. :)

Issue #622, opened 2019-08-07
guest is not shutdown after changing running to "false"


muroj:
### Linux distribution

``` Ubuntu 18.04.2 LTS```

### Terraform version

```sh
Terraform v0.11.14
+ provider.libvirt (unversioned)
+ provider.template v2.1.2
```

### Provider and libvirt versions

```sh
terraform-provider-libvirt 0.5.2
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
```
___

## Checklist

- [ ] Is your issue/contribution related with enabling some setting/option exposed by libvirt that the plugin does not yet support, or requires changing/extending the provider terraform schema?

  - [ ] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.
  - [ ] If it is a very special case, consider using the _XSLT_ support in the provider to tweak the definition instead of opening an issue
  - [ ] Maintainers do not have expertise in every libvirt setting, so please, describe the feature and how it is used. Link to the appropriate documentation

- [x] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

## Description of Issue/Question

Changing `running = "false"` domain argument has no effect. Yet `terraform apply` reports all that update was successful. Same as #119 

```
terraform apply
data.template_file.network_config: Refreshing state...
data.template_file.user_data: Refreshing state...
libvirt_cloudinit_disk.commoninit_guest1: Refreshing state... (ID: /guestimages/data1/terraform/joe_terraf...o;5d4adfc2-9d4c-9f28-f090-f002d2df7cd0)
libvirt_volume.guest1: Refreshing state... (ID: /guestimages/data1/terraform/joe_terraform_demo.qcow2)
libvirt_domain.guest1: Refreshing state... (ID: 8680a09d-72c8-4210-a31a-0733d4f7d1c6)

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  ~ libvirt_domain.guest1
      running: "true" => "false"


Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

libvirt_domain.guest1: Modifying... (ID: 8680a09d-72c8-4210-a31a-0733d4f7d1c6)
  running: "true" => "false"
[...]

PR #668, opened 2019-11-21
fix(domain): update domain running state when refreshed

Naewis:
### Linux distribution

``` Debian GNU/Linux 10 (buster) ```

### Terraform version

```sh
Terraform v0.12.12
```

```sh
Compiled against library: libvirt 5.0.0
Using library: libvirt 5.0.0
Running hypervisor: QEMU 3.1.0
Running against daemon: 5.0.0
```
___

## Checklist

Your issue is it a bug or something that does not work as expected.

## Description of Issue

I'm currently using libvirt locally on my PC. Whenever I shutdown my computer, all the domains are stopped (which is correct, I don't wan't some tests domain to restart whenever I restart the host). Until then, all's good.

But when I want to restart my stack after a reboot, terraform says that there is nothing to do. As the domains are in stopped state, I expected terraform to put them online again (in order to get immutable infrastructure). I noticed that the running property doesn't get updated when terraform tries a refresh.

### What did I do

I used the method domainIsRunning in order to get the domain state, and updated the schema running property. Tests were all good, but there seems to be a bug which is not related to this PR.

___
## Additional information:

Apparmor is disabled.

First PR, and first GO code : sorry if I did something wrong in the way.
Please tell me in order to improve.

On a side note, there is also something wrong with the network state refresh.
When I ran the tests, i got this :
```
Error: Error creating libvirt domain: virError(Code=55, Domain=19, Message='Requested operation is not valid: network 'default' is not active')

$ sudo virsh net-list --all
 Name          State      Autostart   Persistent
--------------------------------------------------
 default       inactive   no          yes
```

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -563,10 +563,15 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	// We save runnig state to not mix what we have and what we want
+	requiredStatus := d.Get("running")
+
 	err = resourceLibvirtDomainRead(d, meta)
 	if err != nil {
 		return err
 	}
+
+	d.Set("running", requiredStatus)
 
 	// we must read devices again in order to set some missing ip/MAC/host mappings
 	for i := 0; i < d.Get("network_interface.#").(int); i++ {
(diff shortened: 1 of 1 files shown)


MalloZup:
@Naewis  thx for PR.  I will have look soonish :+1:

MalloZup:
related on same context : https://github.com/dmacvicar/terraform-provider-libvirt/issues/622

Naewis:
As far as I've tested, this patch fixes the issue #622, but the fix is not really elegant, I must agreed, as I know nothing of Golang (testing seems too much at my current level)

Saddly, I have a little time to fix this issue anymore (as learning Go is not in my todo list for now, maybe later), but if I were fluent in Go, I would fix some issues by not re-reading domain global state when we create the resource (l. 569 : ```err = resourceLibvirtDomainRead(d, meta)``` ) but instead use a separate variable and reinject only ip properties into our current state (would fix the state matter with more style).

I'm also locked as I don't know which property is actually erased implicitly vs not voluntarily erased

Naewis:
Is there anything I need to do anymore for this request ?

MalloZup:
thx @Naewis  for pr and sorry for delay :ping_pong:

Issue #662, opened 2019-10-31
Adding a block device to a domain via xslt override causes terraform crash


muroj:
### Linux distribution

``` Ubuntu 18.04.3 LTS ```

### Terraform version

```sh
terraform -v
Terraform v0.12.12
+ provider.libvirt (unversioned)
+ provider.template v2.1.2
```

### Provider and libvirt versions

```sh
terraform-provider-libvirt -version
./terraform-provider-libvirt 0.6.0+git.1569597268.1c8597df
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
2019/10/31 15:48:57 virError(Code=38, Domain=7, Message='Failed to connect socket to '/run/user/1000/libvirt/libvirt-sock': No such file or directory')
```
___

## Checklist

- [ ] Is your issue/contribution related with enabling some setting/option exposed by libvirt that the plugin does not yet support, or requires changing/extending the provider terraform schema?

  - [ ] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.
  - [ ] If it is a very special case, consider using the _XSLT_ support in the provider to tweak the definition instead of opening an issue
  - [ ] Maintainers do not have expertise in every libvirt setting, so please, describe the feature and how it is used. Link to the appropriate documentation

- [x] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

### Setup

I want to boot a domain from a block device on the host. According to [domain doc](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/domain.html.markdown#handling-disks) it doesn't look like block-based disk devices are directly supported only files, urls, and volumes. So I am trying to use xslt to insert block devices into domain dynamically.

Here is example xml of what I am looking to do:

```xml
<disk type='block' device='disk'>
      <driver name='qemu' type='raw'/>
      <source dev='/dev/disk/by-id/dm-uuid-mpath-20017380030bb13be'/>
      <backingStore/>
      <target dev='vda' bus='virtio'/>
      <alias name='virtio-disk0'/>
      <address type='ccw' cssid='0xfe' ssid='0x0' devno='0x0000'/>
    </disk>
```

main.tf
```hcl
provider "libvirt" {
    uri = "qemu+ssh://kvmuser@kvmhost/system"
}

variable "scsi_disks" {
  description = "Disks to attach to the domain as block devices"
  type = map
  default = {

    disks = [
      { 
[...]

Issue #418, opened 2018-09-16
Block device not supported


Vanuan:
#  Version Reports:

### Distro  version of host:

``` Ubuntu 18.04 ```

### Terraform Version Report
(Provided by running `v0.11.8`.)

### Libvirt version

```4.0.0```

### terraform-provider-libvirt plugin version (git-hash)

```terraform-provider-libvirt-0.4.4.Ubuntu_18.04.amd64.tar.gz```
___
# Description of Issue/Question

Reading documentation I can't find a way to attach disk from block device.
I'm looking for an equivalent of:

```virsh attach-disk storageVM /dev/disk/by-id/ata-HGST_HUSXXX sdb```

I've tried this:

```
resource "libvirt_domain" "test" {
  name = "storageVM"
  memory = "1024"
  vcpu = 1

  disk {
    file = "/dev/disk/by-id/ata-HGST_HUSXXX"
    scsi = true
  }

}
```

Doesn't work. Here's a relevant xml snippet returned from dumpxml:

```
...
    <disk type='block' device='disk'>
      <driver name='qemu' type='raw'/>
      <source dev='/dev/disk/by-id/ata-HGST_HUSXXX'/>
      <backingStore/>
      <target dev='sdb' bus='scsi'/>
      <alias name='scsi0-0-0'/>
      <address type='drive' controller='0' bus='0' target='0' unit='0'/>
    </disk>
...
```

**Comment by MalloZup** [comment]
@Vanuan thx for issue. 

If you want to add it and contribute i can help through the code.

Otherwise , yop it is at moment a missing feature that we might add.

Atm we plan 2 prs for next releases, and other things.

To me  i can have a look once i am finished with my tasks.

For any info feel free to ask thx! :sunflower:

**Comment by dmacvicar** [comment]
Quickly thinking about it probably we need to add `device` like we have `file` and `source` (the last being unrelated to libvirt `source` xml block).

**Comment by Vanuan** [comment]
Attach disk required domain, device and target.
It looks like disk def needs to be extended with `diskDef.Source.Block` type.
And target is better be specified.

So something like this would be most intuitive:
```
resource "libvirt_domain" "local-domain" {
  ...
  disk {
    device = "/dev/disk/by-id/ata-HGST_HUSXXX"
    target = "hda"
    scsi = true
  }
  disk {
    device = "/dev/disk/by-id/ata-HGST_HUSYYY"
    target = "hdb"
    scsi = true
  }
  ...
}
[...]

PR #670, opened 2019-11-25
Block device support for domains

muroj:
My first pass at providing support for block devices. Attempts to address #662 and #418. (Disclaimer: I am new to go).

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -254,10 +254,10 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 	if _, ok := d.GetOk("source"); ok {
 		err = img.Import(newCopier(client.libvirt, volume, volumeDef.Capacity.Value), volumeDef)
 		if err != nil {
-			//  don't save volume ID  in case of error. This will taint the volume after.
-			// If we don't throw away the id, we will keep instead a broken volume.
-			// see for reference: https://github.com/dmacvicar/terraform-provider-libvirt/issues/494
-			d.Set("id", "")
+		//  don't save volume ID  in case of error. This will taint the volume after.
+		// If we don't throw away the id, we will keep instead a broken volume.
+		// see for reference: https://github.com/dmacvicar/terraform-provider-libvirt/issues/494
+		d.Set("id", "")	
 			return fmt.Errorf("Error while uploading source %s: %s", img.String(), err)
 		}
 	}
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -254,10 +254,10 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 	if _, ok := d.GetOk("source"); ok {
 		err = img.Import(newCopier(client.libvirt, volume, volumeDef.Capacity.Value), volumeDef)
 		if err != nil {
-		//  don't save volume ID  in case of error. This will taint the volume after.
-		// If we don't throw away the id, we will keep instead a broken volume.
-		// see for reference: https://github.com/dmacvicar/terraform-provider-libvirt/issues/494
-		d.Set("id", "")	
+			//  don't save volume ID  in case of error. This will taint the volume after.
+			// If we don't throw away the id, we will keep instead a broken volume.
+			// see for reference: https://github.com/dmacvicar/terraform-provider-libvirt/issues/494
+			d.Set("id", "")
 			return fmt.Errorf("Error while uploading source %s: %s", img.String(), err)
 		}
 	}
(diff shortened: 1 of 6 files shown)


libvirt/resource_libvirt_volume.go
```diff
@@ -254,10 +254,10 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 	if _, ok := d.GetOk("source"); ok {
 		err = img.Import(newCopier(client.libvirt, volume, volumeDef.Capacity.Value), volumeDef)
 		if err != nil {
-		//  don't save volume ID  in case of error. This will taint the volume after.
-		// If we don't throw away the id, we will keep instead a broken volume.
-		// see for reference: https://github.com/dmacvicar/terraform-provider-libvirt/issues/494
-		d.Set("id", "")	
+			//  don't save volume ID  in case of error. This will taint the volume after.
```
MalloZup: perhaps we can remove the identation here since not related but ok

MalloZup:
checkout travis is failing

```
+go test -v -covermode=count -coverprofile=profile.cov -timeout=1200s -test.failfast ./libvirt

# github.com/dmacvicar/terraform-provider-libvirt/libvirt [github.com/dmacvicar/terraform-provider-libvirt/libvirt.test]

libvirt/resource_libvirt_domain_test.go:1221:43: cmd.String undefined (type *exec.Cmd has no field or method String)

libvirt/resource_libvirt_domain_test.go:1240:43: cmd.String undefined (type *exec.Cmd has no field or method String)

FAIL	github.com/dmacvicar/terraform-provider-libvirt/libvirt [build failed]

Makefile:36: recipe for target 'testacc' failed

make: *** [testacc] Error 2```

**Comment by muroj** [comment]
I am running same build commands locally and not seeing the errors:

```
:~/go/src/github.com/muroj/terraform-provider-libvirt$ make build
go fmt ./libvirt .
go run golang.org/x/lint/golint -set_exit_status ./libvirt .
go vet ./libvirt
```

Is it because I am running `go1.13.4` instead of `go1.11.13`?

**Comment by MalloZup** [comment]
@muroj  dunno I didn't tried but try it :grin:

File: libvirt/resource_libvirt_volume.go (hunk 1 above)
Comment (muroj) [reply libvirt/resource_libvirt_volume.go]: I tried reverting it back to original. I think the changes were the result of `go-fmt`.

**Comment by MalloZup** [comment]
@muroj thx for pr great job

**Comment by dmacvicar** [comment]
While trying to do a release I realize this PR broke integration tests.

It requires both root access to test this, and on top, it assumes _libvirtd_ is running on the same machine as the test, and assumes _/usr/sbin_ is on _$PATH_ (eg.  mkfs.ext4).

**Comment by dmacvicar** [comment]
Managed to fix this calling executables with full path (sbin). I am still stuck in:

```
Jan 03 13:36:32 piscolita.suse.de libvirtd[5524]: internal error: process exited while connecting to monitor: 2020-01-03T12:36:32.234735Z qemu-system-x86_64: -blockdev {"driver":"host_device","filename":"/tmp/cgvuajorbpkmtkbuetximqhkpcetcwlby","node-name":"libvirt-1-storage","auto-read-only":true,"discard":"unmap"}: 'host_device' driver expects either a character or block device
```

Which I guess comes from:

```golang
var configBlockDevice = fmt.Sprintf(`

resource "libvirt_domain" "%s" {
	name = "%s"
	
	disk {
		block_device = "%s"
	}

}`, randomDomainName, randomDomainName, tmpfile)
```

Which makes total sense. `tmpfile` is not a device. The loop device is a device. I wonder how this works in the libvirt used in travis.

If I change the code to use the loop device, I fall into this:

```golang
Error: Error creating libvirt domain: virError(Code=1, Domain=10, Message='internal error: process exited while connecting to monitor: 2020-01-03T13:19:32.338303Z qemu-system-x86_64: -blockdev {"driver":"host_device","filename":"/dev/loop0","node-name":"libvirt-1-storage","auto-read-only":true,"discard":"unmap"}: The device is not writable: Permission denied')
```

Which makes sense, as _losetup_ is called as:

```golang
// Mount the file to a loop device
cmd = exec.Command("sudo", "/sbin/losetup", "--read-only", strings.TrimRight(string(loopdev), "\n"), filename)
fmt.Printf("Executing command: %s\n", strings.Join(cmd.Args, " "))
```

Which means we are setting up a _rw_ disk on a _ro_ device. However, we haven't implemented _readonly_ for disks.

Fixing _losetup_, I need to fix the check function to check that the source is the device.

Fixing that, I get an error that _tmpfile_ is not used. Looks like the temporary file is not cleaned. I have fixed that as well.

Fixes being prepared here:
https://github.com/dmacvicar/terraform-provider-libvirt/tree/dmacvicar_fix_tests_locally

MalloZup:
@dmacvicar thx and sorry for not having catching it up.

Let me know if I can help somehow. Thx!

dmacvicar:
I put all the fixes in master before doing the release! thanks!

The PR was passing testacc fine on Travis! so not a review problem. openSUSE/libvirt were slightly different.

muroj:
@dmacvicar @MalloZup  Thanks both for your work (and patience) with me on this.

PR #674, opened 2019-12-01
Add dhcp hosts to network resoruce 

goraxe:
This is to allow defining static host leases at the network level 

### DIFF
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -19,6 +19,7 @@ const (
 	netModeRoute    = "route"
 	netModeBridge   = "bridge"
 	dnsPrefix       = "dns.0"
+	dhcpPrefix      = "dhcp.0"
 )
 
 // a libvirt network resource
@@ -214,6 +215,27 @@ func resourceLibvirtNetwork() *schema.Resource {
 							Default:  true,
 							Optional: true,
 							Required: false,
+						},
+						"hosts": {
+							Type:     schema.TypeList,
+							ForceNew: false,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"ip": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"name": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+									"mac": {
+										Type:     schema.TypeString,
+										Optional: true,
+									},
+								},
+							},
 						},
 					},
 				},
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -19,7 +19,7 @@ const (
 	netModeRoute    = "route"
 	netModeBridge   = "bridge"
 	dnsPrefix       = "dns.0"
-	dhcpPrefix      = "dhcp.0"
+	dhcpPrefix      = "dhcp"
 )
 
 // a libvirt network resource
@@ -216,7 +216,7 @@ func resourceLibvirtNetwork() *schema.Resource {
 							Optional: true,
 							Required: false,
 						},
-						"hosts": {
+						"host": {
 							Type:     schema.TypeList,
 							ForceNew: false,
 							Optional: true,
@@ -342,6 +342,12 @@ func resourceLibvirtNetworkUpdate(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Error updating DNS hosts for network %s: %s", networkName, err)
 	}
 
+	// FIXME assuming only one dhcp address range can be active
+	err = updateDHCPHosts(d, 0, network)
+	if err != nil {
+		return fmt.Errorf("Error updating DHCP hosts for network %s: %s", networkName, err)
+	}
+
 	// detect changes in the bridge
 	if d.HasChange("bridge") {
 		networkBridge := getBridgeFromResource(d)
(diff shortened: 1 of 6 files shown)


MalloZup:
thx @goraxe for pr! will review on free time stay tuned!

MalloZup:
@inercia  hi! just quick question. How looks to you this PR? If you have time just take look .tia

inercia:
> @inercia hi! just quick question. How looks to you this PR?
> I don't remember if we supported already this use-case or not.

We currently have the ability to define `hostname`/`IP`/`MAC` for the network interfaces attached to the domain (see [this](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/domain.html.markdown#handling-network-interfaces)). This seems to be more like the ability to create _any_ `hostname`/`IP`/`MAC` association in that network (as libvirt is running a `dnsmasq` server there).

I've never been a big fan of this kind of things, but we already support random `hostname`/`IP` associations in the network (see [this](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/website/docs/r/network.markdown)), so this would be oine step further.

All that said, maybe it would make more sense to add the `MAC` address in the `hosts` block we already have, like:

```hcl
hosts  {
     hostname = "my_hostname"
     ip = "my.ip.address.1"
    mac = "00:00:00:00:00:00"
   }
```

because the underlying library considers these three things as a whole (see [this](https://github.com/libvirt/libvirt-go-xml/blob/master/network.go#L146-L152)). But it would also make more sense to take it out of the `dns` block...

goraxe:
@inercia I tried the ip defined in domain resource. My use case was to to have two different network interfaces attached to a domain one with external internet access (either bridge or nat) and one private network for the domains with known ip addresses.  I needed the IP addresses for the internal network to be static.  When using the ip address assignment as defined on the host resource this just didn't work.  Inspecting the libvirt objects created, I could not see the requested ip addresses making it into the network dhcp section.  I spent some time trying to make that work and could not, so I created this PR ;-)

The model of this within libvirt is dns names are a top level entity within a network model
https://github.com/libvirt/libvirt-go-xml/blob/master/network.go#L237-L258

dhcp host entries are a little further down belonging to a range which belongs to an IP section which is a list of IP ranges which are attached to the network (ie ipv4 & ipv6),  
https://github.com/libvirt/libvirt-go-xml/blob/master/network.go#L165-L173.  So there are a few internal assumptions the provider needs to make to be able to build the xml to send to libvirt

If you are aiming for a 'thin' layer over libvirt the model would kinda look like 

```terraform 
resource "libvirt_network" "kube_network" { 

  # single block
  dns {
     host {
            hostname = "gw"
            ip               = "192.168.0.1"
     }
  }

  # multiple bocks
  ip {
      address = "192.168.0.0"
      prefix     = "24
      family     = "ipv4"
      # single block
      dhcp { 
          range {}
          bootp {}
          # multiple blocks
          host { 
              ip = "my.ip.address.1"
              mac = "00:00:00:00:00:00"
          }
      }
  }
}
```

There is an argument this is too flexible / complex and so you guys might want to expose a simpler model.

to summarizes my use case a bit
Bring up multiple domains, install various types of clustered software, automate testing of install and recovery responses to various fault simulations.
Requirements
* multiple interfaces per domain / vm
* pre-defined ip addresses on private interfaces
* ability to define external dns entries

nice to have
* define dhcp start and end range  (most the software I'm using can figure out node ID from ip address so if I can see the ip address pool is only 5 addresses, then its easy to one shot the config files, if ip addresses are more dynamic I need to collect the ip addresses to build the config files which means more logic & external provisioner to terraform)

betaboon:
any progress on this? i really would like to use this feature :)

goraxe:
@betaboon its been a while since I worked on this and would need to re-sync against current master and remember things.  In principle I would be willing to work on the feature I would just need some guidance from @inercia and @MalloZup on how they would like the data model to work

betaboon:
@goraxe I'm currently running the code from this PR with just one minor change applied onto the current release.

MalloZup:
I guess  I will need to refresh my knowledge on this PR seeing both pov

libvirt/resource_libvirt_network.go
```diff
@@ -320,6 +342,12 @@ func resourceLibvirtNetworkUpdate(d *schema.ResourceData, meta interface{}) erro
 		return fmt.Errorf("Error updating DNS hosts for network %s: %s", networkName, err)
 	}
 
+	// FIXME assuming only one dhcp address range can be active
+	err = updateDHCPHosts(d, 0, network)
```
dmacvicar, line 346: Hi. Thanks for the PR.

The implementation looks good. I am still unsure where we should merge this, and this line highlights why.

The original design of _terraform-provider-libvirt_ was to act as a simple DigitalOcean-like Cloud with Libvirt. You start a VM easily, and therefore the defaults where a single interface, NAT and DHCP.

It wasn't until people used this for testing, especially Kubernetes, when complex network setups started to arise.

At that time, networking was extended so that `addresses` is a plain array of addresses, therefore `dhcp`continues being a global block instead of one associated by address.

This design is what also has me unsure of merging https://github.com/dmacvicar/terraform-provider-libvirt/pull/725, in the past NAT was default, and creating a proper `forward` block means NAT is not longer the default.

I am not sure what the is the right thing to do. A new network schema that specifies `ip` blocks and their own `forward`and `dhcp`blocks would prevent this kind of hacks. At the same time, it would make a default schema more complex than it is today. The libvirt schema for network is deeply nested.

Given that we are seeing increasingly complex network configurations, I am inclined for a full network resource re-write, following the libvirt-schema almost 1:1, and add this kind of features there. After all, both network resources could coexist for a while (eg `libvirt_network_v2`).

dmacvicar (commented):
I would prefer not to merge this due to the "FIXME"s it contains, which can't be fixed until we re-do the network provider.

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #678, opened 2019-12-07
Use config-drive instead of guestfish on s390x

crawford:
This adds support for using the config-drive mechanism to deliver Ignition configs and replaces the guestfish method. Using libguestfish to inject the Ignition config directly into the root disk proved to be difficult to implement when targeting a remote libvirtd instance. This approach instead uses the OpenStack config-drive mechanism which is similar to how cloud-configs are provided to guests. In order for this to work, the Ignition instance in the guest needs to be configured to read its config from the OpenStack provider.


crawford:
I'm proposing this as an alternative to #666. This approach has the advantage that the content in the `coreos_ignition` resource definitions can just be plain text, as it is on other platforms today. In #666, the content needs to be a properly formatted config-drive.

MalloZup:
@crawford  thx for pr ! :+1: I will look at soon

crawford:
Yeah, I can absolutely add some examples. Let me see if I can get that done today.

xkwangcn:
I made a basic test with this change. 
Using terraform apply, it could successfully create a VM in my env for both local and remote libvirtd instance.

crawford:
@MalloZup I wasn't able to come up with an example that could demonstrate this functionality (the existing examples should work), but I did add a bit of documentation to explain how it works and the dependencies. It looks like @xkwangcn was successful with testing, so I think you can merge this once you are happy with it.

MalloZup:
@crawford thx ok from my side I will merge this end of the day thx sofar!

MalloZup:
@crawford @xkwangcn thx for this pr . Merged. Let me know if I can help for other stuff. enjoy

PR #682, opened 2019-12-11
resourceLibvirtVolumeCreate: tolerate existing volumes

moio:
This PR addresses an issue we frequently encounter in our CI, which manifests when builds are interrupted halfway through (either manually or because of bugs) or when manual changes are done to the libvirt host.

Scenario is the following:
 - `main.tf` (actually, in general, any configuration input to Terraform) wants the creation of a certain volume
 - that volume (actually, its key) is not present in `terraform.tfstate` (or, in general, in the Terraform state). That could be because such a volume was never created by Terraform at that point, or because the Terraform state was wiped for any reason
 - actually, though, the volume does exist

My understanding is that:
 - in this circumstance, Terraform will not check for the volume existence (eg. it will not call `resourceLibvirtVolumeExists` on it)
 - rather, it will just just ignore it's there and, at `terraform apply` time, it will call `resourceLibvirtVolumeCreate` directly
 - the following error is produced:

```
Error: Error creating libvirt volume: virError(Code=90, Domain=18, Message='storage volume 'reproducer' exists already')
```

In general, I can see three possible solutions:

1- abort Terraform and signal the error (current behavior). Our problem: this does not self-heal at the next run of the CI (currently we have to clean up the situation manually)
2- destroy the volume and recreate it. This is a clean solution, the downside being that the destruction would not be visible at `terraform plan` time. Users might have valuable data in volumes destroyed this way, and will not know until it's too late
3- tolerate the presence of the volume and just use it as-is. Chances are some attributes, such as the size, will not match the user expectation until the next round of `apply`. **This is proposed in this PR**

I am not aware of a way to detect this at `terraform plan` time, which would be the cleanest solution if it were possible.

### DIFF
--- a/libvirt/resource_libvirt_volume.go
+++ b/libvirt/resource_libvirt_volume.go
@@ -231,7 +231,16 @@ func resourceLibvirtVolumeCreate(d *schema.ResourceData, meta interface{}) error
 	// create the volume
 	volume, err := pool.StorageVolCreateXML(data, 0)
 	if err != nil {
-		return fmt.Errorf("Error creating libvirt volume: %s", err)
+		virErr := err.(libvirt.Error)
+		if virErr.Code != libvirt.ERR_STORAGE_VOL_EXIST {
+			return fmt.Errorf("Error creating libvirt volume: %s", err)
+		}
+		// oops, volume exists already, read it and move on
+		volume, err = pool.LookupStorageVolByName(volumeDef.Name)
+		if err != nil {
+			return fmt.Errorf("Error looking up libvirt volume: %s", err)
+		}
+		log.Printf("[INFO] Volume about to be created was found and left as-is: %s", volumeDef.Name)
 	}
 	defer volume.Free()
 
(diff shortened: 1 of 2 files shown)


moio:
CC @rjmateus

MalloZup:
https://travis-ci.org/dmacvicar/terraform-provider-libvirt/builds/623686141?utm_source=github_status&utm_medium=notification

moio:
I'll take care of them tomorrow. They are expected as the behavior has changed FYI.

moio:
@MalloZup tests are now green at the condition #683 is merged first.

Please review that one first!

MalloZup:
@moio  look  good thx sofar for all the small improvements. I think we are getting more resilient with this. :rocket:

MalloZup:
@moio let me know if you need to rebase  or need to push other things. Otherwise to me is good to merge.

moio:
Rebased and merged, thanks!

Issue #616, opened 2019-07-24
Problem creating bridged network using existing bridge


muroj:
### Linux distribution

``` Ubuntu 18.04.2 LTS```

### Terraform version

```sh
Terraform v0.11.14
+ provider.libvirt (unversioned)
+ provider.template v2.1.2

```

### Provider and libvirt versions

```sh
./terraform-provider-libvirt 0.5.2
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
virsh --version
5.1.0
```

___

## Checklist

- [ ] Is your issue/contribution related with enabling some setting/option exposed by libvirt that the plugin does not yet support, or requires changing/extending the provider terraform schema?

  - [ ] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.
  - [ ] If it is a very special case, consider using the _XSLT_ support in the provider to tweak the definition instead of opening an issue
  - [ ] Maintainers do not have expertise in every libvirt setting, so please, describe the feature and how it is used. Link to the appropriate documentation

- [x] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

## Description of Issue/Question

I am unable to create a terraform managed bridged network using an existing bridge. This is same issue as #364 which was fixed by #531. 

### Setup

main.tf
```
provider "libvirt" {
    uri = "qemu+ssh://${var.kvm_host_user}@${var.kvm_hostname}/system"
}

resource "libvirt_network" "ovs_bridge_network" {
  name = "ovs-vlan${var.vlan_id}-net"
  mode = "bridge"
  bridge = "${var.ovs_bridge_name}"
  autostart = false

  xml {
    xslt = "${data.template_file.domain_xml_override.rendered}"
  }
}

data "template_file" "domain_xml_override" {
  template = "${file("${path.module}/ovsvlan.xsl")}"

  vars = {
      vlan_id = "${var.vlan_id}"
      bridge_name = "${var.ovs_bridge_name}"
  }
}
```

variables.tf
```
variable "kvm_hostname" {
    description = "Hostname or IP of the provisioning system"
    default = "myhost"
}

variable "kvm_host_user" {
    description = "User for the provisioning system"
    default = "myuser"
}

variable "ovs_bridge_name" {
    description = "Name of the Open vSwitch bridge device on the host"
    default = "ovsbr0"
}

variable "vlan_id" {
    description = "VLAN_ID associated with this network"
[...]

Issue #443, opened 2018-10-11
bridge to lan not working


lifeofguenter:
It is currently not possible to set up a simple bridge-to-lan interface as described here: https://libvirt.org/formatdomain.html#elementsNICSBridge

Even if adding the bit manually in (via `virsh edit VM`) tf plan will not detect it:

```
  ~ libvirt_domain.jenkinsn
      network_interface.#:             "0" => "1"
      network_interface.0.addresses.#: "" => <computed>
      network_interface.0.bridge:      "" => "br0"
      network_interface.0.mac:         "" => "AA:BB:CC:11:21:94"
```

tf:

```
resource "libvirt_domain" "jenkinsn" {
  name      = "jenkinsn"
  autostart = true

  cpu {
    mode = "host-passthrough"
  }
  vcpu    = 4
  memory  = 8192

  disk {
    volume_id = "${libvirt_volume.jenkinsn_vol_root.id}"
    scsi      = true
  }

  disk {
    file = "/home/vbox/isos/debian-9.5.0-amd64-netinst.iso"
  }

  boot_device {
    dev = [ "cdrom", "hd"]
  }

  graphics {
    type        = "vnc"
    listen_type = "address"
  }

  network_interface {
    bridge = "br0"
    mac = "AA:BB:CC:11:21:94"
  }
}

resource "libvirt_volume" "jenkinsn_vol_root" {
  name   = "jenkinsn_vol_root"
  size   = "${300 * 1000 * 1000 * 1000}"
  format = "qcow2"
}
```

MalloZup:
@lifeofguenter hi can you add plese full `hcl` and erro msg would help! thx!

brd:
I have an existing bridge, `br0`, that it refuses to bridge to because it already exists.. Config:

```
resource "libvirt_network" "br0" {
  name = "br0"
  mode = "bridge"
  bridge = "br0"
}

# Create the machine
resource "libvirt_domain" "domain-ubuntu" {
  name   = "ubuntu-terraform"
  memory = "512"
  vcpu   = 1

  cloudinit = "${libvirt_cloudinit_disk.commoninit.id}"

  network_interface {
    network_name = "br0"
  }
```

And error:

```
libvirt_network.br0: Creating...                                  
  bridge: "" => "br0"                                              
  mode:   "" => "bridge"                                          
  name:   "" => "br0"
```

Followed by:

```
Error: Error applying plan:

2 error(s) occurred:

[...]

Issue #401, opened 2018-09-06
Unable to attach to existing bridge/and use local file


driftavalli:
#  Version Reports:

### Distro  version of host:

``` Ubuntu 18.04.1 ```

### Terraform Version Report
Terraform v0.11.8
+ provider.libvirt (unversioned)

### Libvirt version

```4.0.0```

### terraform-provider-libvirt plugin version (git-hash)

``` Downloaded precompiled binary: terraform-provider-libvirt-0.4.3.Ubuntu_18.04.amd64.tar.gz```
___
# Description of Issue/Question
1. Terraform hangs when the source of the volume is specified in the format ```"${file("~/tmp/bionic...")}"```
2. Get an error saying default already exists. From the [libvirt documentation](https://libvirt.org/formatnetwork.html#examplesBridge), it looks like it should be possible to use an existing bridge
### Setup
```
provider "libvirt" {
  uri         = "qemu:///system"
}

resource "libvirt_volume" "ubuntu-qcow2" {
  name        = "ubuntu-qcow2"
  pool        = "default"
  source      = "https://cloud-images.ubuntu.com/releases/xenial/release/ubuntu-16.04-server-cloudimg-amd64-disk1.img" //"${file("~/tmp/bionic-server-cloudimg-amd64.img")}"
  format      = "qcow2"
}

resource "libvirt_network" "default" {
  name        = "default"
  mode        = "bridge"
  bridge      = "br0"
}

resource "libvirt_cloudinit" "commoninit" {
  name        = "commoninit.iso"
  ssh_authorized_key  = "ssh-rsa ..."
}

resource "libvirt_domain" "domain-ubuntu" {
  name        = "ubuntu-terraform"
  memory      = "512"
  vcpu        = 1

  cloudinit   = "${libvirt_cloudinit.commoninit.id}"

  network_interface {
    network_name  = "default"
  }

  console {
    type        = "pty"
    target_port = "0"
    target_type = "serial"
  }

  console {
      type        = "pty"
      target_type = "virtio"
      target_port = "1"
  }

  disk {
       volume_id = "${libvirt_volume.ubuntu-qcow2.id}"
  }
  graphics {
    type = "spice"
    listen_type = "address"
    autoport = true
  }
}
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
terraform apply

___
# Additional Infos:

Do you have SELinux or Apparmor/Firewall enabled? Some special configuration?
[...]

Issue #364, opened 2018-08-21
libvirt_network creation fails for existing bridge


vjdhama:
#  Version Reports:

### Distro  version of host:

``` Ubuntu 18.04 ```

### Terraform Version Report

```
Terraform v0.11.8
+ provider.libvirt (unversioned)
```

### Libvirt version

```
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
Using API: QEMU 4.0.0
Running hypervisor: QEMU 2.11.1
```

### terraform-provider-libvirt plugin version (git-hash)

```
0.4.2
```
___
# Description of Issue/Question

By default virsh allows creation on a libvirt network on existing host bridge. 
https://libvirt.org/formatnetwork.html#examplesBridge

That works if you create libvirt network with virsh.

Using terraform it throws 

```
* libvirt_network.default: Error crearing libvirt network: virError(Code=38, Domain=0, Message='error creating bridge interface virbr0: File exists')
```

But the new network is created anyway.

```
<network>
  <name>default</name>
  <uuid>89d03309-b412-43f5-81bc-82b46bf81ec8</uuid>
  <bridge name='virbr0' stp='on' delay='0'/>
  <mac address='52:54:00:0e:cd:ba'/>
</network>
```

Also rerunning `terraform apply` fails with

```
2018-08-21T15:23:36.825+0700 [DEBUG] plugin.terraform-provider-libvirt: 2018/08/21 15:23:36 [ERR] plugin: stream copy 'stderr' error: session shutdown
* libvirt_network.default: Error defining libvirt network: virError(Code=9, Domain=19, Message='operation failed: network 'default' already exists with uuid 89d03309-b412-43f5-81bc-82b46bf81ec8') -   <network>
      <name>default</name>
      <bridge name="virbr0" stp="on"></bridge>
      <domain></domain>
  </network>
```

since the original run fails to collect metadata, id of network.

### Setup
(Please provide the full **main.tf** file for reproducing the issue (Be sure to remove sensitive info)

```
provider "libvirt" {
  uri = "qemu+tcp://root@172.16.255.254/system"
}

resource "libvirt_network" "default" {
  name = "default"
  mode = "bridge"
  bridge = "virbr0"
}
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

terraform init
terraform plan

```libvirt_network.default: Creating...
  bridge: "" => "virbr0"
  mode:   "" => "bridge"
[...]

PR #685, opened 2019-12-18
allow creating a bridged network

goraxe:
There were a couple of 'restrictions' which prevented virtual networks in bridge mode from being created.

1) the forward mode was removed from the xml (yes bridged networks can forward).  This caused libvirt to allocate a new bridge vs use the existing provided bridge.  Resulting in 'file exists' error messages.
2) STP was forced on (this is actually kinda dangerous.  Spanning tree protocol, should only be enabled when the operator/owner expressly asks for it. You can force large networks into learning mode crippling the network, or have a core switch turn off your top of rack uplink port and then you have to go buy your local network admin a bottle of scotch to get it turned back on... yep this is the voice of experience).   In bridged mode setting STP is not allowed.

https://libvirt.org/formatnetwork.html#examplesBridge

Fixes: #616 #443 #401 #364 

### DIFF
--- a/libvirt/helpers_test.go
+++ b/libvirt/helpers_test.go
@@ -249,7 +249,7 @@ func testAccCheckLibvirtNetworkBridge(resourceName string, bridgeName string) re
 			return fmt.Errorf("Bridge type of network should be not nil")
 		}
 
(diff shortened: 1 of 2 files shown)


MalloZup:
thx @goraxe

PR #692, opened 2020-01-25
Add support for decription field

torgiren:
Add support for setting description field in domains

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -45,6 +45,11 @@ func resourceLibvirtDomain() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
 			"metadata": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -456,6 +461,7 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 	domainDef.VCPU = &libvirtxml.DomainVCPU{
 		Value: d.Get("vcpu").(int),
 	}
+	domainDef.Description = d.Get("description").(string)
 
 	domainDef.OS.Kernel = d.Get("kernel").(string)
 	domainDef.OS.Initrd = d.Get("initrd").(string)
@@ -741,6 +747,7 @@ func resourceLibvirtDomainRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.Set("name", domainDef.Name)
+	d.Set("description", domainDef.Description)
 	d.Set("vcpu", domainDef.VCPU)
 	d.Set("memory", domainDef.Memory)
 	d.Set("firmware", domainDef.OS.Loader)
--- a/libvirt/resource_libvirt_domain_test.go
+++ b/libvirt/resource_libvirt_domain_test.go
@@ -66,10 +66,6 @@ func TestAccLibvirtDomain_Description(t *testing.T) {
 					resource.TestCheckResourceAttr(
 						"libvirt_domain."+randomResourceName, "name", randomDomainName),
 					resource.TestCheckResourceAttr(
-						"libvirt_domain."+randomResourceName, "memory", "512"),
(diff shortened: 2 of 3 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -45,6 +45,11 @@ func resourceLibvirtDomain() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
```
MalloZup: I think we should not use `forceNew` to true , instead should be false.

Is there any particular motivation for setting it to true? 

tia

MalloZup:
@torgiren thx for PR.  

2 things:

1) I would like to have tests for this new field. You can have look on howtoContribute there is a guide for tests.

2) see inline comments

Thx for PR!

libvirt/resource_libvirt_domain.go
torgiren: I couldn't find any way to update description on fly.
Without that, terraform will complain about missmatch description every run

libvirt/resource_libvirt_domain_test.go
```diff
@@ -58,9 +58,10 @@ func TestAccLibvirtDomain_Detailed(t *testing.T) {
 			{
 				Config: fmt.Sprintf(`
 				resource "libvirt_domain" "%s" {
-					name   = "%s"
-					memory = 384
-					vcpu   = 2
+					name        = "%s"
```
MalloZup: I would add here a seperate testcase for granularity and separation of concern

libvirt/resource_libvirt_domain.go
MalloZup: ok

MalloZup:
thx a part minor fixes looks good

libvirt/resource_libvirt_domain_test.go
torgiren: ok, i will split this test in a few days :)

MalloZup: thx!

MalloZup: just ping me when done

torgiren: I extracted Description check into separate test.
The travis failed, but I think that my check has passed and it failed later on some storage test - maybe some random fail?

MalloZup: let me check

website/docs/r/domain.html.markdown
```diff
@@ -25,6 +25,8 @@ The following arguments are supported:
 
 * `name` - (Required) A unique name for the resource, required by libvirt.
   Changing this forces a new resource to be created.
+* `description` - (Optional) The description for domain.
```
MalloZup: last nitpick. From User experience point, I would like that we state why this could be useful.

E.g why one should add the description?

We should state also `(Optional) The description for domain. Useful in case of XX for XX bla`

MalloZup:
@torgiren last nitpick and I will merge. Thank you a lot for your efforts sofar :rocket:

website/docs/r/domain.html.markdown
torgiren: To be honest... The description field does not have any technical purpose, it's just notes, so it's very hard for me to find use cases...

MalloZup: why do you want to add it then? :thinking:

torgiren: I use this field for dynamic inventory for ansible. But this is my custom script and no one else use it... The libvirt documentation https://libvirt.org/formatdomain.html#elementsMetadata says that:
```
 description
    The content of the description element provides a human readable description of the virtual machine.
    This data is not used by libvirt in any way, it can contain any information the user wants. 
```

MalloZup: ok . I think we can paste and copy libvirt definition

```    The content of the description element provides a human readable description of the virtual machine.
    This data is not used by libvirt in any way, it can contain any information the user wants. 
```

torgiren: done

PR #698, opened 2020-01-30
Vnc port conf

nutcr4cker:
Hi, 
for me it's really helpful to have control over port/password for VNC, generating and defining this before the start of domain helping with automatically send it via email to users who will need access so they will be able to install OS themselves

maybe it will valuable to someone also...

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -441,6 +441,8 @@ func setGraphics(d *schema.ResourceData, domainDef *libvirtxml.Domain, arch stri
 		}
 
 		autoport := d.Get(prefix + ".autoport").(bool)
+		port := d.Get(prefix + ".port").(int)
+		passwd := d.Get(prefix + ".passwd").(string)
 		listener := libvirtxml.DomainGraphicListener{}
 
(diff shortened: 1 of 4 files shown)


libvirt/resource_libvirt_domain.go
```diff
@@ -236,6 +236,14 @@ func resourceLibvirtDomain() *schema.Resource {
 							Optional: true,
 							Default:  "spice",
 						},
+						"port": {
+							Type:     schema.TypeInt,
+							Optional: true,
+						},
+						"passwd": {
```
MalloZup: I would add here `password` instead of this

libvirt/domain.go
```diff
@@ -473,6 +475,14 @@ func setGraphics(d *schema.ResourceData, domainDef *libvirtxml.Domain, arch stri
 				VNC: &libvirtxml.DomainGraphicVNC{},
 			}
 			domainDef.Devices.Graphics[0].VNC.AutoPort = formatBoolYesNo(autoport)
+			if autoport == false {
```
MalloZup: a comment of this would be good. e.g would be good to have a comment what `-1` means in case of autoport

PR #714, opened 2020-03-05
Set hostname for dhcp networks

### LINKED ISSUE #708: hostname is not set if dhcp is used for assigning addresses
### Linux distribution

``` opensuse leap 15 sp1```

### Terraform version
```
Terraform v0.12.19
+ provider.libvirt (unversioned
```

```
terraform-provider-libvirt was not built correctly
Compiled against library: libvirt 5.1.0
Using library: libvirt 5.1.0
```
Note:  the provider was installed from package `0.6.1+git.1580485189.cc55d5d4-22.1` 

## Checklist

- [X] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

When assigning ip addresses dynamically from a nat network, the hostname specified in the network interface is not respected. If the address is assigned statically, it is respected.

### Setup
```
provider "libvirt" {
  uri = "qemu://system"
}

resource "libvirt_network" "network" {
  name   = "test-network"
  mode   = "nat"
  domain = "test.local"

  dns {
    enabled = true
  }

  addresses = ["10.17.0.0/22"]
}

resource "libvirt_volume" "img" {
  name   = "SLE-15-SP1-JeOS-GMC.qcow2"
  source = "path/to/image"
  pool   = "default"
}
resource "libvirt_volume" "master" {
  name           = "master-volume"
  pool           = "default"
  size           = "25769803776"
  base_volume_id = libvirt_volume.img.id
}

resource "libvirt_domain" "master" {
  name       = "master-domain"

  disk {
    volume_id = libvirt_volume.master.id
  }

  network_interface {
    network_id     = libvirt_network.network.id
    hostname       = "master"
    wait_for_lease = true
  }

  graphics {
    type        = "vnc"
    listen_type = "address"
  }
}
```

### Steps to Reproduce Issue

Apply the configuration as above and the resulting `libvirt_domain` resource will have the  network interface defined as shown below:
```
            "network_interface": [
              {
                "addresses": [
                  "10.17.0.198"
                ],
                "hostname": "",
```

If the configuration is changed to add a fixed ip address, 
```
  network_interface {
    network_id     = libvirt_network.network.id
    hostname      =  "master"
   addresses      =  [ "10.17.1.20"] 
   wait_for_lease = true
  }
```

the hostname is respected.

```
            "network_interface": [
              {
                "addresses": [
                  "10.17.1.20"
                ],
                "hostname": "master",
```

___
## Additional information:

### PR DESCRIPTION
**Pull request by pablochacin**
For dhcp networks, ensure the hostname is set when the ip address is allocated.

Unify behavior when network is referenced either by name or by id.

Fix bug using network after it was freed

Improve test for covering new behavior

Fixes #708 

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -706,23 +706,58 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
 		}
 
 		// connect to the interface to the network... first, look for the network
-		if n, ok := d.GetOk(prefix + ".network_name"); ok {
-			// when using a "network_name" we do not try to do anything: we just
-			// connect to that network
-			netIface.Source = &libvirtxml.DomainInterfaceSource{
-				Network: &libvirtxml.DomainInterfaceSourceNetwork{
-					Network: n.(string),
-				},
+		var network *libvirt.Network
+		var err error
+
+		if networkName, ok := d.GetOk(prefix + ".network_name"); ok {
+			network, err = virConn.LookupNetworkByName(networkName.(string))
+			if err != nil {
+				return fmt.Errorf("Can't retrieve network '%s'", networkName.(string))
 			}
+			defer network.Free()
+
 		} else if networkUUID, ok := d.GetOk(prefix + ".network_id"); ok {
 			// when using a "network_id" we are referring to a "network resource"
 			// we have defined somewhere else...
-			network, err := virConn.LookupNetworkByUUIDString(networkUUID.(string))
+			network, err = virConn.LookupNetworkByUUIDString(networkUUID.(string))
 			if err != nil {
 				return fmt.Errorf("Can't retrieve network ID %s", networkUUID)
 			}
 			defer network.Free()
 
+		} else if bridgeNameI, ok := d.GetOk(prefix + ".bridge"); ok {
+			netIface.Source = &libvirtxml.DomainInterfaceSource{
+				Bridge: &libvirtxml.DomainInterfaceSourceBridge{
+					Bridge: bridgeNameI.(string),
+				},
+			}
+		} else if devI, ok := d.GetOk(prefix + ".vepa"); ok {
+			netIface.Source = &libvirtxml.DomainInterfaceSource{
+				Direct: &libvirtxml.DomainInterfaceSourceDirect{
+					Dev:  devI.(string),
+					Mode: "vepa",
+				},
+			}
+		} else if devI, ok := d.GetOk(prefix + ".macvtap"); ok {
+			netIface.Source = &libvirtxml.DomainInterfaceSource{
+				Direct: &libvirtxml.DomainInterfaceSourceDirect{
+					Dev:  devI.(string),
+					Mode: "bridge",
+				},
+			}
+		} else if devI, ok := d.GetOk(prefix + ".passthrough"); ok {
+			netIface.Source = &libvirtxml.DomainInterfaceSource{
+				Direct: &libvirtxml.DomainInterfaceSourceDirect{
+					Dev:  devI.(string),
+					Mode: "passthrough",
+				},
+			}
+		} else {
+			// no network has been specified: we are on our own
+		}
+
+		// if we got a network
+		if network != nil {
 			networkName, err := network.GetName()
 			if err != nil {
 				return fmt.Errorf("Error retrieving network name: %s", err)
@@ -763,10 +798,13 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
 						// have a valid lease and then read the IP we have been assigned, so we can
 						// do the mapping
 						log.Printf("[DEBUG] Do not have an IP for '%s' yet: will wait until DHCP provides one...", hostname)
+						if err != nil {
+							return err
+						}
 						partialNetIfaces[strings.ToUpper(mac)] = &pendingMapping{
-							mac:      strings.ToUpper(mac),
-							hostname: hostname,
-							network:  network,
+							mac:         strings.ToUpper(mac),
+							hostname:    hostname,
+							networkName: networkName,
 						}
 					}
 				}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: libvirt/domain.go [hunk 1]
```diff
@@ -706,23 +706,59 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
...
+			}
+		} else {
+			// no network has been specified: we are on our own
+		}
+
+		// if we got a network
+		if network != nil {
 			networkName, err := network.GetName()
 			if err != nil {
 				return fmt.Errorf("Error retrieving network name: %s", err)
```
flavio: Before the `return` statement:

```golang
defer network.Free()
```

The `network` instance has not been added to the `partialNetIFace` map, hence we can safely release the object right now. There's no risk of it being free-ed up again as a mistake.

Maybe I would add these lines as comments inside of the block.

libvirt/domain.go
```diff
@@ -763,6 +799,10 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
 						// have a valid lease and then read the IP we have been assigned, so we can
 						// do the mapping
 						log.Printf("[DEBUG] Do not have an IP for '%s' yet: will wait until DHCP provides one...", hostname)
+						err = network.Ref()
+						if err != nil {
+							return err
+						}
```
flavio: I prefer to leave the task of freeing the `network` instance to whoever is going to process `partialNetIfaces`.

GitHub doesn't allow me to add comments to the lines above because they are not affected by the change: there are two more `return` statement, we have to introduce a `defer network.Free()` before them.

libvirt/resource_libvirt_domain.go
```diff
@@ -599,7 +599,8 @@ func resourceLibvirtDomainCreate(d *schema.ResourceData, meta interface{}) error
 			for _, addressI := range addressesI.([]interface{}) {
 				address := addressI.(string)
 				log.Printf("[INFO] Finally adding IP/MAC/host=%s/%s/%s", address, mac, pending.hostname)
-				updateOrAddHost(pending.network, address, mac, pending.hostname)
+				err := updateOrAddHost(pending.network, address, mac, pending.hostname)
+				pending.network.Free()
 				if err != nil {
 					return fmt.Errorf("Could not add IP/MAC/host=%s/%s/%s: %s", address, mac, pending.hostname, err)
 				}
```
flavio: Good catch, `err` wasn't simply associated before :man_facepalming: 

I would refactor the code in this way:

```golang
if err := updateOrAddHost(pending.network, address, mac, pending.hostname); err != nil {
   // log the message somehow, do not return with an error - don' return the `error` object
}
if err := network.Free(); err != nil {
  // log error message, move forward
}
```

I would not return in the middle of this for loop because that would prevent to process other pending addresses if one of them fails for whatever reason. Exiting from the loop would also prevent the de-allocation of the objects.

libvirt/domain.go
pablochacin: I don't see the advantage of doing this. The idea of using the `Ref()` call is to increment the reference counter only if the network is added to partialNetIFace, is that in this way, it is not necessary to consider carefully all the cases when it can be freed. It can always be freed. I think the logic is more straightforward in this way.

pablochacin:
> * Please do not conflate all the changes into a single commit. You're solving different problems with the same PR (which is acceptable), but you should have multiple commits (eg: the fixes about the `network` resource being released twice should be part of a different commit)

I tend to do this except that in this case, without this change, the refactoring for setting the hostname doesn't work. The test fails. I generally don't like to make commits that don't work.

On the other hand, the way the reference to the network is freed depends on the re-organization of the code introduced. I could make a fix for that under the actual code structure, but that change will be completely lost in the second commit.

MalloZup:
thx @flavio and @pablochacin  for this. i will have a look to this PR also next week

libvirt/domain.go
pablochacin: I'm confused by this comment. This code does precisely that: leaves the responsibility of freeing the network to the one processing partialNetIfaces. The use of the `Ref()`, as explained in another comment, is to simplify the logic: always free, but increment the reference only in the case the network is added to `partialNetIfaces`.

libvirt/domain.go
pablochacin: Additionally, if there is more than one interface on the same network that must be added to `partialNetIface`, it is much easier to keep track of the usage of the network using `Ref`to increment the reference count and `Free` it independently for each interface.

flavio:
To elaborate more on why I'm not so convinced about using the `Ref()` method.

When dealing with resource cleanup a Go developer doesn't have a double garbage collector running like in this case, where we have Go's GC and the one implemented inside of libvirt.

When dealing with resources like a database object a Go developer has still to take care of invoking some finalization code before Go's GC takes the object instance out of reach. The developer does that by leveraging the `defer` construct and by taking a close look at how `return` statements are used.

I would prefer to ignore the `Ref()` method and take a more "idiomatic" approach to this problem. I would also, with a later series of PRs, refactor this portion of code to make it more manageable. Right now I admit that understanding when to do a `defer`/`Free`/nothing at all is complicated, but that's caused by the incredible length and branching we have in this area of the codebase.

pablochacin:
@flavio I gave a try to your proposal and the resulting code is extremely convoluted and hard to maintain because it must ensure it will work on every possible code execution path. It is a lot of changes for something that we expect to solve by refactoring the code.

One alternative could be to change the `pendingMapping` structure and pass the network name or ID instead of the reference to the network and then whoever takes care of these pending mappings will obtain and release the network reference from this name or ID. I think this is trade-off between idiomatic solution and refactoring the code (it is a minor change in the code)

If you disagree with this alternative, then do you want to include this refactoring in the present PR? I would prefer not, but this is the only alternative to introducing multiple Free in the code, I prefer to refactor the code.

flavio:
Let's go ahead with the `Ref()` approach for now. We can try to get rid of them with a future PR

libvirt/resource_libvirt_domain_test.go
```diff
@@ -603,31 +603,45 @@ func TestAccLibvirtDomain_NetworkInterface(t *testing.T) {
...
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
 		Providers:    testAccProviders,
 		CheckDestroy: testAccCheckLibvirtDomainDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: config,
+				Config:             config,
+				ExpectNonEmptyPlan: false, // TODO: find why if set to true, the test fails
```
MalloZup: we might need to check this before mergin because it could be a code smell of regression.

same also the `Destroy` I think we never used in Testacc

MalloZup:
@pablochacin @flavio  regarding the Ref or not Ref approach.

I'm fine with both approaches, BUT please add some comment on the design.

I remember to have looked in the past how we process networks and until you don't dive in actively in the code, you can't understand ( the code is also pretty complicated for good and bad reasons).

So if we can have comments explaining how we use the REF and why, this will be appreciate, because I have already forgotten that part of code. :grin: 


A part of this,  I think we might check 2 times the testacc because imho we are modifying them to `pass` in an usual way and this could be there is a regression since we need to modify them to make them pass see my comments.

Thx sofar.. :sun_with_face:

libvirt/resource_libvirt_domain_test.go
pablochacin: I agree. I was unable to identify the reason, except that before the plan had the hostname but after the apply, it was empty. Terraform complained about this. I pointed this out in the issue https://github.com/dmacvicar/terraform-provider-libvirt/issues/708#issuecomment-592489392

pablochacin:
I have updated the PR. I think I've addressed all the concerns except the change in the test, as @MalloZup pointed out. 

I finally changed the code not to pass a reference to the network in the structure of the pending interface, but the network name. I think this way we delegate to the routine that handles the pending interfaces when tp acquire an release the network reference.

libvirt/resource_libvirt_domain_test.go
pablochacin: https://www.terraform.io/docs/extend/best-practices/testing.html#expecting-errors-or-non-empty-plans

flavio:
@MalloZup I would be fine approving this PR. What do you think?

Also, once merged, how can we tag a new release including this fix?

MalloZup:
@flavio  ok I will prepare a new tag including this fix. Then we can build the RPM . @dmacvicar  is ok for you at this point to do a new release? I think we have a release draft but is more a sketch. I will need to go trough the PR merged and summarize them.

Sofar thx lot for helping and contributing. Looking forward :sunflower:  :guitar:

MalloZup:
see  https://github.com/dmacvicar/terraform-provider-libvirt/releases/tag/untagged-30bccf4e7c1526ff1438 but is not complete.

If you want to help feel free @flavio , otherwise I will do it during the day.

@pablochacin @flavio  if you want a RPM package from `master` you could use the `unstable` repo just trigger it from the service.

If you wanna right there feel free to ask via obs I will accept.

https://build.opensuse.org/package/show/systemsmanagement:terraform:unstable/terraform-provider-libvirt

pablochacin:
Thanks a lot, @MalloZup. Not sure what can/should/must do here.

MalloZup:
we need to release a new version. I will make it happen

PR #718, opened 2020-03-18
Use virtio-blk instead of disk for injecting ignition config for s390x and ppc64

Prashanth684:
The method of mimicking what Openstack does for injecting ignition config works for
images which have the provider as Openstack because ignition recognizes the platform
and knows it has to get the ignition config from the config drive. For QEMU images, ignition
supports getting the config from the firmware config device which is not supported by ppc64
and s390x.

The workaround we have used thus far is to use the Openstack image on the QEMU platform but have this provider create the iso containing the ignition config. There was a discussion in ignition
(https://github.com/coreos/ignition/issues/928) to have a more QEMU based method of injecting ignition config and it was decided to use a virtio-blk device with a serial of `ignition` which ignition can recognize. This was mainly because with external devices, it is hard to tell if there is an issue with the device or if the kernel has not detected it yet if it has a long discovery phase.

This PR reverts the ISO method used by Openstack and just creates a virtio-blk device through the QEMU command line options.

Reference PR which supports ignition fetching through virtio-blk for QEMU: https://github.com/coreos/ignition/pull/936


libvirt/domain.go
```diff
@@ -270,17 +258,28 @@ func setCoreOSIgnition(d *schema.ResourceData, domainDef *libvirtxml.Domain, vir
...
-			if err != nil {
-				return err
+			// System Z and PowerPC do not support the Firmware Configuration
+			// device. After a discussion about the best way to support a similar
+			// method for qemu in https://github.com/coreos/ignition/issues/928,
+			// decided on creating a virtio-blk device with a serial of ignition
+			// which contains the ignition config and have ignition support for
+			// reading from the device which landed in https://github.com/coreos/ignition/pull/936
+			domainDef.QEMUCommandline = &libvirtxml.DomainQEMUCommandline{
+				Args: []libvirtxml.DomainQEMUCommandlineArg{
```
jlebon: Hmm, not familiar with these golang libvirt bindings, though it looks like we can use [DomainDisk](https://godoc.org/github.com/libvirt/libvirt-go-xml#DomainDisk) and set the Serial member there.

Prashanth684: done.

MalloZup:
@crawford  I will wait one/2 days if you wanna review otherwise will merge thx

MalloZup:
thx. it will come to next release

PR #722, opened 2020-03-24
Fix: block device test fails searching for loop dev

pablochacin:
Test TestAccLibvirtDomain_BlockDevice fails searching for available loop device because the command /sbin/losetup --find is not executed as root.


MalloZup:
thx

PR #725, opened 2020-04-05
Add dev attribute to network configuration

Xachman:
Fixes https://github.com/dmacvicar/terraform-provider-libvirt/issues/720

This adds the dev attribute if selected to the <forward /> tag in the network xml

### DIFF
--- a/libvirt/network.go
+++ b/libvirt/network.go
@@ -43,6 +43,9 @@ func getNetModeFromResource(d *schema.ResourceData) string {
 	return strings.ToLower(d.Get("mode").(string))
 }
 
+func getNetDevFromResource(d *schema.ResourceData) string {
+	return strings.ToLower(d.Get("dev").(string))
+}
 // getIPsFromResource gets the IPs configurations from the resource definition
 func getIPsFromResource(d *schema.ResourceData) ([]libvirtxml.NetworkIP, error) {
 	addresses, ok := d.GetOk("addresses")
--- a/libvirt/network_def_test.go
+++ b/libvirt/network_def_test.go
@@ -32,7 +32,7 @@ func TestNetworkDefUnmarshall(t *testing.T) {
 			<bridge name="virbr0" stp="on" delay="5" macTableManager="libvirt"/>
 			<mac address='00:16:3E:5D:C7:9E'/>
 			<domain name="example.com" localOnly="no"/>
-			<forward mode='nat'>
+			<forward mode='nat' dev='eth0'>
 				<nat>
 					<address start='1.2.3.4' end='1.2.3.10'/>
 				</nat>
@@ -71,6 +71,9 @@ func TestNetworkDefUnmarshall(t *testing.T) {
 	}
 	if b.Forward.Mode != "nat" {
 		t.Errorf("wrong forward mode: '%s'", b.Forward.Mode)
+	}
+	if b.Forward.Dev != "eth0" {
+		t.Errorf("wrong forward dev: '%s'", b.Forward.Dev)
 	}
 	if len(b.Forward.NAT.Addresses) == 0 {
 		t.Errorf("wrong number of addresses: %s", b.Forward.NAT.Addresses)
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -62,6 +62,11 @@ func resourceLibvirtNetwork() *schema.Resource {
 				ForceNew: true,
 				Default:  netModeNat,
 			},
+			"dev": { // device for forward
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: false,
+			},
 			"bridge": {
 				Type:     schema.TypeString,
 				Optional: true,
@@ -382,6 +387,7 @@ func resourceLibvirtNetworkCreate(d *schema.ResourceData, meta interface{}) erro
 	// check the network mode
 	networkDef.Forward = &libvirtxml.NetworkForward{
 		Mode: getNetModeFromResource(d),
+		Dev: getNetDevFromResource(d),
 	}
 	if networkDef.Forward.Mode == netModeIsolated || networkDef.Forward.Mode == netModeNat || networkDef.Forward.Mode == netModeRoute {
 
--- a/website/docs/r/network.markdown
+++ b/website/docs/r/network.markdown
@@ -20,6 +20,10 @@ resource "libvirt_network" "kube_network" {
 
   # mode can be: "nat" (default), "none", "route", "bridge"
   mode = "nat"
+  
+  # (optional) the network device for forward
+  # should be the network device or bridge. prevents talk between nat networks
+  # dev = "eth0"
 
   #  the domain used by the DNS server in this network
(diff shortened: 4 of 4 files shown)


MalloZup:
looks good thx!

dmacvicar:
While preparing the release I had a look at this one, and I think exposing a functionality that is part of a _forward_ namespace on the main _network_ namespace is not a good idea.

For example, the `forward` element allows for multiple _devices_ to be specified. The way this PR is implemented will not allow to support that.
Any element of `network` that in the future specifies a _device_ for any purpose, will likely clash with this one as well.

I am sorry this happens after the PR is merged. As mentioned in [CONTRIBUTING.md](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/CONTRIBUTING.md), changes to the schema should create a discussion.

Xachman:
Alright the main reason I worked on this was to facilitate isolated networks.  This didn't seem to work and most of the information I read on the issue mentioned adding the dev attribute.   However if this is going to cause issues I'm more than willing to help sort that out.

dmacvicar:
I think the diagnostic was mislead. lbivirt docs mentions:

>If there is no forward element, the network will be isolated from any other network (unless a guest connected to that network is acting as a router, of course). 

But we always create a forward element.

```golang
	// check the network mode
	networkDef.Forward = &libvirtxml.NetworkForward{
		Mode: getNetModeFromResource(d),
		Dev:  getNetDevFromResource(d),
	}
```

The right fix would have been creating the forward element only if either _mode_ or _dev_ was specified. This would have been backwards incompatible. The original problem originates in my opinion from f6a859c2f, which added the _mode_ parameter, but set it as _forward_, _nat_ as default, while libvirt default to _nat_ only if _forward_ is specified (I haven't tested, but according to the docs). Putting the _mode_ attribute in the top level was IMHO, a mistake too.

I am preparing an alternative implementation. But I am decided to break backward compatibility here and default to no forward if _mode_ is not specified.

Xachman:
Adding the dev attribute does prevent cross talk because appropriate iptables rules are applied to prevent it.  If the forward element implementation should change I will fall back on your judgement on that.

dmacvicar:
I am not against the_dev_ attribute. My problem is that we are using _dev_ to prevent cross-talk because we can't create networks without _forward_ and without _nat_. That and the fact that we are polluting the _network_ tag with values specific to forward.

I understand the cross talk happens because even if you don't specify anything, the provider currently adds a _forward_, and by doing this, activates NAT and the host routing stack is the responsible for the cross-talk.

>nat
>    All traffic between guests connected to this network and the physical network will be forwarded to the physical network via the host's IP routing stack, after the guest's IP address is translated to appear as the host machine's public IP address (a.k.a. Network Address Translation, or "NAT"). This allows multiple guests, all having access to the physical network, on a host that is only allowed a single public IP address. If a network has any IPv6 addresses defined, the IPv6 traffic will be forwarded using plain routing, since IPv6 has no concept of NAT. Firewall rules will allow outbound connections to any other network device whether ethernet, wireless, dialup, or VPN. If the dev attribute is set, the firewall rules will restrict forwarding to the named device only. Inbound connections from other networks are all prohibited; all connections between guests on the same network, and to/from the host to the guests, are unrestricted and not NATed.Since 0.4.2

So I prefer to fix this by allowing both really isolated networks, no forward, and networks with forward & NAT which can be restricted to a device. I want to have semantics as close as libvirt as possible.

Xachman:
Keeping the semantics close to libvirt definitely makes sense and I agree with that networking could be reworked here.

dmacvicar:
And I was slightly wrong.

Just after creating always a _forward_ block, the code then does this:

```golang
if networkDef.Forward.Mode == netModeIsolated || networkDef.Forward.Mode == netModeNat || networkDef.Forward.Mode == netModeRoute {

		if networkDef.Forward.Mode == netModeIsolated {
			// there is no forwarding when using an isolated network
			networkDef.Forward = nil
		} else if networkDef.Forward.Mode == netModeRoute {
			// there is no NAT when using a routed network
			networkDef.Forward.NAT = nil
		}
```

Which actually then removes the _forward_ element. So you can have an isolated network without NAT by using _netModeIsolated_, which it is the string `"none"`. This is not even a valid option in libvirt.

It is documented on our side.

>  # mode can be: "nat" (default), "none", "route", "bridge"
 > mode = "nat"

did you try this one?

Xachman:
I think I did try this before.  This probably works but I'm interacting with ssh connections from the host who has no access when mode = none.  I could be something wrong tho if this should work.  I'll try some other setups maybe with vnc and see what the network looks like.

dmacvicar:
I am undecided what is better.

* Keeping `dev` at the top level as in this PR
  * As the whole _network_ part needs a revamp, this attribute would anyway need to disappear later
  * Perhaps less ugly if we call it `forward_dev` (as mode should have been `forward_mode`
* Add a _HCL_  `forward` block with `dev` and `mode`, akin to libvirt `<forward mode='...' dev='...'/>`
  * A bit ugly _internally_ as the code is doing too much magic based on the mode. But it can be done
  * Need to deprecate `mode`, and no longer have NAT as default. No `mode`, no `forward` unless there is a forward block. this may cause breakage
* Do nothing for now. Do not support this use-case for now. Advise to use the XSLT feature to add the device to the forward element.

@MalloZup what do you think?

It is the classic dilemma between mapping the XML 1:1 or deal with the consequences.

MalloZup:
thx @dmacvicar my fault I didn't checked this change more accurately..


I would think either use `XSLT` providing an example or using `forward_dev`.

But lately, I would tend to use the `XSLT` because if we expose the `forward_dev`,

we are considering only **creation** of resource ..

If **update** an existing network created with `forward_dev` this will be not updated ...

https://github.com/Xachman/terraform-provider-libvirt/blob/a4ba6b895176b0631e259a48698f89080a1c8dce/libvirt/resource_libvirt_network.go#L282

that is not something the PR changed but something we do have problematic and that was also  the origin of my discussion on last PR and issue with the `domain` parameter where user is expecting to change.


So imho here I would rather revert the PR providing `XSLT` or maybe using `forward_dev` but we should for each new XML check if we can `update` network and if not we should imho use `ForceNew=True` which will recreate network because we can't update it

MalloZup:
thx @dmacvicar my fault I didn't checked this change more accurately..


I would think either use `XSLT` providing an example or using `forward_dev`.

But lately, I would tend to use the `XSLT` because if we expose the `forward_dev`,

we are considering only **creation** of resource ..

If **update** an existing network created with `forward_dev` this will be not updated ...

https://github.com/Xachman/terraform-provider-libvirt/blob/a4ba6b895176b0631e259a48698f89080a1c8dce/libvirt/resource_libvirt_network.go#L282

that is not something the PR changed but something we do have problematic and that was also  the origin of my discussion on last PR and issue with the `domain` parameter where user is expecting to change.


So imho here I would rather revert the PR providing `XSLT` or maybe using `forward_dev` but we should for each new XML check if we can `update` network and if not we should imho use `ForceNew=True` which will recreate network because we can't update it

Issue #703, opened 2020-02-08
Create 2 KVM libvirt networks failed


magsoftware:
### Linux distribution

```
 Centos7
```

### Terraform version

```sh
Terraform v0.12.20
+ provider.libvirt (unversioned)
```

### Provider and libvirt versions

```sh
Compiled against library: libvirt 4.1.0
Using library: libvirt 4.5.0
Running hypervisor: QEMU 1.5.3
Running against daemon: 4.5.0
```
___

## Description of Issue/Question

It looks like it is not possible to define 2 networks.

### Setup

```
# Create a network for our VMs
resource "libvirt_network" "kube_network" {
  name = "k8s-net"
  mode = "nat"
  addresses = [ "10.0.1.0/24" ]
  dhcp {
    enabled = true
  }
  dns {
    enabled = true
    local_only = true
  }
}

#
# Output
#

output "kube_network" {
  value = libvirt_network.kube_network.name
}

# Create a network for our VMs
resource "libvirt_network" "ext_network" {
  name = "ext-net"
  mode = "nat"
  addresses = [ "10.10.1.0/24" ]
  dhcp {
    enabled = true
  }
  dns {
    enabled = true
    local_only = true
  }
}

#
# Output
#

output "ext_network" {
  value = libvirt_network.ext_network.name
}
```

### Steps to Reproduce Issue

```
# virsh net-list --all
 Name                 State      Autostart     Persistent
----------------------------------------------------------
 default              active     yes           yes

# terraform apply -auto-approve
libvirt_network.kube_network: Creating...
libvirt_network.ext_network: Creating...
libvirt_network.ext_network: Creation complete after 5s [id=278b4352-0843-4ff5-b737-05485d530851]

Error: Error clearing libvirt network: virError(Code=38, Domain=0, Message='error creating bridge interface virbr1: File exists')

  on network.tf line 2, in resource "libvirt_network" "kube_network":
   2: resource "libvirt_network" "kube_network" {


# virsh net-list --all
 Name                 State      Autostart     Persistent
----------------------------------------------------------
 default              active     yes           yes
 ext-net              active     no            yes
 k8s-net              inactive   no            yes
```
___
## Additional information:

SELinus in Permissive mode

If I create 2 different directories and split main.tf so that each main.tf contain only a single network definition than it works and 2 networks are created.

mateuszbanasiewicz:
Problem is connected with bridge interface which is created during creating network. 
[...]

PR #740, opened 2020-05-16
#703 - Permit creation of multiple libvirt networks

vmorris:
Fixes a bug where multiple libvirt networks create but only one would start due to bridge name collisions.

I wasn't sure how to include a test for this, but it is a fairly small change..

### DIFF
--- a/libvirt/network.go
+++ b/libvirt/network.go
@@ -46,6 +46,7 @@ func getNetModeFromResource(d *schema.ResourceData) string {
 func getNetDevFromResource(d *schema.ResourceData) string {
 	return strings.ToLower(d.Get("dev").(string))
 }
+
 // getIPsFromResource gets the IPs configurations from the resource definition
 func getIPsFromResource(d *schema.ResourceData) ([]libvirtxml.NetworkIP, error) {
 	addresses, ok := d.GetOk("addresses")
@@ -132,13 +133,18 @@ func getNetworkIPConfig(address string) (*libvirtxml.NetworkIP, *libvirtxml.Netw
 	return dni, dhcp, nil
 }
 
-// getBridgeFromResource returns a libvirt's NetworkBridge
-// from the ResourceData provided.
+// getBridgeFromResource returns a libvirt's NetworkBridge from the
+// ResourceData provided or creates a new one if not specified
+// based on the network name.
 func getBridgeFromResource(d *schema.ResourceData) *libvirtxml.NetworkBridge {
-	// use a bridge provided by the user, or create one otherwise (libvirt will assign on automatically when empty)
-	bridgeName := ""
+	// use a bridge provided by the user, or create one otherwise
+	log.Printf("VEMVEMVEMVEMVEMVEMVEMVEMVEMVEM")
+	var bridgeName string
 	if b, ok := d.GetOk("bridge"); ok {
 		bridgeName = b.(string)
+	} else {
+		netName, _ := d.GetOk("name")
+		bridgeName = netName.(string) + "-br"
 	}
 
 	bridge := &libvirtxml.NetworkBridge{
--- a/libvirt/network.go
+++ b/libvirt/network.go
@@ -138,7 +138,6 @@ func getNetworkIPConfig(address string) (*libvirtxml.NetworkIP, *libvirtxml.Netw
 // based on the network name.
 func getBridgeFromResource(d *schema.ResourceData) *libvirtxml.NetworkBridge {
 	// use a bridge provided by the user, or create one otherwise
-	log.Printf("VEMVEMVEMVEMVEMVEMVEMVEMVEMVEM")
 	var bridgeName string
 	if b, ok := d.GetOk("bridge"); ok {
 		bridgeName = b.(string)
--- a/libvirt/qemu_agent.go
+++ b/libvirt/qemu_agent.go
@@ -113,7 +113,7 @@ func qemuAgentGetInterfacesInfo(domain Domain, wait4ipv4 bool) []libvirt.DomainI
 		Pending:    []string{qemuGetIfaceWait},
 		Target:     []string{qemuGetIfaceDone},
 		Refresh:    qemuAgentInterfacesRefreshFunc(domain, wait4ipv4),
-		MinTimeout:  1 * time.Minute,
+		MinTimeout: 1 * time.Minute,
 		Delay:      30 * time.Second, // Wait this time before starting checks
 		Timeout:    5 * time.Minute,
 	}
--- a/website/docs/r/network.markdown
+++ b/website/docs/r/network.markdown
@@ -35,8 +35,8 @@ resource "libvirt_network" "kube_network" {
 
   # (optional) the bridge device defines the name of a bridge device
   # which will be used to construct the virtual network.
-  # (only necessary in "bridge" mode)
-  # bridge = "br7"
+  # If left undefined and is required, name will be { resource.name }-br
+  # bridge = "k8snet-br"
 
   # (optional) the MTU for the network. If not supplied, the underlying device's
   # default is used (usually 1500)


dmacvicar:
@vmorris I'd like to revisit this fix.

The PR states the problems originates from the networks being created in parallel.
I would expect then the solution to be around synchronizing the creation of those, or finding out why libvirt does not do that already.

The fact that we always specify a bridge in the XML does not sound right to me, given that libvirt allows to skip it.

MalloZup:
thx @dmacvicar for spotting this. I agree that after some thoughts, @dmacvicar  is right . I owe him a :beer: 

I do think we should investigate why libvirt doesnt' do this natively.

vmorris:
Quite right, it could be handled better. I'm trying to recall what I saw when I poked through it for the PR, but it was too long ago!

MalloZup:
I could not reproduce the issue btw.

Check this: 
https://github.com/dmacvicar/terraform-provider-libvirt/issues/703#issuecomment-699492216

I suspect it could be a libvirt issue with might got fixed.

I could not find anything related in libvirt releases notes

dmacvicar:
If I base a testcase with the provided example, I can reproduce it right away (with this PR reverted).

```golang
func TestAccLibvirtNetwork_TwoNetworks(t *testing.T) {
	skipIfPrivilegedDisabled(t)

	randomNetworkResource1 := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
	randomNetworkName1 := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)

	randomNetworkResource2 := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
	randomNetworkName2 := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)

	resource.Test(t, resource.TestCase{
		PreCheck:     func() { testAccPreCheck(t) },
		Providers:    testAccProviders,
		CheckDestroy: testAccCheckLibvirtNetworkDestroy,
		Steps: []resource.TestStep{
			{
				Config: fmt.Sprintf(`
					resource "libvirt_network" "%s" {
					  name = "%s"
					  mode = "nat"
					  addresses = [ "10.0.1.0/24" ]
					  dhcp {
						enabled = true
					  }
					  dns {
						enabled = true
						local_only = true
					  }
					}

					output "%s" {
					  value = libvirt_network.%s.name
					}

					resource "libvirt_network" "%s" {
					  name = "%s"
					  mode = "nat"
					  addresses = [ "10.10.1.0/24" ]
					  dhcp {
						enabled = true
					  }
					  dns {
						enabled = true
						local_only = true
					  }
					}

					output "%s" {
					  value = libvirt_network.%s.name
					}`, randomNetworkResource1, randomNetworkName1, randomNetworkResource1, randomNetworkResource1,
					randomNetworkResource2, randomNetworkName2, randomNetworkResource2, randomNetworkResource2),
			},
		},
	})
}
```

Then, adding a mutex to have only one network be created in parallel makes the test pass again.

```diff
diff --git a/libvirt/config.go b/libvirt/config.go
index 493d29ae..e29b6d59 100644
--- a/libvirt/config.go
+++ b/libvirt/config.go
@@ -2,6 +2,7 @@ package libvirt
 
 import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/mutexkv"
+	"sync"
 	libvirt "github.com/libvirt/libvirt-go"
 	"log"
 )
@@ -15,6 +16,7 @@ type Config struct {
 type Client struct {
 	libvirt     *libvirt.Connect
 	poolMutexKV *mutexkv.MutexKV
+	networkMutex sync.Mutex
 }
 
 // Client libvirt, generate libvirt client given URI
diff --git a/libvirt/resource_libvirt_network.go b/libvirt/resource_libvirt_network.go
index 0f6e874b..1298b801 100644
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -370,6 +370,10 @@ func resourceLibvirtNetworkUpdate(d *schema.ResourceData, meta interface{}) erro
 // resourceLibvirtNetworkCreate creates a libvirt network from the resource definition
 func resourceLibvirtNetworkCreate(d *schema.ResourceData, meta interface{}) error {
 	// see https://libvirt.org/formatnetwork.html
+	client := meta.(*Client)
+	client.networkMutex.Lock()
+	defer client.networkMutex.Unlock()
+
 	virConn := meta.(*Client).libvirt
 	if virConn == nil {
 		return fmt.Errorf(LibVirtConIsNil)
```

Even further, we can lock only around the _Define_ call, which seems to be the only resolving the bridge name. It seems that at _Create_ time, the definition already has the name.

```diff
diff --git a/libvirt/config.go b/libvirt/config.go
index 493d29ae..e29b6d59 100644
--- a/libvirt/config.go
+++ b/libvirt/config.go
@@ -2,6 +2,7 @@ package libvirt
 
 import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/mutexkv"
+	"sync"
 	libvirt "github.com/libvirt/libvirt-go"
 	"log"
 )
@@ -15,6 +16,7 @@ type Config struct {
 type Client struct {
 	libvirt     *libvirt.Connect
 	poolMutexKV *mutexkv.MutexKV
+	networkMutex sync.Mutex
 }
 
 // Client libvirt, generate libvirt client given URI
diff --git a/libvirt/resource_libvirt_network.go b/libvirt/resource_libvirt_network.go
index 0f6e874b..1298b801 100644
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -370,6 +370,10 @@ func resourceLibvirtNetworkUpdate(d *schema.ResourceData, meta interface{}) erro
 // resourceLibvirtNetworkCreate creates a libvirt network from the resource definition
 func resourceLibvirtNetworkCreate(d *schema.ResourceData, meta interface{}) error {
 	// see https://libvirt.org/formatnetwork.html
+	client := meta.(*Client)
+	client.networkMutex.Lock()
+	defer client.networkMutex.Unlock()
+
 	virConn := meta.(*Client).libvirt
 	if virConn == nil {
 		return fmt.Errorf(LibVirtConIsNil)
```

Before I commit such a fix, I would like to reproduce it with plain libvirt and understand if there is a bug report that can be referenced. If not, I'd create one and reference it in the code so that we can remove the workaround in the future.

MalloZup:
ok nice! I was also thinking to use a networkMutex like the pool one.

Yes make sense to see if there is a bug report and reference the workaround. I could not personally find any bug report in this context.

dmacvicar:
I was able to replicate this with plain _virsh_ and XML files. I opened https://gitlab.com/libvirt/libvirt/-/issues/78

dmacvicar:
Uhm. Golang _defer_ scope is function, so my block does not help much there. Need to think how to restrict it to define. Perhaps an anonymous function.

MalloZup:
@dmacvicar  we don't need to use the standard "sync" mutex.

We use for that the special mutex from TF.

we should create a mutex per network name

eg: https://github.com/terraform-providers/terraform-provider-aws/blob/3b002dfdfaa672c7e2527c1dfd322723b394e624/aws/resource_aws_appsync_resolver.go#L136

https://godoc.org/github.com/hashicorp/terraform/helper/mutexkv#MutexKV.Lock

dmacvicar:
I know about `MutexKV`. I think it is designed to lock by key. For example the _Pool_ mutex does not have as objective to allow one operation on pools, but one operation on a *specific* pool. That is why the lock takes a name.

In the case of networks, we are't looking to protect eg. the same network to be modified twice, but any network to be created in parallel. This is not the case _MutexKV_ is designed for. You have no _id/name_ to give to the lock operation.

dmacvicar:
The problem that has me busy now is trying to make the mutex as granular as possible. Locking entirely _resourceLibvirtNetworkCreate_ starting from `virConn.NetworkDefineXML(data)` works well, but it means networks do not get started in parallel, which adds many seconds to network creation.
If I only lock `virConn.NetworkDefineXML(data)` then `network.Create()` happens in parallel, and this creates the problem that sometimes wrong DHCP ranges are generated:

```
- FAIL: TestAccLibvirtNetwork_TwoNetworks (10.94s)
    testing.go:635: Step 0 error: errors during apply:

        Error: Error defining libvirt network: virError(Code=1, Domain=0, Message='internal error: range 10.0.1.2 - 10.10.1.253 is not entirely within network 10.0.1.1/24') -   <network>
              <name>csidjurkhp</name>
              <forward mode="nat"></forward>
              <bridge stp="on"></bridge>
              <dns enable="yes"></dns>
              <ip address="10.0.1.1" family="ipv4" prefix="24">
                  <dhcp>
                      <range start="10.0.1.2" end="10.10.1.253"></range>
                  </dhcp>
              </ip>
          </network>

          on /tmp/tf-test451369475/main.tf line 2:
          (source code not available)
```

I am not sure if this corruption also happens in libvirt, or it is a problem in our code. Need to continue investigating.

dmacvicar:
It is on our side. When it happens, the generated XML is wrong. It is not easy to trigger, and adding a debug statement makes it go away on my system.

```
2020/09/27 10:52:34 [DEBUG] Generated XML for libvirt network:
  <network>
      <name>tzfyvviodc</name>
      <forward mode="nat"></forward>
      <bridge stp="on"></bridge>
      <dns enable="yes"></dns>
      <ip address="10.0.1.1" family="ipv4" prefix="24">
          <dhcp>
              <range start="10.0.1.2" end="10.10.1.254"></range>
          </dhcp>
      </ip>
  </network>
2020/09/27 10:52:34 [DEBUG] Generated XML for libvirt network:
  <network>
      <name>hazwpqxuhe</name>
      <forward mode="nat"></forward>
      <bridge stp="on"></bridge>
      <dns enable="yes"></dns>
      <ip address="10.10.1.1" family="ipv4" prefix="24">
          <dhcp>
              <range start="10.10.1.2" end="10.10.1.253"></range>
          </dhcp>
      </ip>
  </network>
```

I think at some point we will have to remove all this magic DHCP code, and follow the libvirt schema (eg. no network-level DHCP setting) 1:1.

vmorris:
nice job guys!

PR #742, opened 2020-05-20
Handle correctly destruction of inactive networks

### LINKED ISSUE #739: terraform destroy does not destroy inactive networks it leaves behind when creation fails
### Linux distribution

Centos7.8

### Terraform version

Terraform v0.12.25

./terraform-provider-libvirt 0.6.2+git.1585292411.8cbe9ad0
Compiled against library: libvirt 4.1.0
Using library: libvirt 4.5.0
Running hypervisor: QEMU 1.5.3
Running against daemon: 4.5.0

### Description

I'm not including exact steps because they might vary from system to system.

I guess what I would say is special /different about this which might be the literal cause of this issue is that I already have some libvirt networks setup on the host. I inadvertently used the same ip address space in the terraform network as was being used by one of the other libvirt networks, this caused the network creation to fail.  However, it actually partially created the network, it shows up in virsh net-list --all.  I ran terraform destroy and it destroyed everything, but, after I fixed the issue and applied it again, it failed creating the network stating that it already existed.  Since terraform failed I would have expected for it to undefine whatever it had done.

I tried refreshing the state, manually removing the network from the state (which it wasn't even in the state because it failed to create, but I did it anyway), etc.

I had to virsh net-undefine network-name by hand to get it to work.

Another effect of having existing libvirt networks is that I had to explicitly declare what bridge to use even though I'm not using bridged mode.  This was also causing network creation to fail because virbr# already existed.  Again, had to undefine the partially created network by hand. (I think this would be related to #703 )

I know this might not be considered a bug, since best-practice would be to start with a clean host, but it was unexpected behavior that I thought was worth documenting somewhere.

MalloZup:
fix #739 

@Bischoff would be cool if you or oscar or kostas could test this.

Basically the reproduction is

create a network and make it inactive. I will test myside

### DIFF
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -475,6 +475,12 @@ func resourceLibvirtNetworkCreate(d *schema.ResourceData, meta interface{}) erro
 	}
 	err = network.Create()
 	if err != nil {
+		// in some cases, the network creation fails but an artifact is created
+		// an 'broken network". Remove the network in case of failure
+		// see https://github.com/dmacvicar/terraform-provider-libvirt/issues/739
+		// don't handle the error for destroying
+		network.Destroy()
+		network.Undefine()
 		return fmt.Errorf("Error creating libvirt network: %s", err)
 	}
 	defer network.Free()
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -475,7 +475,7 @@ func resourceLibvirtNetworkCreate(d *schema.ResourceData, meta interface{}) erro
 	}
 	err = network.Create()
 	if err != nil {
-		return fmt.Errorf("Error clearing libvirt network: %s", err)
+		return fmt.Errorf("Error creating libvirt network: %s", err)
 	}
 	defer network.Free()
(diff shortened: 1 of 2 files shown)


libvirt/resource_libvirt_network.go
```diff
@@ -639,22 +639,24 @@ func resourceLibvirtNetworkDelete(d *schema.ResourceData, meta interface{}) erro
...
-		// we have to restart an inactive network, otherwise it won't be
-		// possible to remove it.
-		if err := network.Create(); err != nil {
-			return fmt.Errorf("Cannot restart an inactive network %s", err)
+		if err := network.Undefine(); err != nil {
+			return fmt.Errorf("Couldn't undefine libvirt network: %s", err)
 		}
 	}
+	// network is active, so we need to destroy it and undefine it
+	if active {
```
Bischoff, line 651: What about using `else` ?

MalloZup: yeah , is ok to use else also :grimacing: . Personally, subjectively I don't always like the `else` statement because you need to carry/remember the if, but depends . I mean here, is kind of subjective taste :grin:

PR #743, opened 2020-05-20
Skip display setup for aarch64 as well

dirkmueller:
The two supported options, vnc and spice are currently pulling in CirrosLogic device emulation, which only exists on x86_64. 

eventually this will have to be reworked to offer virtio-gpu based display which is supported on aarch64, but this gets me one step further.


MalloZup:
thx @dirkmueller  looks good.!

PR #761, opened 2020-08-02
Remove unused function and variables

johscheuer:
I looked into the code for a better understanding of some behaviour of the provider and I found that these variables and this function is not used anymore and they can be removed.

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -18,12 +18,6 @@ import (
 	"github.com/libvirt/libvirt-go-xml"
 )
 
-// deprecated, now defaults to not use it, but we warn the user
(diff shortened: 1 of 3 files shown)


libvirt/domain.go
```diff
@@ -18,12 +18,6 @@ import (
 	"github.com/libvirt/libvirt-go-xml"
 )
 
-// deprecated, now defaults to not use it, but we warn the user
-const skipQemuAgentEnvVar = "TF_SKIP_QEMU_AGENT"
```
MalloZup, line 22: can you remove also https://github.com/dmacvicar/terraform-provider-libvirt/blob/b11e5b79ff40dc8f03d762d92226fabe315955dc/travis/run-tests-inside-guest#L21

PR #785, opened 2020-09-25
Fix pty console integration tests

dmacvicar:
`setConsoles()` should create a `pty` console using the `DomainChardevSourcePty` XML struct and not `fallthrough` the `DomainChardevSourceDev` case, which produces an input/output error when _libvirt_ tried to use the console.


MalloZup (approved):
nice thx !

PR #786, opened 2020-09-26
revert PR n 740

MalloZup:
revert #740


dmacvicar:
I will revert it together with the new fix.

PR #788, opened 2020-09-26
Force creation of network when we can't update due to libvirt limitation

### LINKED ISSUE #784: libvirt_network modification fails
### Linux distribution

``` Ubuntu.. ```

### Terraform version

Terraform v0.12.28

terraform-provider-libvirt 0.6.1+git.1578064534.db13b678
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
Running hypervisor: QEMU 2.11.1
Running against daemon: 4.0.0

___

## Checklist

  - [X] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.
  - [X] Maintainers do not have expertise in every libvirt setting, so please, describe the feature and how it is used. Link to the appropriate documentation

- [X] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

I have setup libvirt_network and then removed the domain option, because I decided I don't want the libvirt dhcp server to send domain and the network should be only local. But the network update was not succesful. It seems like a general bug, because the network update should be possible and the virsh command  `net-edit --network herring-app-real` and then destroy/start works.

I got error:
```
Error: Error when updating domain in herring-app-real: virError(Code=35, Domain=19, Message='network_update_xml:1: Document is empty
(null)
^')
```

### Setup
```diff 
resource "libvirt_network" "app-real" {
   name = "${var.cluster_name}-app-real"
   autostart = true
   mode = "none"
-  domain = "app-real"
   addresses = ["123.456.789.0/24"]
   dhcp {
     enabled = true
   }
   dns {
     enabled = false
   }
 }
```

### Steps to Reproduce Issue

Create libvirt_network resource with domain set apply terraform, then remove the domain and apply terraform.

MalloZup:
*Edited by @dmacvicar *

>this PR fixes issue #784
>
>We are passing libvirt.NETWORK_UPDATE_COMMAND_MODIFY with an empty element to remove an element, when the right thing to do should be to pass ibvirt.NETWORK_UPDATE_COMMAND_DELETE and passing the actual element. This creates the error about the empty XML element.
>
>However, it is not possible to update this attribute anyway ([source](https://gitlab.com/search?utf8=%E2%9C%93&search=virNetworkDefUpdateNoSupport&group_id=130330&project_id=192693&scope=&search_code=true&snippets=false&repository_ref=master&nav_source=navbar)), so the right fix is to `ForceNew`

### DIFF
--- a/libvirt/network.go
+++ b/libvirt/network.go
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/libvirt/libvirt-go"
-	"github.com/libvirt/libvirt-go-xml"
+	libvirtxml "github.com/libvirt/libvirt-go-xml"
 )
 
 func waitForNetworkActive(network libvirt.Network) resource.StateRefreshFunc {
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -10,7 +10,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/libvirt/libvirt-go"
-	"github.com/libvirt/libvirt-go-xml"
+	libvirtxml "github.com/libvirt/libvirt-go-xml"
 )
 
 const (
@@ -54,7 +54,8 @@ func resourceLibvirtNetwork() *schema.Resource {
 			"domain": {
 				Type:     schema.TypeString,
 				Optional: true,
-				ForceNew: false,
+				// libvirt cannot update it so force new
+				ForceNew: true,
 			},
 			"mode": { // can be "none", "nat" (default), "route", "bridge"
 				Type:     schema.TypeString,
@@ -342,25 +343,6 @@ func resourceLibvirtNetworkUpdate(d *schema.ResourceData, meta interface{}) erro
 		}
 
 		d.SetPartial("bridge")
-	}
-
-	// detect changes in the domain
-	if d.HasChange("domain") {
-		networkDomain := getDomainFromResource(d)
-
-		data, err := xmlMarshallIndented(networkDomain)
-		if err != nil {
-			return fmt.Errorf("serialize update: %s", err)
-		}
-
-		log.Printf("[DEBUG] Updating domain for libvirt network '%s' with XML: %s", networkName, data)
-		err = network.Update(libvirt.NETWORK_UPDATE_COMMAND_MODIFY, libvirt.NETWORK_SECTION_DOMAIN, -1,
-			data, libvirt.NETWORK_UPDATE_AFFECT_LIVE|libvirt.NETWORK_UPDATE_AFFECT_CONFIG)
-		if err != nil {
-			return fmt.Errorf("Error when updating domain in %s: %s", networkName, err)
-		}
-
-		d.SetPartial("domain")
 	}
 
 	d.Partial(false)


dmacvicar:
Can you explain what problem is being solved, in the PR?

MalloZup:
In (https://github.com/dmacvicar/terraform-provider-libvirt/issues/784) we do have a scenario where user do following:

```
Resource "libvirt_network" "app-real" {
   name = "${var.cluster_name}-app-real"
   autostart = true
   mode = "none"
-  domain = "app-real"
   addresses = ["123.456.789.0/24"]
   dhcp {
     enabled = true
   }
   dns {
     enabled = false
   }
 }
```
#### Steps to Reproduce Issue

1) Create libvirt_network resource with domain set apply terraform
2) Then remove the domain attr from terraform and apply terraform.

### Errors:

```
Error: Error when updating domain in herring-app-real: virError(Code=35, Domain=19, Message='network_update_xml:1: Document is empty
```
I could also reproduce locally.

There are 2 issues.
 * remove the `domain` attributes (like the user did)
 * rename the attribute. ( e.g domain from foo to bar)

Both doesn't work

The terraform code detect that detect the change is called correctly  but the libvirt API call `network.update` doesn't update the network

https://github.com/dmacvicar/terraform-provider-libvirt/blob/d1cf93cdb066b2d3d2c2c046d2fb986808d9a424/libvirt/resource_libvirt_network.go#L357


-->

https://libvirt.org/html/libvirt-libvirt-network.html#virNetworkUpdate


Even if in this API there is a section for updating the Domain, https://libvirt.org/html/libvirt-libvirt-network.html#virNetworkUpdateSection .
I could after not update the value of an existing Network . ( I tried with different flags also  https://libvirt.org/html/libvirt-libvirt-network.html#virNetworkUpdateCommand) etc without sucess.


I did also the test with native `libvirt` with `virsh` and it could not update the network.

I checked the upstream doc  and found out that they don't document the `domain` as possible section that can be updated.
https://wiki.libvirt.org/page/Networking#virsh_net-update

They state also  about "Arbitrary changes to the network" ( see page)

```
Although the most common cases of changing network config can be handled with "virsh net-update", there are some parts of the config that can't be modified in this way, and in those cases you will be left with all running guests detached from the network after it is restarted. In order to solve this problem, one possible approach would be to use a script to re-attach all interfaces on all machines after the network has been started.

An example of such script (which worked at one time in the past, and may still work, but has been reported to *not* work by at least one user) is available here. Links to an updated/verified operational script are welcome. 
```

So that is mostly why I do think that the `network.update` has something faulty on this, especially the domain update sections

I will research more on the evidence in the source code.

MalloZup:
I found also out: https://www.redhat.com/archives/libvirt-users/2010-June/msg00044.html


 libvirt for Network doesn't change network .

https://serverfault.com/a/573255

That is I do think we could imho just Recreate the resources.  I do think handling a `network` udpdate can be tricky, since we might need to hook perform all this `virsh destroy` operation after an update

dmacvicar (changes requested):
Please in future PRs put the summary and motivation in the description of the PR. This one requires reading across comments and linked issues to get the full picture. If you can edit the current description, better.

I think the conclusion you arrived on is correct, however it may be connected to the original report by luck. The original error message is about the document being empty, which is likely related to using `libvirt.NETWORK_UPDATE_COMMAND_MODIFY` vs `libvirt.NETWORK_UPDATE_COMMAND_DELETE` and passing empty XML documents, while [my understanding](https://www.spinics.net/linux/fedora/libvir/msg64080.html) is that to remove we should pass the full _current_ entry and `libvirt.NETWORK_UPDATE_COMMAND_DELETE`. This is the root of the reported problem. Not what this PR fixes.

But, it is correct, given that _any_ modification to the domain attribute would not be supported. There is no point in implementing deletion. I was able to build a testcase (*Note*: we should discuss all PRs over a testcase reproducing the issue) and then implement deletion correctly just to arrive to the "unsupported" message.

For which operations are unsupported I think a forum of 2010 is not the right source, but [this one](https://gitlab.com/search?utf8=%E2%9C%93&search=virNetworkDefUpdateNoSupport&group_id=130330&project_id=192693&scope=&search_code=true&snippets=false&repository_ref=master&nav_source=navbar). We should make a note in the resource schema definition and mark those attributes as `ForceNew`.

MalloZup:
@dmacvicar  thx for your comments and findings. I agree about the testcase if we can find one.

I do actually have tested locally, but I was not precise on documenting my findings with the source code and justifying them. 

I will follow up with some PRs because we do need to force resource for network and create a readme.md update on the design.

If you want to take over on this let me know otherwise I will just do it once I do have free time

dmacvicar:
I have added a note in the source code already.

I don't think there is anything left to do here.

PR #789, opened 2020-09-27
Synchonize network defines to workaround libvirt bridge name race condition

dmacvicar:
This PR means to supersede/replace https://github.com/dmacvicar/terraform-provider-libvirt/pull/740
 
libvirt assigns the same bridge to networks defined in parallel (See https://gitlab.com/libvirt/libvirt/-/issues/78).

This code synchronizes network definition.

Note that there is still a problem in the code that generates wrong DHCP ranges. I could only prove it comes from our generated XML, but I could not figure how the race happens. We will need to address that as a separate issue.
I believe we need to remove most of the network code, and be a thin layer on top of libvirt. It may require us to align our definitions, eg. no _dhcp_ option in the _network_ block, and a explicit _range_ option.

Fixes https://github.com/dmacvicar/terraform-provider-libvirt/issues/703

### DIFF
--- a/libvirt/config.go
+++ b/libvirt/config.go
@@ -4,6 +4,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/mutexkv"
 	libvirt "github.com/libvirt/libvirt-go"
 	"log"
+	"sync"
 )
 
 // Config struct for the libvirt-provider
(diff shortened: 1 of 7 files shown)


MalloZup (approved):
nice work. I always like when people put comments reference/links in code which helps to remember why it exists there :+1: 

e.g I think we missed an opportunity to document why the Pool locks are there. i

MalloZup:
I thnk for the DHCP issue we can have a follow-up issue.

Issue #747, opened 2020-06-09
Use provider within the registry of Terraform hashicorp.


paultyng:
👋 Hi, I'm on the Terraform Providers team at HashiCorp. With the [release of the Terraform 0.13 beta](https://discuss.hashicorp.com/t/terraform-0-13-beta-released/9555), users can now download and install community providers from the registry. We are inviting provider authors (especially those for popular community providers) to publish their providers in a closed beta.

To get invited to the closed beta, please email terraform-registry-beta@hashicorp.com. We need:

- A list of GitHub usernames to add to the beta (you and any team members who will publish)
- List of provider repositories you plan to publish
- A GPG public key in ASCII-armor format, which you will be using to sign your provider releases

You can use one key for all of your providers, or separate keys if you prefer. If you are publishing from an organization, this key or keys will be associated with that namespace. Once in the beta, you can manage personal keys in the UI as well.

MalloZup:
@paultyng hi thx for feedback. We will come to you on this soon

limburgher:
Hi! terraform 0.13 was just released publicly, and I'm hitting this. I'm trying to add a source entry to use the local binary but I can't work out the syntax. Do you anticipate this being published soon?

MalloZup:
@limburgher hi ! yes I plan to update the provider to TF.0  13 soonish. We plan to release a first a fixes releases then the next one will be the TF 0.13.. stay tuned :sunflower:

paultyng:
👍 Let me know if there is anything I can do to help. This should be self-serve now in the registry UI, you just need to create releases that conform to the [publishing criteria](https://www.terraform.io/docs/registry/providers/publishing.html) (we have a [goreleaser](https://github.com/hashicorp/terraform-provider-scaffolding/blob/master/.goreleaser.yml) config file that can do it).

wenerme:
@MalloZup also blocked by this, don't find a way to add locally built provider

limburgher:
Thank you! I may be doing it wrong, as I'm still having issues. I created ~/.local/share/terraform/plugins/registry.terraform.io/dmacvicar/libvirt/0.6.2/linux_amd64, and put 0.6.2 of terraform-provider-libvirt there. Then I added the above block to my providers.tf. terraform init then ends with:

- Finding latest version of -/libvirt...

Error: Failed to query available provider packages

Could not retrieve the list of available versions for provider -/libvirt:
provider registry registry.terraform.io does not have a provider named
[...]

PR #797, opened 2020-10-28
Improve terraform v0.13 migration doc

arbulu89:
Improve the terraform v0.13 migration document to explicitily say that the new providers code snippet must be added to all used terraform modules (not just the main one)/

Based on: https://github.com/dmacvicar/terraform-provider-libvirt/issues/747#issuecomment-678575669


MalloZup (approved):
thx

Issue #795, opened 2020-10-22
dubious code for random mac address determination


Bischoff:
### Terraform version

commit 9e02168780b83e158c050cbe1f32102304861e58 October 11, 2020

### Setup

Seen on source code. problem was introduced by:
```
commit 625e9420e1a89d65f910551f5057c72a0a6a12eb
Author: root <root@zimbra-mta.univ-brest.fr>
Date:   Tue Feb 5 11:18:26 2019 +0100
```

## The issue

This function appends 3 randoms bytes after `52:54:00:` to build a random MAC address:

```
// randomMACAddress returns a randomized MAC address
// with libvirt prefix
func randomMACAddress() (string, error) {
        buf := make([]byte, 3)
        rand.Seed(time.Now().UnixNano())
        _, err := rand.Read(buf)
        if err != nil {
                return "", err
        }

        // set local bit and unicast
        buf[0] = (buf[0] | 2) & 0xfe
        // Set the local bit
        buf[0] |= 2

        // avoid libvirt-reserved addresses
        if buf[0] == 0xfe {
                buf[0] = 0xee
        }

        return fmt.Sprintf("52:54:00:%02x:%02x:%02x",
                buf[0], buf[1], buf[2]), nil
}
```

There is some processing to set "local" bit and "unicast" bit on first byte, as well as to avoid reserved value `0xFE`  for that first byte.

The problem is: the first byte in the array is the fourth byte in the MAC address, and the fourth byte in the MAC address does not have those constraints. The first byte in the MAC address is always `0x52` and is already respecting these constraints.

Reference:

https://upload.wikimedia.org/wikipedia/commons/9/94/MAC-48_Address.svg

## Suggested fix

Either:
* let all 6 bytes be picked at random, as it was before this commit,
or
* remove unnecessary lines 28-37

Note: setting the prefix to `52:54:00` also has the undesirable effect of reducing the entropy (and therefore augmenting the chance of collisions).

Hope that helps.

MalloZup:
thx @Bischoff  I think we might need to spend more time on this.

If you feel , you can still send a PR to help us a bit. thx anyways for issue!

Bischoff:
I am lacking information here.

I would suggest to "fix" it by simply reverting the commit 625e9420e1a89d65f910551f5057c72a0a6a12eb from root@zimbra-mta.univ-brest.fr .

[...]

PR #799, opened 2020-11-02
Remove forgotten code in randomMacAddress

Bischoff:
This PR solves issue #795 by removing leftover code.

### DIFF
--- a/libvirt/utils_net.go
+++ b/libvirt/utils_net.go
@@ -39,23 +39,13 @@ func (l *networkUpdateWorkaroundLibvirt) NetworkUpdate(Net libvirt.Network, Comm
 }
 
 // randomMACAddress returns a randomized MAC address
-// with libvirt prefix
+// with QEMU prefix
 func randomMACAddress() (string, error) {
 	buf := make([]byte, 3)
 	rand.Seed(time.Now().UnixNano())
 	_, err := rand.Read(buf)
 	if err != nil {
 		return "", err
-	}
-
-	// set local bit and unicast
-	buf[0] = (buf[0] | 2) & 0xfe
-	// Set the local bit
-	buf[0] |= 2
-
-	// avoid libvirt-reserved addresses
-	if buf[0] == 0xfe {
-		buf[0] = 0xee
 	}
 
 	return fmt.Sprintf("52:54:00:%02x:%02x:%02x",


Bischoff:
Ping @MalloZup and @Isma399 .

libvirt/utils_net.go
```diff
@@ -16,7 +16,7 @@ const (
 )
 
 // randomMACAddress returns a randomized MAC address
-// with libvirt prefix
+// with QEMU prefix
```
MalloZup, line 19: can we have some documentation or something which can be a comment to refer to this?

MalloZup (commented):
@Bischoff  thx for this I would like however that we document better in the code, maybe with an url or doc all this magic prefix somehow

libvirt/utils_net.go
Bischoff: That will probably be hard :-) From what I have seen in the forums, it is a "known fact" but it is hard to find a trusted, official, source.

If you find one, feel free to add it in the comment.

Bischoff:
> @Bischoff thx for this I would like however that we document better in the code, maybe with an url or doc all this magic prefix somehow

I would love that too. So far I failed to find such a thing, but maybe I just missed it... Isma399 , do you have better insights?

Isma399:
No, I have looked for documentation and didn't find anything. 
Neither in http://standards-oui.ieee.org/oui.txt, nor in qemu documentation.

Bischoff:
The, I would suggest just merging like this. This is apparently the kind of facts that everyone uses but that is documented nowhere.

MalloZup (approved):
I'm ok with this. I'm adding Duncan if he needs more info on this.

thx!

libvirt/utils_net.go
MalloZup: we can put an issue link of this github project. that could be  a doc hint @Bischoff

dmacvicar:
We will skip this for this release. I need some documentation and testcases.

MalloZup:
I think for this PR we need to find out evidence in the codebase https://gitlab.com/libvirt/libvirt and document why we are doing this change and documenting it  + a testcase for this..

Of course previously the behavior was wrong or without technical facts, and  this PR fix it somehow but we need to close the gap and make fix it correctly with a doc comment or link something that justify it more.

Thx @Bischoff and @Isma399  for PR , would be great if one of you could help us to finish this :pray: :sunflower:

dmacvicar (changes requested):
Setting as "Request changes" while tests or documentation is added.

Bischoff:
Test and documentation should have been added to the initial PR that introduced the problem. I won't add them here, it's out of scope of the bug I am trying to fix.

PR #803, opened 2020-11-11
Fix flatcar-linux example network typo

kskewes:
Per: https://docs.flatcar-linux.org/ignition/network-configuration/


dmacvicar:
Thanks!

PR #808, opened 2020-11-11
Remove vendor

dmacvicar:
Packages in OBS will use the go_modules service

This is already done in the [unstable packages](https://build.opensuse.org/package/show/systemsmanagement:terraform:unstable/terraform-provider-libvirt), which use the service to create _vendor.tar.gz_ (and delete vendor/ before unpacking the bundled one).


MalloZup (approved):
looks good thx! @dmacvicar  recently I was thinking also if we could expose the `obs` packaging rspec and beeing managed via git.

Something we do already in various projects.

https://github.com/ClusterLabs/ha_cluster_exporter/blob/master/packaging/obs/prometheus-ha_cluster_exporter/prometheus-ha_cluster_exporter.spec

But I think for the 0.7.0 if the plan is to be in registry the "packaging" story is becoming afaik superficial , so imho we don't need it given our next prios and context.

PR #809, opened 2020-11-12
Add reference to terraform-libvirt-kubespray project

MusicDin:
I've written a [Terraform script](https://github.com/MusicDin/terraform-kvm-kubespray) for setting up HA kubernetes cluster on KVM/libvirt.

I think this could be referenced from your README file, because it's written with libvirt provider and before I wrote a script I couldn't find anything like that on web. 
So maybe it will help someone else out in the same situation.


dmacvicar:
Thank you!

Issue #811, opened 2020-11-24
Disks in domain xml definition end up with driver type='raw' instead of type='qcow2'


bpetit:
### Linux distribution

Ubuntu 20.04.1 LTS

### Terraform version

```sh
terraform -v
Terraform v0.13.5
```
### Provider and libvirt versions

```sh
terraform-provider-libvirt -version
./terraform-provider-libvirt 0.6.3+git.1604843676.67f4f2aa
Compiled against library: libvirt 6.0.0
Using library: libvirt 6.0.0
Running hypervisor: QEMU 4.2.1
Running against daemon: 6.0.0
```
___

## Checklist

- [ ] Is your issue/contribution related with enabling some setting/option exposed by libvirt that the plugin does not yet support, or requires changing/extending the provider terraform schema?

  - [ ] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.
  - [ ] If it is a very special case, consider using the _XSLT_ support in the provider to tweak the definition instead of opening an issue
  - [ ] Maintainers do not have expertise in every libvirt setting, so please, describe the feature and how it is used. Link to the appropriate documentation

- [x] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

### Setup

(Please provide the full _main.tf_ file for reproducing the issue (Be sure to remove sensitive information)

```
resource "libvirt_volume" "scaphtest01" {
    name = "scaphtest01.qcow2"
    source = "/home/bpetit/git/scaphandre/tests/qemu/1/template-debian-10.qcow2"
    pool = libvirt_pool.ansible_iaas_pool.name
    format = "qcow2"
    depends_on = [libvirt_pool.ansible_iaas_pool]
}
resource "libvirt_network" "iaaslocal" {
    name = "iaaslocal"
    mode = "nat"
    domain = "example.local"
        addresses = [
        "192.168.99.1/24"
        ]
    }
output "instance_ip_addr" {
    value = libvirt_domain.scaphtest01.network_interface.0.addresses
}
resource "libvirt_pool" "ansible_iaas_pool" {
  name = "ansible_iaas_pool"
  type = "dir"
  path = "/var/lib/libvirt/images/ansible_iaas"
}
terraform {
 required_version = ">= 0.13"
  required_providers {
    libvirt = {
      source  = "dmacvicar/libvirt"
      version = "0.6.3"
    }
  }
}
provider "libvirt" {
[...]

PR #812, opened 2020-11-24
Fixing issue #811

bpetit:

Fixing https://github.com/dmacvicar/terraform-provider-libvirt/issues/811

For some reason disk.Driver.Type ends up with value 'raw', even when file name of the disk drive has a ".qcow2" suffix.
Is seems like checking that the suffix is there and then apply qcow2 as the driver type is not done explicitely, so raw is choosen in this case.

This fix proposal fixed the workflow (described in issue 811) on my environment, tests are passing successfully. I propose you this fix and hope it is useful. Feel free to slap my fingers if I missed something in the contribution process or assumed something wrong or anything :)

Thank you for your great work ! I use the provider a lot and love it.

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -593,7 +593,9 @@ func setDisks(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *lib
 				numOfISOs++
 			}
 
-			if !strings.HasSuffix(file.(string), ".qcow2") {
+			if strings.HasSuffix(file.(string), ".qcow2") {
+				disk.Driver.Type = "qcow2"
+			} else {
 				disk.Driver.Type = "raw"
 			}
 		} else if blockDev, ok := d.GetOk(prefix + ".block_device"); ok {


dmacvicar:
Thanks for the PR!

Do you think you could reproduce the issue with a testcase ie. enhancing the ones in resource_libvirt_volume_test.go enough to make it fail before applying the fix?

bpetit:
I'll have a look at that and try !

joe-e-brown:
The bug doesn't affect the volume definition, but rather the domain definition.
Whereas a volume may have qcow2 format, the domain is defined with XML which specifies the RAW driver instead.
This results in the domain being unbootable because the volume driver doesn't match the volume format.
I propose that the tests in [resource_libvirt_domain_test.go](https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_domain_test.go) should be enhanced to test the volume driver selection after domain creation.

bpetit:
I didn't take time to look at the tests implementation yet. It will require some digging on my side so if someone has a clear idea on how it should be done properly, please do :)

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #813, opened 2020-11-28
Pure Go terraform-provider-libvirt experiment

dmacvicar:
This is an _experiment_ to attempt to make the provider independent of the operating system. The motivation is to be able to distribute it via the terraform registry (https://github.com/dmacvicar/terraform-provider-libvirt/issues/747). As an experiment, it is not guaranteed that the changes will make into the main branch.

The strategy is to move from [libvirt-go](https://github.com/libvirt/libvirt-go) to [go-libvirt](https://github.com/digitalocean/go-libvirt). The later implements the _libvirt_ RPC protocol natively in Go.

The main challenges are:

* The [go-libvirt](https://github.com/digitalocean/go-libvirt) API is different. Closer to the [C API](https://libvirt.org/html/index.html).

  The [libvirt-go](https://github.com/libvirt/libvirt-go) API attempts to be more object oriented. This means carrying the libvirt connection object into many helper functions, resulting in diffs like this:

```diff
-func domainIsRunning(domain libvirt.Domain) (bool, error) {
-	state, _, err := domain.GetState()
+func domainIsRunning(virConn *libvirt.Libvirt, domain libvirt.Domain) (bool, error) {
+	state, _, err := virConn.DomainGetState(domain)
 	if err != nil {
 		return false, fmt.Errorf("Couldn't get state of domain: %s", err)
 	}
 
-	return state == libvirt.DOMAIN_RUNNING, nil
+	return libvirt.DomainState(state) == libvirt.DomainRunning, nil
 }
 ```

* There are missing features, like _PolicyKit_ authentication (https://github.com/digitalocean/go-libvirt/issues/110)
* We don't know much about the maintainability of the native client library, though Digital Ocean is supposedly using it internally

(we will document more as we learn in this branch)

To make it easier:

* No refactorings, and just have as a goal compilation and later passing most of the testsuite
* The old import was renamed from _libvirt_ to _libvirtc_ and opens both connections simultaneously, so we can port the code that calls one client with code that calls the native client

Most of the changes are tedious and mechanic. This is a great opportunity to read a lot of the code and get an idea of what improvements and cleanups can be done later.

If you are interested in helping this experiment, get in touch with me so that we work in different areas/files.

## Ideas for post-port refactorings and clenups

Brainstorming some ideas [here](https://gist.github.com/dmacvicar/e6a694b7b18a00ddd15261022c486007)

PR #814, opened 2020-11-30
LVM storage pool implementation

dpap-qs:
Logical storage pool implementation.

Our team needed the lvm storage pool, so I implemented it. We tested, it seems fine.
What do you think  @dmacvicar ?
If you say it's ok, then I write acceptance tests, and add the documentation.

If everything is fine, I think we should discuss about this:
The volumes created in logical storage pool, don't preserve de volume `format`. See: [pool](https://libvirt.org/storage.html#StorageBackendLogical) and [volume](https://libvirt.org/formatstorage.html#StorageVol). It's a problem when you try to determine the disk driver based on volume format during the domain creation here:https://github.com/dmacvicar/terraform-provider-libvirt/blob/30595333cabd9d1651449c18acfaa2f99aaef928/libvirt/domain.go#L513
So right now we can only use the default disk driver, which is raw, so e.g. cqow2 cannot be used.
I see two options:
1. The more elegant: extend the `disk` resource with a `driver` attribute. If it's provided, we use it, otherwise we should use the current logic.
2. Use the volume name to figure out the driver (as you do with files and url-s) e.g. if it ends with `.raw` or `.qcow2`, use them.

### DIFF
--- a/libvirt/resource_libvirt_pool.go
+++ b/libvirt/resource_libvirt_pool.go
@@ -67,6 +67,16 @@ func resourceLibvirtPool() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
+
+			// logical-specific attributes
+			"source_devices": {
+				Type:     schema.TypeList,
+				Optional: true,
+				ForceNew: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
 		},
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
@@ -81,8 +91,8 @@ func resourceLibvirtPoolCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	poolType := d.Get("type").(string)
-	if poolType != "dir" {
-		return fmt.Errorf("Only storage pools of type \"dir\" are supported")
+	if poolType != "dir" && poolType != "logical" {
+		return fmt.Errorf("Only storage pools of type \"dir\" and \"logical\" are supported")
 	}
 
 	poolName := d.Get("name").(string)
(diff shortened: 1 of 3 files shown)


dmacvicar:
Thanks for the PR. I need to dive a bit into the libvirt documentation to understand well how this is supposed to work.

x70b1:
Any progress? 

@dpap-qs I have tested this. Its awesome! :sunglasses:

MalloZup:
we will review this PR after the `13` release see #813 PR which . Stay tuned :)

x70b1:
@dpap-qs A suggestion: At the moment the name of the storage pool in libvirt is the name of the volume group. 
We could reuse the `path` setting to select a volume group (my-volumegroup) and create a volume pool (example-pool):

```
resource "libvirt_pool" "example" {
  name = "example-pool"
  type = "logical"
  path = "my-volumegroup"
}
```

atlasloewenherz:
any update on this guys??

libvirt/resource_libvirt_pool.go
```diff
@@ -99,17 +109,56 @@ func resourceLibvirtPoolCreate(d *schema.ResourceData, meta interface{}) error {
...
 
-	poolDef := libvirtxml.StoragePool{
-		Type: "dir",
-		Name: poolName,
-		Target: &libvirtxml.StoragePoolTarget{
-			Path: poolPath,
-		},
+	var poolDef *libvirtxml.StoragePool
+	// In some cases we don't need to build the pool
+	needToBuild := true
```
dmacvicar, line 115: Is this really needed with the `VIR_STORAGE_POOL_BUILD_NO_OVERWRITE` flag? eg. [see here](https://github.com/libvirt/libvirt/blob/master/src/storage/storage_backend_logical.c).

dmacvicar (changes requested):
I have rebased this (without testing it), as we change libvirt libraries in between.

I think we could bring it to a point where it can be merged, but I am not happy with the "need to build" exception in the code.

Then someone would need to drive it to complete documentation and tests.

jimnydev:
I'm interested in seeing this implemented. Are there any plans to continue with this? 
If not, perhaps I can try to do a new PR against current `master` instead?

Thanks!

varac:
Can this get closed after https://github.com/dmacvicar/terraform-provider-libvirt/pull/1088 got merged ?

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #816, opened 2020-12-02
Set firmware attribute to efi for aarch64

Prashanth684:
This new firmare attribute was added to newer versions of libvirt-go-xml. This enables to specify the firmware as
'bios' or 'efi' and automatically fills in the loader and NVRAM elements without having to specify them manually.
Refer to https://libvirt.org/formatdomain.html#bios-bootloader for this. As part of this change, libvirt-go and
libvirt-go-xml needs to be updated. The old functionality of specifying the loader and NVRAM is also left as such
so those can continue to be specified if we do not want automatic selection

Setting this automatically for aarch64 as it is a EFI architecture by default.

### DIFF
--- a/libvirt/domain_def.go
+++ b/libvirt/domain_def.go
@@ -117,6 +117,12 @@ func newDomainDefForConnection(virConn *libvirt.Connect, rd *schema.ResourceData
 		d.OS.Type.Arch = arch
 	}
 
+	if d.OS.Type.Arch == "aarch64" {
+		// for aarch64 speciffying this will automatically select the firmware and NVRAM file
+		// reference: https://libvirt.org/formatdomain.html#bios-bootloader
+		d.OS.Firmware = "efi"
+	}
+
 	caps, err := getHostCapabilities(virConn)
 	if err != nil {
 		return d, err
--- a/go.mod
+++ b/go.mod
@@ -7,8 +7,8 @@ require (
 	github.com/hashicorp/terraform-plugin-sdk v1.4.0
 	github.com/hooklift/assert v0.0.0-20170704181755-9d1defd6d214 // indirect
 	github.com/hooklift/iso9660 v1.0.0
-	github.com/libvirt/libvirt-go v5.0.0+incompatible
-	github.com/libvirt/libvirt-go-xml v5.0.0+incompatible
+	github.com/libvirt/libvirt-go v5.10.0+incompatible
+	github.com/libvirt/libvirt-go-xml v5.10.0+incompatible
 	github.com/mattn/goveralls v0.0.2
 	github.com/pborman/uuid v1.2.0 // indirect
 	github.com/stretchr/testify v1.5.1
--- a/go.sum
+++ b/go.sum
@@ -178,10 +178,10 @@ github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348 h1:MtvEpTB6LX3v
 github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348/go.mod h1:B69LEHPfb2qLo0BaaOLcbitczOKLWTsrBG9LczfCD4k=
 github.com/kylelemons/godebug v1.1.0 h1:RPNrshWIDI6G2gRW9EHilWtl7Z6Sb1BR0xunSBf0SNc=
 github.com/kylelemons/godebug v1.1.0/go.mod h1:9/0rRGxNHcop5bhtWyNeEfOS8JIWk580+fNqagV/RAw=
-github.com/libvirt/libvirt-go v5.0.0+incompatible h1:IUbnpRXdveEHGNqZrckQAihg65OuHYb/b90yV89c6yQ=
-github.com/libvirt/libvirt-go v5.0.0+incompatible/go.mod h1:34zsnB4iGeOv7Byj6qotuW8Ya4v4Tr43ttjz/F0wjLE=
-github.com/libvirt/libvirt-go-xml v5.0.0+incompatible h1:zSZ7uRWGGgvyzDHPrA+3sYrTf+Jtx4izPWYQhHfAOqc=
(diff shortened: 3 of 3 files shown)


Prashanth684:
would be nice to expose this as a terraform variable..but there is already a `firmware` variable which actually refers to the firmware file path for the loader. should that be changed  to some other name ?

dmacvicar:
Thanks for the PR!

The issue you describe is common. We try to expose an attribute in HCL without exposing the whole XML sub-block. In this case https://github.com/dmacvicar/terraform-provider-libvirt/pull/50 choice of name `firmware` was not good.

I am not convinced where to put this. Pehaps the default domain definition should handle this already? ie. set it to EFI for ARM when returning the initial empty definition?

Prashanth684:
> Thanks for the PR!
> 
> The issue you describe is common. We try to expose an attribute in HCL without exposing the whole XML sub-block. In this case #50 choice of name `firmware` was not good.
> 
> I am not convinced where to put this. Pehaps the default domain definition should handle this already? ie. set it to EFI for ARM when returning the initial empty definition?

Thanks for the feedback. That makes sense. I added it to the default definition but had to pass in arch as a parameter to set it for ARM. just used "x86_64" in places where it was just used for go tests.

Prashanth684:
> > Thanks for the PR!
> > The issue you describe is common. We try to expose an attribute in HCL without exposing the whole XML sub-block. In this case #50 choice of name `firmware` was not good.
> > I am not convinced where to put this. Pehaps the default domain definition should handle this already? ie. set it to EFI for ARM when returning the initial empty definition?
> 
> Thanks for the feedback. That makes sense. I added it to the default definition but had to pass in arch as a parameter to set it for ARM. just used "x86_64" in places where it was just used for go tests.

@dmacvicar - does this change look ok ?

libvirt/domain_def.go
```diff
@@ -101,20 +101,27 @@ func newDomainDef() libvirtxml.Domain {
 		},
 	}
 
+	if arch == "aarch64" {
+		// for aarch64 speciffying this will automatically select the firmware and NVRAM file
```
MalloZup, line 105:
```suggestion
		// for aarch64 specifying this will automatically select the firmware and NVRAM file
```

libvirt/domain_def.go
```diff
@@ -17,7 +17,7 @@ func getXMLDomainDefFromLibvirt(domain *libvirt.Domain) (libvirtxml.Domain, erro
 		return libvirtxml.Domain{}, fmt.Errorf("Error retrieving libvirt domain XML description: %s", err)
 	}
 
-	domainDef := newDomainDef()
+	domainDef := newDomainDef("x86_64")
```
MalloZup, line 20: I was thinking that looks a bit hardcoded the `x86_64` here.

Personally I would prefer having the `newDomainDef()` with void parameter.

After when the domainDef is created modify the code depending the arch.

I dunno if you did this initially.

What do you think @dmacvicar ?

MalloZup:
thx for PR ! I have added a comment for discussion.

libvirt/domain_def.go
dmacvicar: I think the best way to get a good default is to use ` virConnectGetDomainCapabilities`, which gives you the arch for the domains before creating it.

Originally I thought that we get the capabilities after we create the empty definition (and we are, but that is complicated to explain). However, in the "Read" section, we call:

```golang
domainDef, err := newDomainDefForConnection(virConn, d)
```

Definition options are overwritten with the capabilities one if they are not specified by the domain definition. I really wonder about this part. Perhaps has to do that "Read" is called at the end of "Create", and this makes sure computed attributes are set if not specified in the terraform data. Dunno.

//cc @Prashanth684 

This method is a method that returns a definition having access to the connection and capabilities, which means it has access to the default architecture.

Perhaps using this call in the "Create" section instead of the vanilla `newDomainDef` is a good solution. We can have `newDomainDefForConnection` set the default bootloader just after it sets the arch (either specified, or from the capabilities).

I don't know if using `newDomainDefForConnection` in "Create" has other consequences.

Prashanth684: @MalloZup  @dmacvicar makes sense...i did put it in `newDomainDefForConnection` first but thought to put it in `newDomainDef` later in case we miss some places where we set it, but doesn't look like it. i will make that change. i didn't very much like hardcoding it as well.

Yes , that `Read` call seems to just fill the schema with all the computed attributes , but i also wonder if we need to set things like `OS.Type.Machine` in `newDomainDefForConnection` since libvirt should figure those automatically, but that warrants a separate discussion.

Prashanth684:
updated PR per comments.

MalloZup (approved):
I think this PR is good enough however I would like we reference a follow-up issue to address the discussion we started during review which is maybe a cleaner refactor.

dmacvicar (commented):
I really like where the code is right now, but as I mentioned `newDomainDefForConnection` is executed on `Read` which happens on an explicit `Read` or at the _end_ of `Create`, which means after the XML is injected into _libvirt_.

Are you sure this is working and producing the effect you expect?

Prashanth684:
> I really like where the code is right now, but as I mentioned `newDomainDefForConnection` is executed on `Read` which happens on an explicit `Read` or at the _end_ of `Create`, which means after the XML is injected into _libvirt_.
> 
> Are you sure the code working and producing the desired effect?

`newDomainDefForConnection` is called in at the beginning of create also: https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/resource_libvirt_domain.go#L443. Yes, i tested and it works

dmacvicar (approved):

dmacvicar:
Thanks for the contribution @Prashanth684

Prashanth684:
> Thanks for the contribution @Prashanth684

Thank you for helping to get this in!

Issue #642, opened 2019-09-10
RFE: Support dnsmasq options


zeenix:
Libvirt 5.6 added support for [passing verbatim options](https://libvirt.org/formatnetwork.html#elementsNamespaces) to `dnsmasq`. In Openshift, we'd likely want to use that to create generic DNS records with wildcards. I'm sure there are other uses of this and this is why this new feature was added. I haven't yet figured how this support would/could look like so please share ideas on how you think it should be done.

zeenix:
Perhaps something very simple as:

```
resource "libvirt_network" "net" {
...
  dnsmaq_option = {
     option_name1 = option_value1
     option_name2 = option_value2
  }
}
```

zeenix:
> Remember that we have also the `xlst` feature which let already implement without coding any libvirt functionality.

Right. I didn't think of that. Thanks! :) I'll close this now since if we can avoid adding support for workaround-API like this, that would be great. I'll try with XSLT and re-open if that doesn't work for some reason.

Thanks @MalloZup and @jaypoulz for suggestions.

jichenjc:
@zeenix just for learning reason , so we will try to include the xml file per 
https://github.com/dmacvicar/terraform-provider-libvirt/tree/master/examples/v0.12/xslt given so that we can inject dnsmasq configuration through the xml definition? Thanks

zeenix:
@jichenjc Yes. :)

jichenjc:
@zeenix do you have any info about this effort? Some PR or some issues in other Repo? I want to catch up with this .. thanks

zeenix:
> @zeenix do you have any info about this effort? Some PR or some issues in other Repo? I want to catch up with this .. thanks

If you mean the fix for https://github.com/openshift/installer/issues/1007, then no. I've only been trying out things by manually editting/hardcoding the XML for now. Also let's not discuss Openshift specific issues here. :)

jichenjc:
sure :) thanks for the reminder ~

PR #820, opened 2020-12-14
Support dnsmasq options

Prashanth684:
So this is implementing #642 using datasource as suggested in the issue. The issue was closed at that time because
of a workaround using xslt. But now we are seeing that this is needed for our Openshift CI for non-x86 arches which
runs on libvirt. This would enable us to automate this option through the openshift installer byproviding it through
the install-config at installation time.

### DIFF
--- a/libvirt/data_source_libvirt_network.go
+++ b/libvirt/data_source_libvirt_network.go
@@ -156,3 +156,58 @@ func resourceLibvirtNetworkDNSSRVRead(d *schema.ResourceData, meta interface{}) 
 
 	return nil
 }
+
+// a libvirt network dnsmasq template datasource
+//
+// Datasource example:
+//
+// data "libvirt_network_dnsmasq_options_template" "options" {
+//   count = length(var.libvirt_dnsmasq_options)
+//   option_name = keys(var.libvirt_dnsmasq_options)[count.index]
+//   option_value = values(var.libvirt_dnsmasq_options)[count.index]
+// }
+//
+// resource "libvirt_network" "k8snet" {
+//   ...
+//   dnsmasq_options = [{
+//     options = [ "${flatten(data.libvirt_network_dnsmasq_options_template.options.*.rendered)}" ]
+//   }]
+//   ...
+// }
+//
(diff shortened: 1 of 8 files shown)


MalloZup (approved):
seems good to me thx @Prashanth684 .

Prashanth684:
@MalloZup thanks for the quick review!  if the changes look good could these be merged please?

Prashanth684:
thank you so much @MalloZup !! just one follow up question - is there a plan to cut a `v0.6.4` tag sometime soon ? i'm asking because we plan to use these changes in the openshift installer.

MalloZup:
@Prashanth684  on the short-term for a 0.6.4 I think we might need to wait a bit.

Our focus is for a 0.7.0 with this mostly:  https://github.com/dmacvicar/terraform-provider-libvirt/pull/813

for a 0.6.4 I think if we have like 3/4 PRs with some changes , I think we might be able to release this year a minor version in case is needed. this is my pov, maybe @dmacvicar  has other thoughts, but I personally think a 0.6.4 in a short-term is not something we could be against. ( maybe we could have just the github release without pkg etc.).

Prashanth684:
> @Prashanth684 on the short-term for a 0.6.4 I think we might need to wait a bit.
> 
> Our focus is for a 0.7.0 with this mostly: #813
> 
> for a 0.6.4 I think if we have like 3/4 PRs with some changes , I think we might be able to release this year a minor version in case is needed. this is my pov, maybe @dmacvicar has other thoughts, but I personally think a 0.6.4 in a short-term is not something we could be against. ( maybe we could have just the github release without pkg etc.).

@MalloZup agreed. I would definitely be ok with a minor release without pkgs, just something to vendor in is what i am looking for.

BTW, #813 is interesting. i will read a bit more and see if i can contribute to it is some way..

MalloZup:
@Prashanth684  thx for your initiative and help. 

Check also the channel/chat where we are discussing it a bit:

https://gitter.im/terraform-provider-libvirt/Lobby

PR #828, opened 2021-01-27
Network config in examples for openSUSE should be 1, not 2

### LINKED ISSUE #827: Network is not setup for openSUSE following examples
### Linux distribution

openSUSE Leap 15.2 and Tumbleweed

### Terraform version

The issue is in the examples for opensuse in master: https://github.com/dmacvicar/terraform-provider-libvirt/tree/master/examples
___

## Checklist

### Setup

see network_config files for opensuse in https://github.com/dmacvicar/terraform-provider-libvirt/tree/master/examples

### Steps to Reproduce Issue

Network is not configured. /etc/sysconfig/network does not contain ifcfg-eth0 file.

However, for tumbleweed, firstboot setup creates the config, so the issue is not visible unless you use the same config with Leap or SLE.

___
## Additional information:

version should be 1 in https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/examples/v0.12/tubleweed/network_config.cfg#L2 .

See https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v1.html

jordimassaguerpla:
The syntax follows version 1 https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v1.html

The issue is that, if you use version 2, you don't get the network
configured. However, this is not visible because we are running the
firstboot setup which will set the network config.

The issue is visible if you use an image without firstboot setup
enabled.

fixes #827

### DIFF
--- a/examples/v0.11/leap15/network_config.cfg
+++ b/examples/v0.11/leap15/network_config.cfg
@@ -1,5 +1,5 @@
 network:
-  version: 2
+  version: 1
   config:
   - type: physical
     name: ens3
--- a/examples/v0.12/tubleweed/network_config.cfg
+++ b/examples/v0.12/tubleweed/network_config.cfg
@@ -1,5 +1,5 @@
 network:
-  version: 2
+  version: 1
   config:
   - type: physical
     name: ens3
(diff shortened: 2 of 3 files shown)


MalloZup (approved):

MalloZup:
@jordimassaguerpla  thank you for PR

PR #843, opened 2021-04-06
Including the domain Id in the error output, when it couldn't retrieve IP address

srbarrios:
This is just a small addition to the error output when it couldn't retrieve the IP address.

As, using terraform from our Jenkins pipelines, the direct output we can see in the Jenkins console is not the `log.Printf` but the `fmt.Errorf`. 
So I hope that adding the domain id in this output, even if it could smell duplicated, will help people in our situation to discover which is the domain failing in case they have a long list of VMs to deploy.


MalloZup (approved):

MalloZup:
thx!

Bischoff:
will help a lot! thanks

PR #845, opened 2021-04-14
more than 6 scsi devices on a single domain

M0WA:
this commit forces all scsi devices of a certain domain
to be attached to a single virtio scsi controller.
this eliminates the limit of 6 scsi devices per domain.
upper limit for this implementation is 256 scsi devices per vm.


M0WA:
will reopen with nicer code

PR #846, opened 2021-04-14
support more than 6 scsi devices on a single domain

M0WA:
this commit forces all scsi devices of a certain domain
to be attached to a single virtio scsi controller.
this eliminates the limit of 6 scsi devices per domain.
upper limit for this implementation is 256 scsi devices per domain.

some background information:
https://bugzilla.redhat.com/show_bug.cgi?id=1269577

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -476,6 +476,14 @@ func setDisks(d *schema.ResourceData, domainDef *libvirtxml.Domain, virConn *lib
 		if d.Get(prefix + ".scsi").(bool) {
 			disk.Target.Bus = "scsi"
 			scsiDisk = true
+			tmpInt := uint(0)
+			tmpIdx := uint(i)
+			disk.Address = &libvirtxml.DomainAddress{Drive: &libvirtxml.DomainAddressDrive{
+				Controller: &tmpInt,
+				Bus:        &tmpInt,
+				Target:     &tmpInt,
+				Unit:       &tmpIdx,
+			}}
 			if wwn, ok := d.GetOk(prefix + ".wwn"); ok {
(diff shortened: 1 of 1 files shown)


MalloZup (commented):
thx @M0WA  for this. Can you implement or update if possible the `testacc` for such change?

M0WA:
Hi @MalloZup 

i have added a new testacc for that.

M0WA:
hey @MalloZup,

does this actually have a chance to be merged? just wanna ask, i have no hurries :)

M0WA:
Hi @MalloZup, @dmacvicar,

can you say something, if/when it could get merged?

dmacvicar (changes requested):
Thanks for the PR. I had already looked into it but it took me a while to understand the problem.

I really dislike adding code that requires me or other contributors to dig into bugzilla comments of vendors to understand months later why a piece of code is there.

I can't make sense of the description, which mentions that attaching the disks to a single controller solves the problem.

* What does libvirt does if one does not explicitly attach it to a single controller?
* Why the problem seems to be more related to specifying `unit` in the disk address and to make it a `virtio` controller than the fact that it is a single controller?
* Then, why is Index being changed in the SCSI controller itself? According to the documentation:

> mandatory attribute index which is the decimal integer describing in which order the bus controller is encountered (for use in controller attributes of <address> elements). Since 1.3.5 the index is optional; if not specified, it will be auto-assigned to be the lowest unused index for the given controller type. 

Which looks totally unrelated.



I still don't feel comfortable with the description of this PR. It claims that because it attaches the disks to a single SCSI controller

dmacvicar (changes requested):
Waiting for changes

PR #873, opened 2021-08-13
Fix/qemuagent

maseman:
### Reimplements **qemu-guest-agent** interface query using go-libvirt interface.

Needed to query dhcp allocated IPs on bridged networks. 

### Testcase
I wanted to produce a decent reproducible test case so I generated an VM image (using alpine scripts) that runs qemu-guest-agent and configures a static IP that can be queried. See **TestAccLibvirtDomain_NetworkInterfaceQemuGuestAgentStaticIP**.

The VM test image in https://github.com/maseman/testvm-tflibvirt is generated with a workflow as a release asset. At ~100MB I thought it best NOT to add the file to terraform-provider-libvirt /testdata and instead setup a download in the testcase.

Alpine VMs like this (and with small overlay configurations) might be useful for other more complex test scenarios in the future.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/digitalocean/go-libvirt v0.0.0-20210615174804-eaff166426e3
 	github.com/fatih/color v1.10.0 // indirect
 	github.com/google/uuid v1.1.2
+	github.com/hashicorp/go-getter v1.4.0 // indirect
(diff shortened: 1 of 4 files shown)


dmacvicar:
I have manually rebased and committed this.

I don't like the 120M image for the test, so the test is pending and I opened #913 to revisit it.

PR #887, opened 2021-09-30
Add support for passwords using the SSH URI's

jli-cparta:
While a trivial addition, this allows using secret providers (e.g. Hashicorp Vault) to provide plaintext passwords instead of copying private key files around.

```
provider "vault" {
  address = "https://vault.example.com:8200"
}

data "vault_generic_secret" "admin_auth" {
  path = "op/vaults/servers/items/admin-at-kvm-server"
}

provider "libvirt" {
  uri = "qemu+ssh://admin:${data.vault_generic_secret.admin_auth.data["password"]}@kvm-server.example.com/system"
}
```

### DIFF
--- a/libvirt/helpers_test.go
+++ b/libvirt/helpers_test.go
@@ -59,7 +59,7 @@ func skipIfAccDisabled(t *testing.T) {
 
 func skipIfPrivilegedDisabled(t *testing.T) {
 	if os.Getenv("TF_LIBVIRT_DISABLE_PRIVILEGED_TESTS") != "" {
-		t.Skip("skipping test; Enviornemnt variable `TF_LIBVIRT_DISABLE_PRIVILEGED_TESTS` is set")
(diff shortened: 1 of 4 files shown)


dmacvicar:
Thanks for the PR!

[libvirt connection URI](https://libvirt.org/uri.html) does [not support password auth without agent](https://bugzilla.redhat.com/show_bug.cgi?id=1258726#c4) , but I think the way you implemented it is clever, because it does not break the spec.

It would still likely expose the password in the logs and output, would it?

jli-cparta:
> It would still likely expose the password in the logs and output, would it?

Not that I can find. It's in the `tfstate` files (of course). And if you enable debug-logging from terrafrom it will show up.

But it's not visible in standard output, nor is it in any log files. Not any that I can find, anyway. Searched everything in `/var/log/*` and the home directory of the user.

PR #889, opened 2021-10-05
Support specifying websocket port for VNC

klausenbusk:
Installing a VNC client isn't always a option. In these cases VNC can be
accessed over WebSocket with noVNC[1] in the browser.

[1] https://novnc.com


klausenbusk:
Pinging @dmacvicar per the [contributor guidelines](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/CONTRIBUTING.md).

Changing the `graphics` options doesn't change the XML (after the initial `terraform apply`), but it seems unrelated to this change.

PR #890, opened 2021-10-05
Support TPM devices

### LINKED ISSUE #888: Support TPM devices
## Checklist

  - [x] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.

`libvirt` [supports TPM devices](https://libvirt.org/formatdomain.html#tpm-device), either passthrough or a software TPM. It would be useful if the provider could add a TPM device to the domain. 

We have a custom Linux image requiring Secure Boot and a TPM device. The TPM device is used for unlocking a encrypted partition with systemd's [`crypttab`](https://www.freedesktop.org/software/systemd/man/crypttab.html). [Windows 11 is also requiring a TPM device](https://www.tomsguide.com/news/windows-11-tpm-requirements-microsoft-finally-clears-up-confusion), so a TPM device will be a hard requirement for booting Windows guests in the future.

klausenbusk:
"TPM (Trusted Platform Module) is a computer chip (microcontroller) that
can securely store artifacts used to authenticate the platform (your PC
or laptop). These artifacts can include passwords, certificates, or
encryption keys. A TPM can also be used to store platform measurements
that help ensure that the platform remains trustworthy. Authentication
(ensuring that the platform can prove that it is what it claims to be)
and attestation (a process helping to prove that a platform is
trustworthy and has not been breached) are necessary steps to ensure
safer computing in all environments."[1]

A TPM device is required by Windows 11 and it is also useful for testing
TPM support and can be used protecting secrets in guests.

[1] https://trustedcomputinggroup.org/resource/trusted-platform-module-tpm-summary/
[2] https://docs.microsoft.com/en-us/windows/whats-new/windows-11-requirements

Fix #888


dmacvicar:
In the next weeks.

PR #900, opened 2021-11-09
Support specifying "open" forward mode

klausenbusk:
"The new forward mode 'open' is just like mode='route', except that no
firewall rules are added to assure that any traffic does or doesn't
pass. It is assumed that either they aren't necessary, or they will be
setup outside the scope of libvirt."[1]

[1] https://github.com/libvirt/libvirt/commit/25e8112d7c32ab271b9cae28f3ccbf5835206693

---

This is a useful mode if you want libvirt to manage the ip address and DHCP/DNS server, but want to manage the firewall rules yourself (ex: if the firewall rules are too restrictive).


klausenbusk:
@dmacvicar could you have a look? :)

dmacvicar:
Thanks for the PR @klausenbusk 

Just a note: there is a `network_v2` resource in the making, and this will need to be forward ported.

PR #904, opened 2021-11-15
Add timeout for TLS connection URI

gxben:
Uses Go context to add a 2s dial timeout the same way it is done for TCP connections.

### DIFF
--- a/libvirt/uri/tls.go
+++ b/libvirt/uri/tls.go
@@ -1,6 +1,7 @@
 package uri
 
 import (
+	"context"
 	"crypto/tls"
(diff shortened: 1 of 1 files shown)


dmacvicar (changes requested):
Thanks for the PR.

Can you rebase against main? linter should fail now, as I added `gomnd`.

Then you can replace the timeout with `dialTimeout` const.

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #907, opened 2021-11-18
RBD volume support

gxben:
@dmacvicar This PR adds minimal changes to provide RBD volume support.
As documented, it relies on libvirt sysadmin to have provisioned template images on the Ceph/RBD pool and have the named pool defined at libvirt's level first, as to minimize configuration impact at Terraform level.

libvirt currently does not support source volume copying through backing store for anything else than directory pool type.
So when RBD volumes come into play, we use the volume clone and resize functions instead.

### DIFF
--- a/examples/v0.13/rbd/ubuntu-rbd-example.tf
+++ b/examples/v0.13/rbd/ubuntu-rbd-example.tf
@@ -1,0 +1,57 @@
+terraform {
+ required_version = ">= 0.13"
+  required_providers {
+    libvirt = {
+      source  = "gxben/libvirt"
+      version = "0.6.11"
+    }
+  }
+}
+
+# instance the provider
+provider "libvirt" {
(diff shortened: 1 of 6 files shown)


dmacvicar:
@gxben thanks for the PR.

I don't know much about RBD. Why do we need to put configuration in the disk, instead of having a pool of type RBD and have the disk refer to the volume and pool by name/id?

gxben:
Hi @dmacvicar ,
Care to detail which part of the "configuration in the disk" you mean so I can help ?

dmacvicar (changes requested):
All these `rbd_*` attributes in the disk element. Why are these not in the pool definition?

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #909, opened 2021-11-24
 Add multi-hosts support.

gxben:
This allows specifying multiple hosts (and so associated libvirt connections) to the same provider and tagging them by region and/or availability zones.
This way, it becomes possible to iterate resources over multiple libvirt instances as Terraform does not yet support iteration over providers.

### DIFF
--- a/libvirt/config.go
+++ b/libvirt/config.go
@@ -3,6 +3,7 @@ package libvirt
 import (
 	"fmt"
 	"log"
+	"math/rand"
 	"sync"
 
 	libvirt "github.com/digitalocean/go-libvirt"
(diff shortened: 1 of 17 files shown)


dmacvicar:
Sorry. I don't feel comfortable with this PR. I never heard the concept of Availability Zone and Region in libvirt.

What does this solve vs https://www.terraform.io/docs/language/providers/configuration.html#alias-multiple-provider-configurations

gxben:
There's no such thing as AZ or Region in libvirt indeed.
The thing is that libvirt only provides host-level control and has no top-layer orchestration layer that would allow control of a cluster of libvirt hosts.
This PR somehow simulates this concept and comes as a workaround to have multiple libvirt hosts detailed in a single libvirt provider, because TF doesn't allow to iterate over provider.

So, let's imagine the following scenario: you have multiple libvirt hosts in multiple datacenters or locations and you want to create VMs in each having some kind of host-affinity of high-availability.
The only way currently offered by Terraform is to copy/paste your VMs TF configuration for each libvirt provider (regardless of the use of aliases) which is just nuts.

So this PR allows for having a single codebase that one can use for_each to target multiple libvirt hosts within a single provider.

dmacvicar:
Thanks for the PR, but to me this introduces concepts which are alien to both libvirt and terraform and do not belong in the terraform provider layer and I don't feel like maintaining this complexity.

Not sure how to implement something like this, perhaps a libvirt API proxy?

PR #927, opened 2022-02-04
Examples update

shafer:
This PR removes the CoreOS example and replaces it with a working/tested one for Fedora CoreOS. I verified with and without qemu agent.


dmacvicar:
Thanks! :partying_face:

PR #933, opened 2022-02-17
SSH config support

jbeisser:
This adds native SSH configuration support, with some minor caveats:

- It's not perfect. OpenSSH and Kevin Burke's ssh_config library behave slightly differently in compiling results. OpenSSH wants least to most specific, Burke's library wants most specific to least.
- It's ugly, but it does work. The `ssh.go` dialer could use some additional love.
- It also supports alternate ports in the SSH configuration file, but not very intelligently.

This DOES allow you to recycle configurations in the provider section, without setting a static username on the URI. So, if users are managed via PAM or LDAP access to the socket, it should just work.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.24.1
 	github.com/hooklift/iso9660 v1.0.0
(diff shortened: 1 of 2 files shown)


jbeisser:
Updated to reflect other changes in 0.7.0.

pstrzelczak:
@jbeisser could you resolve conflicts here? @dmacvicar it is cumbersome to be forced to pass extra parameters to uri.

dmacvicar:
thanks :pray:

PR #935, opened 2022-02-22
Removed dependency towards mkisofs system binary

### LINKED ISSUE #934: libvirt_cloudinit_disk does not work on Terraform Cloud
### Linux distribution

Latest version of Ubuntu LTS as mentioned [here](https://www.terraform.io/cloud-docs/run/install-software#installing-additional-tools)

### Terraform version

1.1.5

Latest available from Hashicorps Terraform Provider registry

___

## Checklist

  - [x] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.
  - [x] If it is a very special case, consider using the _XSLT_ support in the provider to tweak the definition instead of opening an issue
  - [x] Maintainers do not have expertise in every libvirt setting, so please, describe the feature and how it is used. Link to the appropriate documentation

- [x] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

### Setup

```terraform
terraform {
  required_providers {
    libvirt = {
      source = "dmacvicar/libvirt"
    }
  }
}

provider "libvirt" {
  uri = "qemu+ssh://virt:${var.ssh_pass}@THEIP/system?sshauth=ssh-password&knownhosts=./known_hosts"
}

locals {
    libvirt = {
        pools = {
            vm = {
                name = "virtualMachines"
            }
        }
    }
}

resource "libvirt_pool" "virtualMachines" {
    name = local.libvirt.pools.vm.name
    type = "dir"
    path = "/aPath/${local.libvirt.pools.vm.name}"
}

  template = file("${path.module}/cloud_init.cfg")
}

data "template_file" "network_config" {
  template = file("${path.module}/network_config.cfg")
}

  name           = "commoninit.iso"
  user_data      = data.template_file.user_data.rendered
  network_config = data.template_file.network_config.rendered
  pool           = libvirt_pool.virtualMachines.name
}
```

### Steps to Reproduce Issue

Attempt to create a libvirt_cloudinit_disk resource using Terraform running on Hashicorp managed Terraform Cloud runners.

___
## Additional information:

The call to an external dependency seems to happen [here](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/cloudinit_def.go#L148).

According to Terraform Cloud, one should not install additional tools on their runners - While it is supported, it should only be as a last resort. This is documented [here](https://www.terraform.io/cloud-docs/run/run-environment#the-terraform-worker-vms) and [here](https://www.terraform.io/cloud-docs/run/install-software#installing-additional-tools).

While this does not make the provider entirely incompatible with Terraform Cloud, it sure does make it very tiresome to work with.

See thee attached screenshot for exact error message
Screenshot 2022-02-22 at 11 46 59

MadsRC:
This fixes #934

It introduces a new dependency that replaces the need for the external `mkisofs` binary.
I'd also like for it to replace the `hooklift/iso9660` dependency, but that would potentially require changing the signature of a few functions (setCloudInitDataFromExistingCloudInitDisk takes an os.File, where go-diskfs prefers to handle the files itself).

Before making considerable changes, I'd like to ask the current maintainers what they prefer be done?

### DIFF
--- a/contrib/Debian/Dockerfile_all_in_one
+++ b/contrib/Debian/Dockerfile_all_in_one
@@ -67,13 +67,13 @@ COPY --from=terraform /bin/terraform /bin/
 COPY --from=libvirt /go/bin/terraform-provider-libvirt /root/.terraform.d/plugins/${GO_OS}_${GO_ARCH}/
 
 # Install Dependencies
-# libvirt0 is needed to run the provider. xsltproc needed to use XML/XSLT. mkisofs needed to use cloud init images
+# libvirt0 is needed to run the provider. xsltproc needed to use XML/XSLT.
 # ca-certificates to avoid terraform init 509 error. openssh-client to talk to remote libvirt server
 RUN apt-get update \
     && apt-get upgrade -y \
     && apt-get install \
     -y --no-install-recommends \
-    libvirt0 xsltproc mkisofs ca-certificates openssh-client \
+    libvirt0 xsltproc ca-certificates openssh-client \
     && rm -rf /var/lib/apt/lists/*
 
 # Copy Terraform Files
--- a/contrib/Debian/Dockerfile_build_dependent
+++ b/contrib/Debian/Dockerfile_build_dependent
@@ -32,13 +32,13 @@ COPY --from=terraform /bin/terraform /bin/
 COPY --from=libvirt /go/bin/terraform-provider-libvirt /root/.terraform.d/plugins/${GO_OS}_${GO_ARCH}/
 
 # Install Dependencies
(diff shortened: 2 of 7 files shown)


dmacvicar:
Would it make sense to use this library to generate VFAT images and also switch to `virtio` devices instead of CDROM? It would at least address the pitfalls raised in https://github.com/dmacvicar/terraform-provider-libvirt/pull/895

jbeisser:
I need to test some things, but under libvirt 4.5 on CentOS 7 I had some issues with vfat being hooked up correctly. 

It may be due to missing parameters.

PR #960, opened 2022-07-09
Make 'option_value' for dnsmasq  optional

e4t:
To support value-less dnsmasq options ('no-hosts') this change set makes `option_value` optional in which case `option_name` is inserted in the dnsmasq.conf without traling '='. 
While this change invalidates the separation into 'name' and 'value' to some extend, it retains compatibility.
An update to the documentation and test cases are also included.

### DIFF
--- a/libvirt/data_source_libvirt_network.go
+++ b/libvirt/data_source_libvirt_network.go
@@ -185,7 +185,7 @@ func datasourceLibvirtNetworkDnsmasqOptionsTemplate() *schema.Resource {
 			},
 			"option_value": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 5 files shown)


e4t:
Hi @dmacvicar could you have a look at this? Thanks!

dmacvicar:
I will look into this after I merge the SDK v2 update. Thanks.

dmacvicar:
Thanks for the PR

e4t:
> Thanks for the PR

@dmacvicar Thanks for merging! - A belated Xmas present ;)
... and Happy New Year to you!

PR #974, opened 2022-09-09
modify the volume size type to support units

rgl:
This allows us to specify the volume size as:

```hcl
resource "libvirt_volume" "example_root" {
  size = "123GiB"
}
```

Currently, the units are case-insensitive, should they be case-sensitive?

Maybe also expand this to the other size/capacity properties like the memory size?

Please note that this is still missing the documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,11 @@
-LDFLAGS += -X main.version=$$(git describe --always --abbrev=40 --dirty)
+HOSTNAME=registry.terraform.io
+NAMESPACE=dmacvicar
+NAME=libvirt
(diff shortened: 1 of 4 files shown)


dmacvicar:
I am not convinced about this change, as it changes the type to string. Is it backward compatible?

Also, there is some push to have this in HCL itself:
https://github.com/hashicorp/hcl/issues/326

rgl:
I think we can still use `size = 123`. But I can confirm if this can go forward.

dmacvicar:
Lets wait for native support in https://github.com/hashicorp/hcl/issues/326

PR #983, opened 2022-11-17
Fix IPv6 subnet size regression

omertuc:
b0489a5cb128b6acae1976e0383e22bf3d870277 introduced a regression where
the calculation for the number of IPv6 IP addresses always yields a
negative or 0 value, causing users to *always* encounter the following
error when creating IPv6 libvirt networks:

`netmask seems to be too strict: only <0 or negative> IPs available (ipv6)`

(An example is available at the end of [this](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-test-infra/1883/pull-ci-openshift-assisted-test-infra-master-e2e-metal-assisted-ipv6/1583020018084352000/artifacts/e2e-metal-assisted-ipv6/assisted-common-test/build-log.txt) CI log)

That commit attempted to fix the wrong use of the `^` operator in the
calculation, which was truely wrong. But it was just wrong in a
relatively "harmless" way, as it wasn't completely blocking users.

The fix in that commit had its own bug - a `1` shifted by `128` always
gives `0`, and not the desired `2 to the power of 128`, because the
latter doesn't fit in a primitive integer type.

To fix this, I've changed the calculation to simply consider the number
of bits available for the subnet, rather than the number of IP addresses
available for the subnet, as that is obviously a much smaller number,
one that the primitive Go integer types can handle

### DIFF
--- a/libvirt/network.go
+++ b/libvirt/network.go
@@ -115,9 +115,15 @@ func getNetworkIPConfig(address string) (*libvirtxml.NetworkIP, *libvirtxml.Netw
 	if bits == (net.IPv6len * 8) {
 		family = "ipv6"
 	}
-	ipsRange := (1 << bits) - (1 << ones)
-	if ipsRange < 4 {
-		return nil, nil, fmt.Errorf("netmask seems to be too strict: only %d IPs available (%s)", ipsRange-3, family)
+
+	const minimumSubnetBits = 3
+	if subnetBits := bits - ones; subnetBits < minimumSubnetBits {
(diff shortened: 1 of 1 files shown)


osherdp:
@dmacvicar just FYI as it's currently a blocker for our usage of the new version v0.7.0

dmacvicar:
Thanks for the patch @omertuc

dmacvicar:
This is included in 0.7.1

osherdp:
Thank you very much!
I'll test it right away

PR #1014, opened 2023-04-24
Default machine

e4t:
The default machine types picked by libvirt on non-x86* architectures are often quite limited and not very useful. For compatibility reasons libvirt picks the machine type that was the first one support for this architecture. virt-installer therefore maintains its own settings.
This patchset implements the same setting for libvirt if the user does not explicitly specify a machine type.

@dmacvicar - could you have a look?

### DIFF
--- a/libvirt/domain_def.go
+++ b/libvirt/domain_def.go
@@ -139,8 +139,11 @@ func newDomainDefForConnection(virConn *libvirt.Libvirt, rd *schema.ResourceData
 
 	if machine, ok := rd.GetOk("machine"); ok {
 		d.OS.Type.Machine = machine.(string)
-	} else if len(guest.Arch.Machines) > 0 {
-		d.OS.Type.Machine = guest.Arch.Machines[0].Name
+	} else {
+		d.OS.Type.Machine, err = getMachineTypeForArch(caps, d.OS.Type.Arch, d.OS.Type.Type)
+		if err != nil {
+			return d, err
+		}
 	}
 
 	canonicalmachine, err := getCanonicalMachineName(caps, d.OS.Type.Arch, d.OS.Type.Type, d.OS.Type.Machine)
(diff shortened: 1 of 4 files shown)


dmacvicar:
Thanks. I am unsure if this is a good idea. The default for the setting would not longer be depending on the schema. Why is this default not improved in libvirt itself?

e4t:
> Thanks. I am unsure if this is a good idea. The default for the setting would not longer be depending on the schema. Why is this default not improved in libvirt itself?

The machine type set in libvirt is the one that was first implemented for this architecture. The argument for not changing it to a more capable machine type would break libvirt's backward compatibility guarantees [see this email](https://listman.redhat.com/archives/libvir-list/2023-March/238809.html)
As mentioned in this email, the issue has therefore been addressed on a higher level -[ ie 'virt-install'](https://github.com/virt-manager/virt-manager/blob/33ff193ee9fcfdb74f95d946a1b93239a1a12a61/virtinst/guest.py#L152-L180). 
I would argue that this terraform provider is on the same level as 'virt-install' as it replaces it.

The fix  avoids unpleasant surprises if the same terraform module that works on x86_64 fails horrifically on aarch64. 
The user is still free to set the machine type explicitly. This is optional, however, in which case the default is what has been set elsewhere.

beddari:
This is a great change and would be in line with what most users expect. Also, /when/ they see the non-obvious default behaviour of libvirt, they tend to not know or find out about how to deal with this in a general way.

+1.

dmacvicar:
Ok. I am convinced, but please add a comment explaining what we are doing above the call to `getMachineTypeForArch`, as nobody reading the code will come back to this PR.

e4t:
> Ok. I am convinced, but please add a comment explaining what we are doing above the call to `getMachineTypeForArch`, as nobody reading the code will come back to this PR.

Done. Please check if it fits.

dmacvicar:
Thanks :partying_face:

e4t:
@dmacvicar - Thank you!

PR #1030, opened 2023-09-11
Fix malformed connection remote name when using ssh remote uri.

tiaden:
Fixes an issue where the remote uri qemu+ssh://username@hostname:2222/system?foo=bar&bar=foo does not resolve to qemu:///system as remote name


michaelbeaumont (approved):

michaelbeaumont:
Thanks!

PR #1042, opened 2023-10-24
feat: add data sources to extract node and device information

muresan:
Hello,

this PR adds support for new data sources:

* `libvirt_node_info` - to retrieve node information and make it available to Terraform. (`virsh nodeinfo`)
* `libvirt_node_devices` - retrieves the list of devices available in the node, can be filtered by capability. (`virsh nodedev-list`)
* `libvirt_node_device_info` - retrieves information about a single device. (`virsh nodedev-dumpxml`)

Apart from the changes to `docs` and `provider.go` to plugin in the documentation and the new resources, changes are in new files.


libvirt/data_source_libvirt_node_info.go
```diff
@@ -0,0 +1,91 @@
...
+	d.Set("cpu_sockets", sockets)
+	d.Set("cpu_threads_per_core", threads)
+	d.Set("numa_nodes", nodes)
+	d.Set("memory_total_kb", memory)
+	d.SetId(strconv.Itoa(hashcode.String(fmt.Sprintf("%v%v%v%v%v%v%v", model, memory, cpus, nodes, sockets, cores, threads))))
+
+	return nil
+}
+
+func Int8ToString(bs [32]int8) string {
```
rustydb, line 85: `Int8ToString` doesn't need to be exported since it is only used within this `package`, it could be made private to the package.

```suggestion
	d.Set("cpu_model", int8ToString(model))
	d.Set("cpu_cores_total", cpus)
	d.Set("cpu_cores_per_socket", cores)
	d.Set("numa_nodes", nodes)
	d.Set("cpu_sockets", sockets)
	d.Set("cpu_threads_per_core", threads)
	d.Set("numa_nodes", nodes)
	d.Set("memory_total_kb", memory)
	d.SetId(strconv.Itoa(hashcode.String(fmt.Sprintf("%v%v%v%v%v%v%v", model, memory, cpus, nodes, sockets, cores, threads))))

	return nil
}

func int8ToString(bs [32]int8) string {
```

libvirt/data_source_libvirt_node_device_info.go
```diff
@@ -0,0 +1,183 @@
...
+
+	var device_xml Device
+
+	err = xml.Unmarshal([]byte(xml_desc), &device_xml)
+	if err != nil {
+		log.Fatalf("failed to unmarshal device_xml into XML: %v", err)
+	}
+
+	capability := map[string]interface{}{}
+	if device_xml.Capability.Type == "pci" {
```
michaelbeaumont, line 143: Any reason we don't support [all the `Type` options](https://libvirt.org/formatnode.html#capability-types) from the get go? Seems fine to only support a select few at first, just curious.

libvirt/data_source_libvirt_node_device_info.go
```diff
@@ -0,0 +1,183 @@
...
+		capability["type"] = device_xml.Capability.Type
+		capability["class"] = device_xml.Capability.Class
+		capability["domain"] = fmt.Sprintf("%d", device_xml.Capability.Domain)
+		capability["bus"] = fmt.Sprintf("%d", device_xml.Capability.Bus)
+		capability["slot"] = fmt.Sprintf("%d", device_xml.Capability.Slot)
+		capability["function"] = fmt.Sprintf("%d", device_xml.Capability.Function)
+		capability["product_id"] = device_xml.Capability.Product.Id
```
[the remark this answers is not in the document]
muresan: I've tried to find an example with a map in a map but best I could find is list of map, so the outcome is something like:
```
    capability = [
        {
            bus     = "0"
            class   = "0x060000"
            domain  = "0"
            product = [
                {
                    id   = "0x1480"
                    name = "Starship/Matisse Root Complex"
                },
            ]
            type    = "pci"
        },
    ]
```
which ends up as capability.0.product.0.id. Ugly, but if you can show me an example from somewhere on how to generate the TypeMap in a TypeMap I'd appreciate it.
I agree with your suggestion, I'm still looking into how to implement it.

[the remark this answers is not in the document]
muresan: After many hours investigating options, it seems Terraform does not support nested maps, so list(map(list(map) is the best option.
With the exception of a couple of small changes the attributes from XML are mapping directly to terraform attributes.

libvirt/data_source_libvirt_node_device_info.go
muresan: Added the rest of relevant devices, with the exception of S390 hardware which, if someone has access to, can add.

website/libvirt.erb
```diff
@@ -30,6 +30,21 @@
             </li>
           </ul>
         </li>
+        <li<%= sidebar_current("docs-libvirt-data-source") %>>
+          <a href="#">Data Sesources</a>
```
michaelbeaumont, line 34:
```suggestion
          <a href="#">Data Resources</a>
```

michaelbeaumont (approved):
Thanks!

dmacvicar:
Still puzzled how the test ever passed:

```
=== RUN   TestAccLibvirtNodeInfoDataSource
    data_source_libvirt_node_info_test.go:11: Step 1/2 error: Check failed: Check 1/1 error: data.libvirt_node_info.info: Attribute 'cpus' didn't match "^\\d+", got ""
-```

```
	d.Set("cpu_model", int8ToString(model))
	d.Set("cpu_cores_total", cpus)
	d.Set("cpu_cores_per_socket", cores)
	d.Set("numa_nodes", nodes)
	d.Set("cpu_sockets", sockets)
	d.Set("cpu_threads_per_core", threads)
	d.Set("numa_nodes", nodes)
	d.Set("memory_total_kb", memory)
```
Test tests for an attribute cpu, which has never been set. It is clear integration tests were not run, and shows the need to get them to run in CI with nested virtualization.

PR #1044, opened 2023-11-03
Fix copy of the URL in RemoteName

### LINKED ISSUE #1040: version 0.7.4 breaks qemu+ssh (libvirt-sock) connection
##  System Information
amd64

### Linux distribution
Ubuntu22.04

### Terraform version
v1.6.1

0.7.4

If that gives you "was not built correctly", get the Git commit hash from your local provider repository:
na
___
I use the plugin to provision vm's on rack within my network.
I always implement an install with the latest Terraform and plugin versions.
With version 0.7.4. is get the following error when executing my code:
 Error: failed to connect: dial unix /var/run/libvirt/libvirt-sock: connect: permission denied
│ 
│   with provider["registry.terraform.io/dmacvicar/libvirt"],
│   on connection.tf line 14, in provider "libvirt":
│   14: provider "libvirt" {
 
Falling back to version 0.7.1 (since it is the latest version that is useable): it works again.

### Setup
my connection.tf not working
```
terraform {
 required_version = ">= 1.6.1"
  required_providers {
    libvirt = {
      version  = "0.7.4"
      source  = "dmacvicar/libvirt"
    }
  }
}


# instance the provider
provider "libvirt" {
  uri = "qemu+ssh://user@xxx.xxx.hosts/system?keyfile=/home/ruben/.ssh/xxx&sshauth=privkey"
}
```

my connection.tf working
```
terraform {
 required_version = ">= 1.6.1"
  required_providers {
    libvirt = {
      version  = "0.7.1"
      source  = "dmacvicar/libvirt"
    }
  }
}


# instance the provider
provider "libvirt" {
  uri = "qemu+ssh://user@xxx.xxx.hosts/system?keyfile=/home/ruben/.ssh/xxx&sshauth=privkey"
}
```

X-Cli:
Fixes #1040

The copy performed in RemoteName was shallow and this corrupted the original URI that is used during the Dial() call made by the go-libvirt library.

The issue always existed but did not manifest when libvirt.New(conn) was used because the connection was established before the call to RemoteName. The problem manifested when the switch to libvirt.NewWithDialer was made, because the dialing is now made during the call to ConnectToURI, which receives as paramater the result of the call to RemoteName.

### DIFF
--- a/libvirt/uri/connection_uri.go
+++ b/libvirt/uri/connection_uri.go
@@ -31,13 +31,13 @@ func Parse(uriStr string) (*ConnectionURI, error) {
 // The name passed to the remote virConnectOpen function is formed by removing
 // transport, hostname, port number, username and extra parameters from the remote URI
 // unless the name option is specified.
-func (u *ConnectionURI) RemoteName() string {
+func (u ConnectionURI) RemoteName() string {
 	q := u.Query()
 	if name := q.Get("name"); name != "" {
 		return name
 	}
 
-	newURI := *u
+	newURI := *u.URL
 	newURI.Scheme = u.driver()
 	newURI.User = nil
 	newURI.Host = ""


libvirt/uri/connection_uri.go
```diff
@@ -31,17 +31,16 @@ func Parse(uriStr string) (*ConnectionURI, error) {
 // The name passed to the remote virConnectOpen function is formed by removing
 // transport, hostname, port number, username and extra parameters from the remote URI
 // unless the name option is specified.
-func (u *ConnectionURI) RemoteName() string {
+func (u ConnectionURI) RemoteName() string {
```
michaelbeaumont, line 34: This is better style but ultimately does nothing, since the copied `u` value still contains a pointer to the same `url.URL` value.

libvirt/uri/connection_uri.go
```diff
@@ -31,17 +31,16 @@ func Parse(uriStr string) (*ConnectionURI, error) {
...
 
-	newURI := *u
-	newURI.Scheme = u.driver()
-	newURI.User = nil
-	newURI.Host = ""
-	newURI.RawQuery = ""
+	newURI := url.URL{
+		Scheme: u.driver(),
+		Path:   u.Path,
+	}
```
michaelbeaumont, line 43:
```suggestion
    newURI := *u.URL
	newURI.Scheme = u.driver()
	newURI.User = nil
	newURI.Host = ""
	newURI.RawQuery = ""
```

michaelbeaumont: @X-Cli This should be a proper shallow copy of the `url.URL`, could you see if that also fixes the issue for you?

X-Cli: @michaelbeaumont This suggestion works too. My feeling is that building a new clean instance, defining only the required fields is less likely to lead to surprises than copying all data then setting the false value on the unnecessary fields. It is also more concise. But ultimately, both suggestions work :)

michaelbeaumont: I would generally agree _but_ [the "spec" explicitly defines the remote name by what is removed from the connection URI](https://libvirt.org/uri.html#transport-configuration)

X-Cli: Fair point. Thank you for your comments!

michaelbeaumont (approved):
@X-Cli sorry, I didn't mean to say we shouldn't include the pointer receiver change at all, I think it's better with `(u ConnectionURI)`. feel free to add it back! If not I'll merge this in a few minutes

michaelbeaumont:
Thanks @X-Cli!

PR #1046, opened 2023-11-05
Fix firmware for armv7l

thequailman:
armv7l needs the same firmware as aarch64:

From a vanilla arm libvirt config:
```xml
<domain type="qemu">
  <os firmware="efi">
    <type arch="armv7l" machine="virt-7.2">hvm</type>
    <boot dev="hd"/>
  </os>
```


michaelbeaumont:
Thanks @thequailman

I'm not familiar with why even aarch64 needs `"efi"`. Is it about a sensible default or is it impossible to run without `firmware="efi"`?

Issue #1047, opened 2023-11-06
Waiting for lease finishes prematurely if domain has network interface WITHOUT IP addess assigned yet.


pstrzelczak:
### Linux distribution

```Rocky 8.8 ```

### Terraform version

```
2023-11-06T10:41:59.199Z [INFO]  Terraform version: 1.6.2
2023-11-06T10:41:59.199Z [DEBUG] using github.com/hashicorp/go-tfe v1.36.0
2023-11-06T10:41:59.199Z [DEBUG] using github.com/hashicorp/hcl/v2 v2.19.1
2023-11-06T10:41:59.199Z [DEBUG] using github.com/hashicorp/terraform-svchost v0.1.1
2023-11-06T10:41:59.199Z [DEBUG] using github.com/zclconf/go-cty v1.14.1
2023-11-06T10:41:59.199Z [INFO]  Go runtime version: go1.21.1
2023-11-06T10:41:59.199Z [INFO]  CLI args: []string{"terraform", "--version"}
2023-11-06T10:41:59.200Z [DEBUG] Attempting to open CLI config file: /home/docker_user/.terraformrc
2023-11-06T10:41:59.200Z [DEBUG] File doesn't exist, but doesn't need to. Ignoring.
2023-11-06T10:41:59.200Z [DEBUG] ignoring non-existing provider search directory terraform.d/plugins
2023-11-06T10:41:59.200Z [DEBUG] ignoring non-existing provider search directory /home/docker_user/.terraform.d/plugins
2023-11-06T10:41:59.200Z [DEBUG] ignoring non-existing provider search directory /home/docker_user/.local/share/terraform/plugins
2023-11-06T10:41:59.200Z [DEBUG] ignoring non-existing provider search directory /usr/local/share/terraform/plugins
2023-11-06T10:41:59.200Z [DEBUG] ignoring non-existing provider search directory /usr/share/terraform/plugins
2023-11-06T10:41:59.200Z [INFO]  CLI command args: []string{"version", "--version"}
Terraform v1.6.2
on linux_amd64
+ provider registry.terraform.io/dmacvicar/libvirt v0.7.1
+ provider registry.terraform.io/hashicorp/null v3.2.1

Your version of Terraform is out of date! The latest version
is 1.6.3. You can update by downloading from https://www.terraform.io/downloads.html

```

### Provider and libvirt versions

```
v0.7.1
```
___

### Setup

[...]

PR #1039, opened 2023-10-14
Retry the polling when guest agent not ready

rgl:
Closes https://github.com/dmacvicar/terraform-provider-libvirt/issues/1037

PR #1048, opened 2023-11-07
Make `domainIfaceHasAddress` return true only if there is at least one IP address assigned to a network interface with matching physical address.

pstrzelczak:
Fixes https://github.com/dmacvicar/terraform-provider-libvirt/issues/1047


pstrzelczak:
@dmacvicar could you review this PR? Also, once merged could you release new version as without this and one more pending [contribution](https://github.com/dmacvicar/terraform-provider-libvirt/pull/1039) current plugin is unusable.

dmacvicar:
Thanks for the patch. :pray:

PR #1053, opened 2023-12-14
Qemu Agent Enhancement

tiaden:
The idea of this PR is to make sure that the guest qemu agent is running by explicitly query it. This will make ip retrieval for example more reliable. let me know what you think

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -602,6 +602,18 @@ func resourceLibvirtDomainCreate(ctx context.Context, d *schema.ResourceData, me
 	d.SetId(id)
 	log.Printf("[INFO] Domain ID: %s", d.Id())
 
+	if d.Get("qemu_agent").(bool) && d.Get("running").(bool) {
+		// waiting for qemu agent to be available
+		err := waitingForAgentRunning(ctx, virConn, domain, d.Timeout(schema.TimeoutCreate), d)
+		if err != nil {
+			agentNotFound := "Please make sure that qemu-agent is installed \n" +
+				"IMPORTANT: This error is not a terraform libvirt-provider" +
+				" error, but an error caused by your KVM/libvirt" +
+				" infrastructure configuration/setup"
+			return diag.Errorf("couldn't connect to the qemu agent of the domain id: %s. %s \n %s", d.Id(), agentNotFound, err)
+		}
+	}
+
 	if len(waitForLeases) > 0 {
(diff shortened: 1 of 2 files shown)


mleone87:
Checked this PR locally and it works! Upstream code with `bridged network` and `qemu_agent = true `does not work since it fails at t=0 without retries. You have to run `terraform refresh` some time to make it discover the IP ad have a corret tfstate

@dmacvicar this should be merged I guess

tiaden:
@dmacvicar, any chance of getting this merged ?

libvirt/resource_libvirt_domain.go
```diff
@@ -602,6 +602,18 @@ func resourceLibvirtDomainCreate(ctx context.Context, d *schema.ResourceData, me
 	d.SetId(id)
 	log.Printf("[INFO] Domain ID: %s", d.Id())
 
+	if d.Get("qemu_agent").(bool) && d.Get("running").(bool) {
```
dmacvicar, line 605: Why does the code check for `d.Get("qemu_agent")` if then `waitingForAgentRunning` checks again?

dmacvicar, line 605: Why does the code check for d.Get("qemu_agent") if then waitingForAgentRunning checks again?

libvirt/resource_libvirt_domain.go
```diff
@@ -769,6 +781,18 @@ func resourceLibvirtDomainUpdate(ctx context.Context, d *schema.ResourceData, me
 		}
 	}
 
+	if d.Get("qemu_agent").(bool) && d.Get("running").(bool) {
```
dmacvicar, line 784: Why is this wait executed twice?

dmacvicar (changes requested):
Thanks for the PR :pray: 

I think it makes sense to merge some kind of check. However, I'd need to understand the level of defensiveness of the code (eg. checking for `qemu_agent` inside and outside of the function, and running the same block twice). This code smells like "just in case".

tiaden:
@dmacvicar, I replied to your comments. As of the level of defensiveness (eg. checking for qemu_agent inside and outside of the function), it's just my style of programming. But you are right. what do you prefer ? keeping the check inside or outside the function ? as for your comment about running the same block twice, as I explained in the comment, those blocks happened to be inside different functions

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #1055, opened 2023-12-17
ssh improvements

MaxMatti:
* fix ipv6 handling for non-default ports
* add ssh config support for the following:
  * hostname
    * if user-specified uri points to ssh config entry with hostname setting, use that
  * port
    * user-specified port > config-specified port > default port

### DIFF
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -145,7 +145,7 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
 		port = defaultSSHPort
 	}
 
-	sshClient, err := ssh.Dial("tcp", fmt.Sprintf("%s:%s", u.Hostname(), port), &cfg)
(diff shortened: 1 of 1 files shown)


libvirt/uri/ssh.go
```diff
@@ -133,6 +133,11 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
 		username = sshu
 	}
 
+	hostname, err := sshcfg.Get(u.Hostname(), "HostName")
```
zaneb, line 136: `Get()` doesn't return an error when the key is not found, so this will result in the hostname ~never being read from the URI directly (i.e. it will _only_ work when there is an alias in `~/.ssh/config`).

And yes, the username thing immediately above is also broken.

memetb: fyi, #1059 fixes all of these

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

Issue #1058, opened 2023-12-20
[FEATURE] Add more comprehensive support for ssh_config, specifically bastion hosts


memetb:
### Provider and libvirt versions `9adc2521e37b0c8938330bf81111d3c633825c76-dirty`
___

Hello @dmacvicar, I've coded a new feature for the project and am opening an issue prior to submitting a WIP PR.

I have updated the code for [`ssh.go`](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/uri/ssh.go) to add more thorough support for making use of `ssh_config` parameters.

This code change resolves the following specific and important issues: 
- inability to use a identity file other than `~/.ssh/id_rsa`
- inability to use bastion hosts as they are defined in the ssh_config file

This feature requires upgrading `ssh_config` module to v1.2.0 which is the as of now current module (is currently at v0.0.0 in the master branch). Other than this, all changes are confined to `ssh.go`.

SimonWoidig:
This is something that we would love to see aswell - especially the proxies.

scabala:
since #1059 was merged, I think this can be closed

dmacvicar:
and 0.8.0 is released

PR #1059, opened 2023-12-20
expanded ssh_config parameters for qemu+ssh uri option

memetb:
Hello @dmacvicar,

this PR is in relation to the issue #1058 I recently opened.

~~This is a preliminary PR and work in progress. I have listed the known issues with this PR at this point. I would also like to get feedback from you before putting more effort in.~~

This feature is a must-have for my use case since I require to access my bare metal servers through a bastion host.

Known issues:
1. using ProxyJump with openssh implementation of ssh will ask the bastion host to resolve the HostName. This allows for local resolution (e.g. private and/or dynamic IP addresses) to be used for the hostname resolution (i.e. a machine may have a name that is only locally resolvable on the bastion). The current code will not forward the name resolution from the bastion host.
2. there is a hardcoded maximum depth of 10 jump hosts
~~3. little to no effort on code style and documentation: this was a quick and dirty feature feasibility test to see if the desired workflow worked correctly~~
4. no unit tests have been added


libvirt/uri/ssh.go
```diff
@@ -91,74 +125,184 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
...
+		}
 	}
 
-	address := q.Get("socket")
-	if address == "" {
-		address = defaultUnixSock
+	if cfg.User == "" {
+		sshu, err := sshcfg.Get(target, "User")
+		log.Printf("[DEBUG] SSH User for target '%v' is '%v'", target, sshu)
+		if err != nil {
```
memetb, line 263: This may be inverted or should be checking `sshu != nil`.

tuxillo:
will this ever be merged?

memetb:
@tuxillo fyi, I'm using my forked project in production. I haven't published that project on terraform specifically because I want to keep the noise ratio down, but if you so wish, you can use the forked project's production branch which has this merged as well as a couple of other small features merged in.

gthieleb:
> @tuxillo fyi, I'm using my forked project in production. I haven't published that project on terraform specifically because I want to keep the noise ratio down, but if you so wish, you can use the forked project's production branch which has this merged as well as a couple of other small features merged in.

@dmacvicar This looks like good improvement regarding the ssh connection. May you have a look regarding compatibility, etc. if this can be merged into upstream?

libvirt/uri/ssh.go
```diff
@@ -53,17 +74,27 @@ func (u *ConnectionURI) parseAuthMethods() []ssh.AuthMethod {
...
-				log.Printf("[ERROR] Failed to read ssh key: %v", err)
-				continue
-			}
+			for _,keypath := range sshKeyPaths {
+				log.Printf("[DEBUG] Reading ssh key '%s'", keypath)
+				path := os.ExpandEnv(keypath)
+				if strings.HasPrefix(path, "~/") {
+					home, err := os.UserHomeDir()
+					if err == nil {
+						path = strings.Replace(path, "~", home, 1)
```
dmacvicar, line 83: Instead of replacing, you could just join:

```go
if strings.HasPrefix(path, "~/") {
    dirname, _ := os.UserHomeDir()
    path = filepath.Join(dirname, path[2:])	
}
```

libvirt/uri/ssh.go
```diff
@@ -53,17 +74,27 @@ func (u *ConnectionURI) parseAuthMethods() []ssh.AuthMethod {
 			agentClient := agent.NewClient(conn)
 			result = append(result, ssh.PublicKeysCallback(agentClient.Signers))
 		case "privkey":
-			sshKey, err := os.ReadFile(os.ExpandEnv(sshKeyPath))
-			if err != nil {
-				log.Printf("[ERROR] Failed to read ssh key: %v", err)
-				continue
-			}
+			for _,keypath := range sshKeyPaths {
```
dmacvicar, line 77: Was this run over go fmt? does it really allow no spaces after comma?

libvirt/uri/ssh.go
```diff
@@ -91,74 +125,184 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
...
+			log.Printf("[DEBUG] HostName is overriden to: '%s'", hostName);
+		}
+	}
+
+	// we must check for knownhosts and verification for each host we connect to.
+	// the query string values have higher precedence to local configs
 	knownHostsPath := q.Get("knownhosts")
 	knownHostsVerify := q.Get("known_hosts_verify")
-	doVerify := q.Get("no_verify") == ""
+	skip_verify := q.Has("no_verify")
```
dmacvicar, line 179: Golang does not use snake-case for variables.

libvirt/uri/ssh.go
```diff
@@ -91,74 +125,184 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
...
+	var bastion *ssh.Client
+	if err == nil && proxy != "" {
+		log.Printf("[DEBUG] found ProxyJump '%v'", proxy)
+
+		// this is a proxy jump: we recurse into that proxy
+		bastion, err = u.dialHost(proxy, sshcfg, depth + 1)
+		if err != nil {
+			return nil, fmt.Errorf("failed to connect to bastion host '%v': %w", proxy, err)
+		}
 	}
```
dmacvicar, line 258: If the err != nil, it is silently ignored instead of returning. My suggestion: if err != nil, return, then only condition on proxy being non empty to execute the recursion.

libvirt/uri/ssh.go
```diff
@@ -91,74 +125,184 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
...
+		// Dial a connection to the service host, from the bastion
+		conn, err := bastion.Dial("tcp", net.JoinHostPort(hostName, port))
+		if err != nil {
+			log.Fatal(err)
+			return nil, fmt.Errorf("failed to connect to remote host '%v': %w", target, err)
+		}
+
+		ncc, chans, reqs, err := ssh.NewClientConn(conn, target, &cfg)
+		if err != nil {
+			log.Fatal(err)
```
dmacvicar, line 291: This makes no sense. `log.Fatal` will `os.Exit` so the next line will never be executed, and you don't want to use `os.Exit` in a plugin. Just return the error, unless you wanted to achieve something else.

libvirt/uri/ssh.go
```diff
@@ -91,74 +125,184 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to libvirt on the remote host: %w", err)
 	}
+
+	return c, nil
+}
+
+func (u *ConnectionURI) dialHost(target string, sshcfg *ssh_config.Config, depth int) (*ssh.Client, error) {
+
+	if depth > 10 {
```
dmacvicar, line 151: Lets put this magic constant as a constant at the top.

dmacvicar (changes requested):
I am all for merging this. It needs some quality changes and solving the conflicts

libvirt/uri/ssh.go
memetb: I'm not totally clear what you're saying. Could you rephrase please?

libvirt/uri/ssh.go
memetb: roger that. I'll rename to noVerify. The reason I switched from doVerify to skip was because the semantics of the exposed interface were inverted from the internal semantics and it was throwing me off in the debugging (i.e. "do verify" isn't exactly the negation of "no verify" - there are circumstances where there is no verification to be done where "do verify" loses its meaning).

libvirt/uri/ssh.go
memetb: My bad. I was unfamiliar with go's log.Fatal and I assumed it was simply like in python indicating logging facility severity.

I have removed log.Fatal as the intent was simply to return a high severity log trace while returning the error up.

libvirt/uri/ssh.go
memetb: Unclear, can you rephrase please? Did you mean "if the error == nil" ?

Are you perhaps saying there can be a condition in which both `bastion` and `err` are nil?

memetb:
@dmacvicar made the changes you requested, 2 outstanding question. Let me know when you get a chance.

Thanks.

libvirt/uri/ssh.go
dmacvicar: yes. I expect `_, keypath`. Please change. I am surprised go fmt does not change that by itself when you ran it?

libvirt/uri/ssh.go
dmacvicar: What I mean is that `proxy, err = sshcfg.Get(target, "ProxyJump")` can return error, and in that case you would continue with the flow, instead of returning the function. Yes you would skip dialing the proxy (assuming it returned empty string in addition to the error), but the flow would continue anyway.

Wouldn't it make more sense to return early if `proxy, err = sshcfg.Get(target, "ProxyJump")` returns an error. And then you execute the code that dials the proxy only testing for `proxy != ""` ?

dmacvicar (changes requested):

libvirt/uri/ssh.go
memetb: `ProxyJump` can be absent, in which case `Get()` (as opposed to `GetStrict()`)returns empty string [as per sshconfig docs](https://github.com/kevinburke/ssh_config/blob/1d09c0b50564c4a7f8c56c9d5d6d935e06ee94da/config.go#L113-L118). This isn't an error per-se as it simply means there's no proxy. This covers case 1 of the conditional on line 250, which is `err == nil` and `proxy == ""`. In this case, the flow should proceed without proxy.

The second case is if the sshconfig file had a parse error (i.e. `err != nil`) and we don't care about `proxy`. In this case, it means that the sshconfig file couldn't be parsed. Expected behaviour in those circumstances (e.g. when using ssh from cli) is to simply honor the command itself and ignore any ssh_config directives altogether, which in this case corresponds to a flow-through without Proxy.

> Wouldn't it make more sense to return early if proxy, err = sshcfg.Get(target, "ProxyJump") returns an error. And then you execute the code that dials the proxy only testing for proxy != "" ?

I think there's a case to be made that (maybe) the very first successful call to `sshcfg.Get()` guarantees that every subsequent call is successful. However, I'd not feel comfortable breaking the abstraction and depending on this behaviour: it would make the code very convoluted and introduce race conditions (i.e. if `ssh_config` gets touched during this setup).

Thoughts?

libvirt/uri/ssh.go
memetb: >yes. I expect _, keypath. Please change. I am surprised go fmt does not change that by itself when you ran it?

Truthfully, I am not a master of go. I was able to build and also run `go fmt` without any complaints just now.

Change is made.

libvirt/uri/ssh.go
memetb: @dmacvicar see above for my thoughts on this codepath.

dmacvicar:
Thanks for your contribution @memetb , I will try to wrap it up in a release soon.

memetb:
You're welcome. Thanks for the very useful library.

dmacvicar:
Do you see anything worth documenting here?

memetb:
@mhtr please open a new issue, we can look at it over there.

mhtr:
@memetb 
[1116](https://github.com/dmacvicar/terraform-provider-libvirt/issues/1116)

PR #1064, opened 2024-01-06
Mark disk wwn and nvram arguments as computed

wfdewith:
These changes are meant to keep the libvirt provider idempotent, i.e. running terraform apply twice with no changes in between should not result in any updated resources.

The disk WWN is randomly generated if it is not specified, which means that the `disk.wwn` argument will always be different when it is regenerated. This difference causes Terraform to perform an in-place update.

When specifying the `firmware` argument, libvirt will automatically generate the corresponding NVRAM definitions. Since the generated definitions differ from the unspecified `nvram` argument in the resource and the `nvram` argument is marked as force new, Terraform will replace the entire libvirt domain.

The `nvram.file` argument was required, but this is not necessary for libvirt. Libvirt will autogenerate the location of the NVRAM variables file even if only the template is specified.

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -382,7 +382,10 @@ func setFirmware(d *schema.ResourceData, domainDef *libvirtxml.Domain) {
 		}
 
 		if _, ok := d.GetOk("nvram.0"); ok {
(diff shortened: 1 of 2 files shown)


dmacvicar:
Thanks for the contribution.

Issue #356, opened 2018-08-14
Destroying a domain should try to power it off first


multani:
#  Version Reports:

### Distro  version of host:

Debian stretch

### Terraform Version Report
v0.11.7

### Libvirt version

`3.0.0`

### terraform-provider-libvirt plugin version (git-hash)

`58cc06611667bbbdddafacb411cd6bc8da1c2b00`
___
# Description of Issue/Question

When a domain is destroyed by Terraform, it should be possible to send it a "power-off" signal so that it has the chance to shut down properly.

Althoug the current method is very fast at replacing VMs, it is also very brutal and doesn't offer any chance for the VM to react correctly before disappearing (such as deregistering services upon shutdown, etc.).

This is [the last part of that wiki paragraph](https://wiki.libvirt.org/page/VM_lifecycle#Stopping_or_rebooting_a_domain):

> An inelegant shutdown, also known as hard-stop:
> 
> `# virsh destroy <domain>`
>
> This is equivalent to unplugging the power cable.

dmacvicar:
I think there are flags on the libvirt API that controls the destruction of the domain to do something like this. However this would need some research.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

dmacvicar:
My biggest concern here is that we would need to introduce time-outs which would slow down everything, or am I wrong?

MalloZup:
@multani  just for curiosity, if you are going to destroy/remove forever a VM why ``` deregistering services upon shutdown``` is important ? I mean if the VM is removed you should stop the service before removing it, or if you are going to remove a vm the service inside the VM are not anymore important.
I cannot figure out a use case for it. But maybe you know better on the subject.
To me at this point i agree with @dmacvicar  that performance can be really impacted by this, especially if we are waiting for services to gracefull shutdown , this can take a while, so i would stay with the current destroy destroy 

TIA :sunflower:

multani:
So, just to take this out: I like the current behavior which allows for very fast destroy/creation cycle and I think it should be kept somehow!

@MalloZup As for the use case: we have some services (Consul and Sensu, in particular) which are running on these VMs (they provide a way to register the VMs automatically in the service discovery when they come in). Destroying the VMs with the current behavior offers no chances at all for these services to signal they are shutting down (properly) and for the rest of the world, these VMs simply stopped working and we have to manually deregister them.
[...]

Issue #1060, opened 2023-12-27
Graceful shutdown of domain resource when recreate is required


karelvanhecke:
### Linux distribution

``` CentOS Stream 8 ```

### Terraform version

```sh
$ terraform -v
Terraform v1.6.6
on linux_amd64
+ provider registry.terraform.io/dmacvicar/libvirt v0.7.6
```

### Provider and libvirt versions

```sh
dmacvicar/libvirt v0.7.6
libvirtd (libvirt) 8.0.0
```

___

## Description of Issue/Question

Currently, when a domain resource has been modified and requires a recreate, the domain is destroyed and recreated.
While this is the most performant solution, it isn't necessarily the safest when you consider that the attached volumes aren't necessarily destroyed and recreated.

This could in theory lead to data corruption.

In #356 graceful shutdown was proposed, but this issue was closed.

marshallford:
I'm interested in this feature as well, particularly for the use case of running immutable infrastructure with Fedora CoreOS or Flatcar Container Linux where the established pattern is to replace VMs when making configuration/version changes (with ignition) rather than using a config mgmt tool over the top.

Silvenga:
This seems important, as changing the memory of a domain will cause the domain to be replaced. This means the next boot of the same volume will be after an unclean shutdown.

scabala:
Related #1069

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

This issue report is relevant to the legacy version of the provider, which is now in the v0.8 branch of this repository.

Future development is based on a new provider, which is not compatible with this one, nor does share code.

As the new provider solves many issues with the legacy, and to make development in my free time more enjoyable, I have decided to close all bugs for the legacy provider, and to ask to check if the bug would apply to the new one. This also to encourage trying the new version.

- https://registry.terraform.io/providers/dmacvicar/libvirt/latest

and check the documentation:

- https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/README.md
- https://registry.terraform.io/providers/dmacvicar/libvirt/latest/docs

You are free to reopen the bug for v0.8. It may encourage someone to fix it. My efforts will go into the new provider.

I ask you to check the new provider and re-evaluate this bug. 🙏

nicholas-rees:
I have been trying to figure out this issue in an effort to understand why the `destroy.graceful` values specified in the schema did not seem to be effective.

[...]

PR #1066, opened 2024-01-23
adds wait_for_shutoff to resource_libvirt_domain

marshallford:
Open to feedback, there is likely a better way to implement this. I took a first swing at the docs, but I wasn't sure where to start on testing this change.

Tested with Terraform version 1.7.0.

```

Fixes: #356, #1060

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -871,3 +871,16 @@ func destroyDomainByUserRequest(virConn *libvirt.Libvirt, d *schema.ResourceData
 
 	return nil
 }
+
+func waitForDomainShutoff(virConn *libvirt.Libvirt, domain libvirt.Domain) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		state, _, err := virConn.DomainGetState(domain, 0)
+		if err != nil {
+			return nil, "", err
+		}
+		if libvirt.DomainState(state) == libvirt.DomainShutoff {
+			return domain, "SHUTOFF", nil
+		}
+		return domain, "NOT-SHUTOFF", err
+	}
+}
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -15,6 +15,7 @@ import (
 	libvirt "github.com/digitalocean/go-libvirt"
 	"github.com/dmacvicar/terraform-provider-libvirt/libvirt/helper/suppress"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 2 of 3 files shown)


dmacvicar:
While I acknowledge the problem, I am not sure we are fixing it in the right place.

- How do other cloud providers behave here? do we have some similar setting to model against?
- Given that we already track "running" as state, wouldn't make more sense to implement graceful shutdown there?
- If we implement it in `Destroy`, wouldn't make more sense to always do it graceful and just have a timeout for destroy? (those [are customizable](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts), so one could have a 0 timeout)

Silvenga:
> How do other cloud providers behave here?

I don't think cloud providers ultimately need to destroy the resource to update it e.g. scaling up typically requires just rebooting the node. This is mostly a TF'isum, where it's just easier to think of the resources as immutable.

To my knowledge, the cloud providers that I'm aware of don't do graceful shutdowns on destroying (Digital Ocean and Azure).

> If we implement it in Destroy, wouldn't make more sense to always do it graceful and just have a timeout for destroy? (those [are customizable](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts), so one could have a 0 timeout)

IMO, this ultimately should be opt-in - as there are a lot of cases that would prevent libvirt from communicating with the guest for a graceful shutdown. Some OS's don't even respond to graceful shutdowns, one being at least some versions of VyOS (a router OS, "common" in public clouds). I also don't believe the agent is installed by default in at least the slim version of Ubuntu Server.

scabala:
I really like [this](https://github.com/dmacvicar/terraform-provider-libvirt/issues/356#issuecomment-508782709) comment on #356 - there should be just a timeout variable that controls the logic. If it is greater than 0, do graceful shutdown and force shutdown when it expires. If it is 0, do forceful shutdown as currently it is done.

dmacvicar:
Yes, it is a good solution. I hope people don't confuse it with the destroy timeout that terraform provides as a setting (which is the timeout of the destroy function itself).

tiaden:
Here is a link to an implementation of this issue that I worked on some time ago, which may provide an alternative perspective:
[GitHub link](https://github.com/tiaden/terraform-provider-libvirt/blob/005fb3ffd5bb22503fdc41b987831027fd384e0d/libvirt/domain.go#L903).

I have successfully used this solution in my projects without encountering any problems. It leverages the "running" attribute to shut down the virtual machine, utilizing the graceful default shutdown method or QEMU (if specified), and, as a last resort, forcibly powers off the machine if the default Terraform timeout is reached. If you'd like to test it quickly, you can use the associated Terraform provider available here: [Terraform Provider](https://registry.terraform.io/providers/tiaden/libvirt/latest).

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #1088, opened 2024-07-09
feat: lvm storage pool impl

jimnydev:
Revamped #814.


dmacvicar:
I want to merge this one, but I'd like to move 'source_devices` to a `source` and `device` blocks, mimicking libvirt  `source` and `target` and deprecating the top level `path` attribute so that we make room for additional pool alternatives.

I will try to get it done on main, and then I will rebase this one, if you don't mind.

libvirt/resource_libvirt_pool.go
```diff
@@ -207,11 +254,14 @@ func resourceLibvirtPoolRead(ctx context.Context, d *schema.ResourceData, meta i
 		poolPath = poolDef.Target.Path
 	}
 
-	if poolPath == "" {
-		log.Printf("Pool %s has no path specified", pool.Name)
-	} else {
-		log.Printf("[DEBUG] Pool %s path: %s", pool.Name, poolPath)
-		d.Set("path", poolPath)
+	// for logical pool the path auto-generated, so we don't set/read it
```
dmacvicar, line 257: But it is missing reading the source devices. I am fixing that in my local branch which I will push here.

libvirt/resource_libvirt_pool.go
```diff
@@ -248,9 +347,17 @@ func resourceLibvirtPoolDelete(ctx context.Context, d *schema.ResourceData, meta
...
 
-	err = virConn.StoragePoolDelete(pool, 0)
-	if err != nil {
-		return diag.Errorf("error deleting storage pool: %s", err)
+	poolDef, dia := newDefPoolFromLibvirt(virConn, pool)
+	if dia != nil {
+		return dia
+	}
+
+	// if the logical pool has no source device then the volume group existed before we created the pool, so we don't delete it
```
dmacvicar, line 355: @jimnydev can you explain more the logic behind this? why would in this case the pool not have a source?

I don't have used these myself, so I may be missing some context.

PR #1092, opened 2024-08-20
Fix domain creation with libvirt (version 9.2.0 and greater) on s390x

telpelt:
On the s390x architecture libvirt (version 9.2.0 and greater) do not allow the features defined in function `newDomainDef()`. This patch removed the `Features: &libvirtxml.DomainFeatureList{...}` object for this specific architecture. This code was successfully tested with RHEL 9.x on IBM z14.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,17 +22,17 @@ jobs:
       matrix:
         go-version: ["1.21.x"]
     steps:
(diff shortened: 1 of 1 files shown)


dmacvicar:
Messed your branch while trying to rebase, so I pushed this directly to main. Thanks!

PR #1096, opened 2024-09-02
Fix multi line string in integration tests

cbosdo:
Fix the integration test failure introduced in PR #1068


dmacvicar:
Thanks. I will try to get a release out in the next days.

Issue #947, opened 2022-04-21
Provider does not track active state of network


agorgl:
When creating a libvirt network without autostart and rebooting the hypervisor, the network and the instances (domains) are inactive / stopped, as expected.

Running `terraform plan` in this scenario we get this:

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # libvirt_domain.instance will be updated in-place
  ~ resource "libvirt_domain" "instance" {
        id          = "7dbfcef6-0dd3-4ee2-854e-0f31d3c30083"
        name        = "instance"
      ~ running     = false -> true
        # (10 unchanged attributes hidden)


        # (3 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Changes to Outputs:
  - instance_ip = "10.3.0.61" -> null

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

libvirt_domain.instance: Modifying... [id=7dbfcef6-0dd3-4ee2-854e-0f31d3c30083]
╷
│ Error: error creating libvirt domain: Requested operation is not valid: network 'network' is not active
│
│   with libvirt_domain.instance,
│   on main.tf line 80, in resource "libvirt_domain" "instance":
│   80: resource "libvirt_domain" "instance" {
│
╵
```

This shows that provider is able to track if the libvirt_domain is running and try to bring it up again, but it does NOT! track the state of the network it is connected to and we manually need to run `virsh net-start network` before running `terraform plan` again in order to get a successful result.

dmacvicar:
```golang
// resourceLibvirtNetworkUpdate updates dynamically some attributes in the network
func resourceLibvirtNetworkUpdate(d *schema.ResourceData, meta interface{}) error {
	// check the list of things that can be changed dynamically
	// in https://wiki.libvirt.org/page/Networking#virsh_net-update
	virConn := meta.(*Client).libvirt
	if virConn == nil {
		return fmt.Errorf(LibVirtConIsNil)
	}

	network, err := virConn.NetworkLookupByUUID(parseUUID(d.Id()))
	if err != nil {
		return fmt.Errorf("can't retrieve network with ID '%s' during update: %s", d.Id(), err)
	}

	d.Partial(true)

	activeInt, err := virConn.NetworkIsActive(network)
[...]

Issue #1095, opened 2024-08-31
Network is not started if inactive


dansrogers:
### Linux distribution

```Ubuntu 24.04 LTS```

### Terraform version

```sh
$ terraform -v
Terraform v1.9.5-dev
on linux_amd64
+ provider registry.terraform.io/dmacvicar/libvirt v0.7.6

Your version of Terraform is out of date! The latest version
is 1.9.5. You can update by downloading from https://www.terraform.io/downloads.html
```

### Provider and libvirt versions

```sh
registry.terraform.io/dmacvicar/libvirt v0.7.6

libvirt-daemon                                 10.0.0-2ubuntu8.3                        amd64        Virtualization daemon

```
___

- [X ] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

## Description of Issue/Question

Network is not started if it is down.  If network is not set to autostart on boot, then after a restart it will be inactive.  Terraform apply will not activate the network, and thus VMs that depend on that network will fail to start.

### Setup
```
terraform {
  required_providers {
    libvirt = {
      source = "dmacvicar/libvirt"
      version = "0.7.6"
    }
  }
}

# Configure the Libvirt provider
provider "libvirt" {
  uri = "qemu:///system"
}

resource "libvirt_network" "cluster" {
  name = "cluster"
  mode = "route"
  addresses = ["10.0.1.0/24"]
  dhcp {
    enabled = true
  }
}
```
### Steps to Reproduce Issue
```sh
$ terraform apply
$ virsh net-destroy cluster
$ terraform apply
# note that terraform doesn't bring up the network.  It must be manually started with virsh net-start cluster
```

If you include a domain definition, you'll get an error similar to the following if the cluster is not active:
```
╷
│ Error: error creating libvirt domain: Requested operation is not valid: network 'cluster' is not active
│ 
│   with module.worker.libvirt_domain.devlab_cluster,
│   on modules/libvirt_talos_host/main.tf line 27, in resource "libvirt_domain" "devlab_cluster":
│   27: resource "libvirt_domain" "devlab_cluster" {
│ 
╵
```
___
## Additional information:

Do you have SELinux or Apparmor/Firewall enabled? Some special configuration?

no

scabala:
I have encountered it as well. I believe this is just a limitation of current implementation and probably needs to be addressed.

dmacvicar:
[...]

PR #1097, opened 2024-09-11
fix: Network is not started if inactive #1095

ne-bknn:
Includes network activity status in terraform state, thus allowing terraform to update network resource in case it is inactive (e.g. after hypervisor reboot or `virsh net-destroy`)

Network activity is checked in Update method, but not in Read. This makes terraform think that network should not be updated (since the state matches recorded one). Then, VM initialization fails, since it depends on an inactive network. 

First brought up in #1095.

### DIFF
--- a/libvirt/resource_libvirt_network.go
+++ b/libvirt/resource_libvirt_network.go
@@ -91,6 +91,13 @@ func resourceLibvirtNetwork() *schema.Resource {
 				Required: false,
 				Computed: true,
 			},
+			"active": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+				ForceNew: false,
+				Required: false,
+			},
 			"dns": {
 				Type:     schema.TypeList,
 				Optional: true,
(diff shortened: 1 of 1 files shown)


ne-bknn:
Uh, actually referenced here with explanation why this isn't yet fixed: #947

I've just copy-pasted `running` variable from `libvirt_domain` schema. My reasoning is that they are semantically equivalent, so if this variable config is correct for `libvirt_domain` it should be correct for `libvirt_network`. What's the catch?

Full disclosure: only user-level experience with terraform, no experience with libvirt. My assumptions may be horribly wrong.

dmacvicar:
Thanks for the attempt.

It is more complicated than that. It requires on acting on the status based of what you read to bring the network up if it was up last time and it is now down.

ne-bknn:
I understand that I can be and most probably is wrong, but I do not understand why. Network state is checked in `Update` and `Update` brings it up if its not up: https://github.com/dmacvicar/terraform-provider-libvirt/blob/f5a8334bbecd7ca778739f34d50b71e9c479e795/libvirt/resource_libvirt_network.go#L313-L323

We just have to signal to terraform that network state is not the expected one. My (manual and naive) tests show that it works - if the network is up, it won't be updated, if it is down, it will be. What case I fail to see?

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

Issue #1105, opened 2024-09-30
v0.8.0 plugin crahes


mvrk69:
### Linux distribution

``` Fedora 40 ```

### Terraform version

```
Terraform v1.9.6
on linux_amd64
+ provider registry.terraform.io/dmacvicar/libvirt v0.8.0
```

### Setup

00-provider.tf
terraform {
  required_providers {
    libvirt = {
      source = "dmacvicar/libvirt"
    }
  }
}

provider "libvirt" {
  uri   = "qemu+ssh://root@${var.compute_host}/system?keyfile=${var.keyfile}&sshauth=privkey"
}

-------------

01-variables.tf
variable "compute_host" {
  description = "Compute Hostname"
  type        = string
}

variable "keyfile" {
  description = "Compute host ssh key"
  type        = string
}

variable "datastore" {
  description = "Datastore"
  type        = string
}

---------------

02-variables-values.tfvars
compute_host = "xxx.xxx.xxx.xxx"
keyfile = "/home/user/.ssh/id_rsa"
datastore = "vms"

----------------

### Steps to Reproduce Issue

terraform plan -var-file="02-variables-values.tfvars"

Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Plugin did not respond
│
│   with provider["registry.terraform.io/dmacvicar/libvirt"],
│   on 00-provider.tf line 13, in provider "libvirt":
│   13: provider "libvirt" {
│
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ConfigureProvider call. The plugin logs may contain more details.
╵

Stack trace from the terraform-provider-libvirt_v0.8.0 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x79911c]

goroutine 12 [running]:
github.com/kevinburke/ssh_config.(*Config).Get(0x0, {0xc00004e910, 0xb}, {0xe88d73, 0x8})
	github.com/kevinburke/ssh_config@v1.2.0/config.go:343 +0x5c
github.com/dmacvicar/terraform-provider-libvirt/libvirt/uri.(*ConnectionURI).dialHost(0xc000140170, {0xc00004e910, 0xb}, 0x0, 0x0)
	github.com/dmacvicar/terraform-provider-libvirt/libvirt/uri/ssh.go:166 +0x259
[...]

PR #1109, opened 2024-10-13
Sshconfig missing bugfix - addresses issue #1105

memetb:
This PR addresses issue #1105. It is a minimal codefix, marking as WIP as I've made the decision to make it a minimal fix as opposed to a larger rewrite. 

@dmacvicar I am open to any directives you have to give. I will separately comment on my design choice.

### DIFF
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -116,21 +116,17 @@ func (u *ConnectionURI) parseAuthMethods(target string, sshcfg *ssh_config.Confi
 // construct the whole ssh connection, which can consist of multiple hops if using proxy jumps,
 // the ssh configuration file is loaded once and passed along to each host connection.
 func (u *ConnectionURI) dialSSH() (net.Conn, error) {
+	var sshcfg* ssh_config.Config = nil
+
 	sshConfigFile, err := os.Open(os.ExpandEnv(defaultSSHConfigFile))
 	if err != nil {
 		log.Printf("[WARN] Failed to open ssh config file: %v", err)
-		sshConfigFile = strings.NewReader("") // use empty file instead
-	}
-
-	sshcfg, err := ssh_config.Decode(sshConfigFile)
-	if err != nil {
-		log.Printf("[WARN] Failed to parse ssh config file: '%v' - sshconfig will be ignored.", err)
-
-		sshcfg, err = ssh_config.Decode(strings.NewReader(""))
+	} else {
+		sshcfg, err = ssh_config.Decode(sshConfigFile)
 		if err != nil {
-			log.Printf("[ERROR] unrecoverable error occurred during ssh configuration")
-			return nil, err
+			log.Printf("[WARN] Failed to parse ssh config file: '%v' - sshconfig will be ignored.", err)
 		}
+
 	}
 
 	// configuration loaded, build tunnel
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -238,7 +238,7 @@ func (u *ConnectionURI) dialHost(target string, sshcfg *ssh_config.Config, depth
 		if sshcfg != nil {
 			keyAlgs, err := sshcfg.Get(target, "HostKeyAlgorithms")
 			if err == nil && keyAlgs != "" {
-				log.Printf("[DEBUG] HostKeyAlgorithms is overriden to '%s'", keyAlgs)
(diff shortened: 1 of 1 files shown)


memetb:
The log output provided by @mvrk69 has the following relevant output:

```
2024-10-03T14:45:39.141+0200 [INFO]  provider.terraform-provider-libvirt_v0.8.0: 2024/10/03 14:45:39 [WARN] Failed to open ssh config file: open /home/user/.ssh/config: no such file or directory: timestamp="2024-10-03T14:45:39.141+0200"
2024-10-03T14:45:39.141+0200 [INFO]  provider.terraform-provider-libvirt_v0.8.0: 2024/10/03 14:45:39 [WARN] Failed to parse ssh config file: invalid argument: timestamp="2024-10-03T14:45:39.141+0200"
2024-10-03T14:45:39.141+0200 [INFO]  provider.terraform-provider-libvirt_v0.8.0: 2024/10/03 14:45:39 [INFO] establishing ssh connection to 'XXX.XXX.XXX.XXX': timestamp="2024-10-03T14:45:39.141+0200"
2024-10-03T14:45:39.143+0200 [DEBUG] provider.terraform-provider-libvirt_v0.8.0: panic: runtime error: invalid memory address or nil pointer dereference
2024-10-03T14:45:39.143+0200 [DEBUG] provider.terraform-provider-libvirt_v0.8.0: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x79911c]
```

This means that the loading of the sshconfig file initially fails [here](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/uri/ssh.go#L119), 
```
sshConfigFile, err := os.Open(os.ExpandEnv(defaultSSHConfigFile))
```

then code execution proceeds to [here](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/uri/ssh.go#L124)
```
sshcfg, err := ssh_config.Decode(sshConfigFile)
```

which returns a nil for the `sshcfg` object. 

This object is then queried multiple times inside the `dialHost` call [here](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/uri/ssh.go#L130). 

It's important to note that `dialHost` is quite dynamic in that the same call (e.g. `sshcfg.Get(target, "HostName")`) might yield different results at multiple tunnel hop levels. Furthermore, there are several spots where `sshcfg` is queried.

Because of the above complexity, and because I didn't want to change all the sites where the `sshcfg` object is interacted with, I chose to load an empty configuration file and simply proceeding with a guarantee that `dialHost` will never receive a `nil` `sshcfg` arg.

memetb:
Here is the updated code. Allows for graceful absence of `.ssh/config` file, unreadable ssh config file, and empty ssh config file.

dmacvicar:
I am ok with the current fix. But in retrospective I think the whole introduction of ssh config file support in this way was a mistake.

What we should do is abstract the whole code flow to work with a final configuration. How this configuration is built should be in a single place. The ssConfig object being all over the place is not nice.

The code with the loading of the file, overriding etc, should all be in a single place, and that is where the code that loads a empty file could be as well.

memetb:
@dmacvicar I understand the point you're making: there's an aspect of the current implementation that is messy.

The thing to keep in mind if we do refactor, is that the ssh configuration can't be pre-computed without some serious state management.

For instance, I've worked in environments where I had:

```
Match Host corp.asset-*
       User service
       Hostname %h
       ProxyJump mothership.corp.com
```

determining the outcome of these and pre-computing them into a dictionary is simply not possible.

I think a second area of concern is the "level" the query string applies to: does it apply to each host along the way or just the end host etc?

My vote would be to formalize the query string configuration into an object and formalize the library (terraform-provider-libvirt) api documentation to show the precedence of configuration, and then if it must be done, create a wrapper object which polls sshconfig and query string according to those rules.

The important thing is that the config must specify the `target` attribute every time it calls `Get`.

zaneb:
@memetb on the assumption  you missed it, FYI I added ssh Dialer support to go-libvirt directly, along with libvirt URI parsing, in https://github.com/digitalocean/go-libvirt/pull/209

There's some backward compatibility issues for terraform-provider-libvirt, which IIRC accepts query options for the [`libssh` transport](https://libvirt.org/uri.html#libssh-and-libssh2-transport) in `ssh` URIs. But those could easily be handled by some light pre-processing.

Anyhow, if you're planning to refactor anyway, you could perhaps consider contributing the sshconfig support there.

Issue #1108, opened 2024-10-12
no check for id_ed25519 ssh key (only id_rsa is checked)


korncola:
### Linux distribution

``` Arch Linux ```

### Terraform version

1.9.7

### Provider and libvirt versions

provider: 0.8.0

## Description of Issue/Question

module only checks for `id_rsa` ssh key, not for other types like the recommend state of the art key type`id_ed25519`

### Steps to Reproduce Issue

* have no `id_rsa` just a `id_ed25519`
* ssh connect to qemu fails with:
* `[INFO]  provider.terraform-provider-libvirt_v0.8.0: 2024/10/12 19:53:03 [ERROR] Failed to read ssh key '${HOME}/.ssh/id_rsa': open /Users/korncola/.ssh/id_rsa: no such file or directory: timestamp="2024-10-12T19:53:03.332+0200"`
* `[ERROR] provider.terraform-provider-libvirt_v0.8.0: Response contains error diagnostic: tf_req_id=6d164000-3306-e756-43da-748d1f0823de tf_rpc=Configure @caller=github.com/hashicorp/terraform-plugin-go@v0.24.0/tfprotov5/internal/diag/diagnostics.go:58 diagnostic_detail="" diagnostic_severity=ERROR tf_proto_version=5.6 @module=sdk.proto diagnostic_summary="failed to connect: failed to connect to remote host 'kvm.example.com': ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain" tf_provider_addr=provider timestamp="2024-10-12T19:53:03.373+0200"`

**Comment by memetb** [comment]
@korncola confirm that the latest release v0.8.1 in PR #1112 solves your problem.

**Comment by mhtr** [comment]
@memetb hi
I have the same issue on Debian 12 with v.0.8.1 provider version.

```
2024-10-24T09:08:15.008Z [INFO]  provider.terraform-provider-libvirt_v0.8.1: 2024/10/24 09:08:15 [DEBUG] Reading ssh key '/home/username/.ssh/id_rsa': timestamp=2024-10-24T09:08:15.008Z
2024-10-24T09:08:15.008Z [INFO]  provider.terraform-provider-libvirt_v0.8.1: 2024/10/24 09:08:15 [INFO] SSH connecting to '<libvirt_host_ip>' (<libvirt_host_ip>): timestamp=2024-10-24T09:08:15.008Z
[...]

PR #1117, opened 2024-10-23
Bugfix: ssh port override for #1116

### LINKED ISSUE #1116: Ignoring Port directive from SSH config
### OS

``` MacOS 15.0.1```

### Terraform version

```sh
terraform -v
Terraform v1.9.2
on darwin_arm64
```

```sh
terraform-provider-libvirt -version
0.8.1
```

Unfortunately it ignores the settings in ~/.ssh/config that concern the Port directive. For the bastion node and for the host a custom port is specified, but the connection still goes to port 22.

```
Host <bastion_ip>
  Port 12345
  User username

Host <libvirt_host>
 Port 12345
 User username
 ProxyJump <bastion_ip>
```

Error:
`Error: failed to connect: failed to connect to bastion host '<bastion_ip>': failed to connect to remote host '<bastion_ip>': dial tcp <bastion_ip>:22: connect: connection refused`

If I don't use bastion and specify port in `uri` like this:
```
provider "libvirt" {
    uri = "qemu+ssh://username@<libvirt_host>:12345/system"
}
```
I get an error:
`Error: failed to connect: failed to connect to remote host '<libvirt_host>:12345': dial tcp: lookup <libvirt_host>:12345: no such host`

memetb:
This issue fixes #1116.

It is a direct fix that maintains backward compatibility, however it introduces an outcome that breaks the principle of least astonishment: the port specified in the query string will take precedence over the port specified in *any* configured targets along the way (e.g. client -> bastion1 -> bastion2 -> host).

My opinion is that config should have priority over query string, however, this behaviour is not backwards compatible.

@dmacvicar : will need your input on this.

### DIFF
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -169,12 +169,14 @@ func (u *ConnectionURI) dialHost(target string, sshcfg *ssh_config.Config, depth
 			configuredPort, err := sshcfg.Get(target, "Port")
 			if err == nil && configuredPort != "" {
 				port = configuredPort
+				log.Printf("[DEBUG] using ssh port from ssh_config: '%s'", port)
 			}
 		}
 
 	} else {
-		log.Printf("[DEBUG] ssh Port is overridden to: '%s'", port)
+		log.Printf("[DEBUG] using ssh port from querystring: '%s'", port)
 	}
+	log.Printf("[DEBUG] port for ssh connection is: '%s'", port)
 
 	hostName := target
 	if sshcfg != nil {
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -168,9 +168,21 @@ func (u *ConnectionURI) dialHost(target string, sshcfg *ssh_config.Config, depth
 	//  3. defaultSSHPort
 	port := ""
 
-	if sshcfg != nil && (configuredPort, err := sshcfg.Get(target, "Port")); err == nil && configuredPort != "" {
+	if sshcfg != nil {
+		configuredPort, err := sshcfg.Get(target, "Port")
+		if err != nil {
+			log.Printf("[WARN] error reading Port attribute from ssh_config for target '%v'", target)
+		} else {
+			port = configuredPort
 
-		port = configuredPort
+			if port == "" {
+				log.Printf("[DEBUG] port for target '%v' in ssh_config is empty", target)
+			}
+		}
+	}
+
+	if port != "" {
+
 		log.Printf("[DEBUG] using ssh port from ssh_config: '%s'", port)
 
 	} else if u.Port() != "" {
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -162,21 +162,26 @@ func (u *ConnectionURI) dialHost(target string, sshcfg *ssh_config.Config, depth
 
 	q := u.Query()
 
-	port := u.Port()
-	if port == "" {
-		port = defaultSSHPort
-		if sshcfg != nil {
-			configuredPort, err := sshcfg.Get(target, "Port")
-			if err == nil && configuredPort != "" {
-				port = configuredPort
-				log.Printf("[DEBUG] using ssh port from ssh_config: '%s'", port)
-			}
-		}
+	// port override order of precedence (starting with highest):
+	//  1. specific stanza entry in ssh_config for this target (this includes default global entries in ssh config)
+	//  2. port specified in connection string
+	//  3. defaultSSHPort
+	port := ""
 
+	if sshcfg != nil && (configuredPort, err := sshcfg.Get(target, "Port")); err == nil && configuredPort != "" {
+
+		port = configuredPort
+		log.Printf("[DEBUG] using ssh port from ssh_config: '%s'", port)
+
+	} else if u.Port() != "" {
+
+		port = u.Port()
+		log.Printf("[DEBUG] using connection string port ('%s')", port)
 	} else {
-		log.Printf("[DEBUG] using ssh port from querystring: '%s'", port)
+
+		port := defaultSSHPort
+		log.Printf("[DEBUG] using default port for ssh connection ('%s')", port)
 	}
-	log.Printf("[DEBUG] port for ssh connection is: '%s'", port)
 
 	hostName := target
 	if sshcfg != nil {
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -132,7 +132,7 @@ func (u *ConnectionURI) dialSSH() (net.Conn, error) {
 	}
 
 	// configuration loaded, build tunnel
-	sshClient, err := u.dialHost(u.Host, sshcfg, 0)
+	sshClient, err := u.dialHost(u.Hostname(), sshcfg, 0)
 	if err != nil {
 		return nil, err
 	}
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -165,6 +165,13 @@ func (u *ConnectionURI) dialHost(target string, sshcfg *ssh_config.Config, depth
 	port := u.Port()
 	if port == "" {
 		port = defaultSSHPort
+		if sshcfg != nil {
+			configuredPort, err := sshcfg.Get(target, "Port")
+			if err == nil && configuredPort != "" {
+				port = configuredPort
+			}
+		}
+
 	} else {
 		log.Printf("[DEBUG] ssh Port is overridden to: '%s'", port)
 	}


scabala:
I think that it is not breaking principle of least astonishment because, IIRC specifying port on commandline takes precedence than port specification in config file. I do not know, however, how does it work with libvirt's uri string.

memetb:
@scabala can you reproduce this?

From [man page](https://www.man7.org/linux/man-pages/man5/ssh_config.5.html):

```
ssh(1) obtains configuration data from the following sources in
     the following order:

           1.   command-line options
           2.   user's configuration file (~/.ssh/config)
           3.   system-wide configuration file (/etc/ssh/ssh_config)

     Unless noted otherwise, for each parameter, the first obtained
     value will be used.  The configuration files contain sections
     separated by Host specifications, and that section is only
     applied for hosts that match one of the patterns given in the
     specification.  The matched host name is usually the one given on
     the command line (see the CanonicalizeHostname option for
     exceptions).

     Since the first obtained value for each parameter is used, more
     host-specific declarations should be given near the beginning of
     the file, and general defaults at the end.

     The file contains keyword-argument pairs, one per line.  Lines
     starting with ‘#’ and empty lines are interpreted as comments.
     Arguments may optionally be enclosed in double quotes (") in
     order to represent arguments containing spaces.  Configuration
     options may be separated by whitespace or optional whitespace and
     exactly one ‘=’; the latter format is useful to avoid the need to
     quote whitespace when specifying configuration options using the
     ssh, scp, and sftp -o option.

```

But contrast with:

```
ProxyJump
     Specifies one or more jump proxies as either
     [user@]host[:port] or an ssh URI.  Multiple proxies may
     be separated by comma characters and will be visited
     sequentially.  Setting this option will cause ssh(1) to
     connect to the target host by first making a ssh(1)
     connection to the specified ProxyJump host and then
     establishing a TCP forwarding to the ultimate target from
     there.  Setting the host to none disables this option
     entirely.

     Note that this option will compete with the ProxyCommand
     option - whichever is specified first will prevent later
     instances of the other from taking effect.

     Note also that the configuration for the destination host
     (either supplied via the command-line or the
     configuration file) is not generally applied to jump        <------------- !!!
     hosts.  ~/.ssh/config should be used if specific
     configuration is required for jump hosts.
```

I also did a quick test on my local machine and it behaves as I expect:

```
memet@machine:~$ head ~/.ssh/config | tail -n+2
UseRoaming no

Host fake-jump
        Port 2222
        HostName 192.168.50.165

Host bogey
        Port 2222
        HostName localhost
        ProxyJump fake-jump
memet@machine:~$ ssh fake-jump
ssh: connect to host 192.168.50.165 port 2222: Connection refused
memet@machine:~$ ssh fake-jump -p 22 
# success
memet@machine:~$
logout
Connection to 192.168.50.165 closed.
memet@machine:~$ ssh bogey
ssh: connect to host 192.168.50.165 port 2222: Connection refused
kex_exchange_identification: Connection closed by remote host
Connection closed by UNKNOWN port 65535
memet@machine:~$ ssh bogey -p 22
ssh: connect to host 192.168.50.165 port 2222: Connection refused
kex_exchange_identification: Connection closed by remote host
Connection closed by UNKNOWN port 65535
```

scabala:
I was writing from memory but what from example you provided it looks like it works as I wrote - commandline has priority over config.
When connecting to `fake-jump`, port from config is 2222 and it does not work, only when overridden by port 22 from commandline.

Second example does not work because IMO it first tries to port 2222 on `fake-jump` which IMO is fine - I do not know of any way to override this from commandline.

Now when I read your description in PR it is more clear to me what you meant. I must say I was lost a bit - this SSH handling part grew significantly. Yeah, I guess that might indeed be a little surprising to user but I think this is can be dealt with some little documentation.

memetb:
Please see **TL;DR** at end. The answer is trivial.

> it looks like it works as I wrote - commandline has priority over config. 

I'm not sure about that, I've re-done the same output for more explicit clarity. You can try the below stanzas locally too...

``` bash
memet@machine:~$ head -n 16 ~/.ssh/config | tail -n+2
UseRoaming no

Host fake-jump
        Port 2222
        HostName 192.168.50.165 # this is my local DHCP address

Host bogey-1
        Port 2222
        HostName localhost
        ProxyJump fake-jump

Host bogey-2
        Port 2222
        HostName localhost

memet@machine:~$ ssh bogey-2    # FAILS WITHOUT OVERRIDE
ssh: connect to host localhost port 2222: Connection refused
memet@machine:~$ ssh bogey-2 -p 22  # SUCCEEDS WITH FLAG
Linux machine 6.1.0-22-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.94-1 (2024-06-21) x86_64
# SUCCESS ...
memet@machine:~$ ssh fake-jump # THIS FAILS WITHOUT OVERRIDE
ssh: connect to host 192.168.50.165 port 2222: Connection refused
memet@machine:~$ ssh fake-jump -p 22 # THIS SUCCEEDS WITH FLAG
Linux machine 6.1.0-22-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.94-1 (2024-06-21) x86_64
# SUCCESS ...
memet@machine:~$
logout
Connection to 192.168.50.165 closed.
memet@machine:~$ ssh bogey-1 # THIS FAILS WITHOUT OVERRIDE
ssh: connect to host 192.168.50.165 port 2222: Connection refused
kex_exchange_identification: Connection closed by remote host
Connection closed by UNKNOWN port 65535
memet@machine:~$ ssh bogey-1 -p 22 # THIS STILL FAILS WITH FLAG 
ssh: connect to host 192.168.50.165 port 2222: Connection refused 
kex_exchange_identification: Connection closed by remote host
Connection closed by UNKNOWN port 65535
memet@machine:~$ ssh bogey-1 -p 22 -v
OpenSSH_9.2p1 Debian-2+deb12u2, OpenSSL 3.0.13 30 Jan 2024
debug1: Reading configuration data /home/memet/.ssh/config
debug1: /home/memet/.ssh/config line 2: Deprecated option "useroaming"
debug1: /home/memet/.ssh/config line 8: Applying options for bogey-1
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 19: Applying options for *
debug1: Setting implicit ProxyCommand from ProxyJump: ssh -v -W '[%h]:%p' fake-jump
debug1: Executing proxy command: exec ssh -v -W '[localhost]:22' fake-jump
debug1: identity file /home/memet/.ssh/****** type 0
debug1: identity file /home/memet/.ssh/******-cert type -1
debug1: Local version string SSH-2.0-OpenSSH_9.2p1 Debian-2+deb12u2
OpenSSH_9.2p1 Debian-2+deb12u2, OpenSSL 3.0.13 30 Jan 2024
debug1: Reading configuration data /home/memet/.ssh/config
debug1: /home/memet/.ssh/config line 2: Deprecated option "useroaming"
debug1: /home/memet/.ssh/config line 4: Applying options for fake-jump
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 19: Applying options for *
debug1: Connecting to 192.168.50.165 [192.168.50.165] port 2222.
debug1: connect to address 192.168.50.165 port 2222: Connection refused
ssh: connect to host 192.168.50.165 port 2222: Connection refused
kex_exchange_identification: Connection closed by remote host
Connection closed by UNKNOWN port 65535
```

I can connect to `bogey-2` via -p override, I can connect to `fake-jump` via -p override, but when I do -p override on `bogey-1`, it does *NOT* override the port on `fake-jump`. You can see the connection fails on the first hop.


> Second example does not work because IMO it first tries to port 2222 on fake-jump which IMO is fine - I do not know of any way to override this from commandline.

Yes, this is what the man page says will happen as the exception, not the rule. Specifically for JumpHost.

> Now when I read your description in PR it is more clear to me what you meant. I must say I was lost a bit - this SSH handling part grew significantly.

It didn't "grow" so much as it moved beyond trivial handling. The last release had an [outdated sshconfig package (v0.0.0)](https://github.com/dmacvicar/terraform-provider-libvirt/blob/v0.7.6/go.mod#L10) and it *only* used the ["User"](https://github.com/dmacvicar/terraform-provider-libvirt/blob/v0.7.6/libvirt/uri/ssh.go#L123) parameter.

Besides, without sshconfig, bastion hosts were not possible and this is a big no-go for my (any?) corporate environment.

All we need to finalize is how to be backward compatible while maintaining good hygiene moving forward...

I just realized that there is no backward compatibility issue at all. v0.7.6 simply did not support proxy jumps, therefore nobody has production code with proxy jumps in them that can break due to this change.

mhtr:
@memetb hi
I tested your fix, it seems that with the setting of port from the ssh config everything is ok, but there is a problem like in [1108](https://github.com/dmacvicar/terraform-provider-libvirt/issues/1108)

mhtr:
@memetb hi
Thanks for your fix, it worked for me. But now I have another issue [1121](https://github.com/dmacvicar/terraform-provider-libvirt/issues/1121)

dmacvicar:
There is no need to be backward compatibility and go with the behavior closer to ssh/libvirt.

 We are not in 1.0 yet. Lets just mention it in  the release notes.

PR #1120, opened 2024-10-23
feat: do not fail if server returns 200 OK

### LINKED ISSUE #958: can't retrieve partial header
Hi,

I am trying to deploy some Rocky Linux 8.5 VMs. The first time I run it, I always get the error described below. Only when I run it again does the deployment work.

Volker

### Linux distribution

```
Arch Linux (Host)
Rocky Linux 8.5 (Target)
```

### Terraform version

```sh
Terraform v1.2.3
on linux_amd64
+ provider registry.terraform.io/dmacvicar/libvirt v0.6.14
+ provider registry.terraform.io/hashicorp/template v2.2.0
```

```sh
0.6.14
```

## Checklist

- [x] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

### Setup

```tf
resource "libvirt_network" "gh_runner" {
  name      = var.network_name
  mode      = var.network_mode
  domain    = var.network_mode == "nat" ? var.network_domain_name : ""
  addresses = var.network_mode == "nat" ? var.network_addresses : []
  bridge    = var.network_mode == "bridge" ? var.network_bridge : ""
  autostart = true

  # (optional) the MTU for the network. If not supplied, the underlying device's
  # default is used (usually 1500)
  # mtu = 9000

  # (Optional) DNS configuration
  dns {
    # (Optional, default false)
    # Set to true, if no other option is specified and you still want to
    # enable dns.
    enabled = true

    # (Optional, default false)
    # true: DNS requests under this domain will only be resolved by the
    # virtual network's own DNS server
    # false: Unresolved requests will be forwarded to the host's
    # upstream DNS server if the virtual network's DNS server does not
    # have an answer.
    local_only = true

    # (Optional) one or more DNS forwarder entries. One or both of
    # "address" and "domain" must be specified. The format is:
    # forwarders {
    #     address = "my address"
    #     domain = "my domain"
    #  }
    #

    # (Optional) one or more DNS host entries. Both of
    # "ip" and "hostname" must be specified. The format is:
    hosts {
      hostname = "router"
      ip = "192.168.175.1"
    }

    # (Optional) one or more DNS SRV entries. Both of
    # "service" and "protocol" must be specified. The format is:
    # srvs {
    #   service = "my-service"
    #   protocol = "my-protocol"
    # }


    # (Optional) one or more static routes.
    # "cidr" and "gateway" must be specified. The format is:
    # routes {
    #     cidr = "10.17.0.0/16"
    #     gateway = "10.18.0.2"
    #   }
  }

  # (Optional) Dnsmasq options configuration
  dnsmasq_options {
    # (Optional) one or more option entries.  Both of
    # "option_name" and "option_value" must be specified.  The format is:
    # options  {
    #     option_name = "server"
    #     option_value = "/base.domain/my.ip.address.1"
    #   }
    # options {
    #     option_name = "address"
    #     ip = "/.api.base.domain/my.ip.address.2"
    #   }
    #
  }
}

resource "libvirt_pool" "ghrunner" {
  name = var.storage_pool_name
  type = "dir"
  path = var.storage_pool_path
}

resource "libvirt_volume" "rocky" {
  name    = "rocky_linux.qcow2"
  source  = "https://download.rockylinux.org/pub/rocky/8/images/Rocky-8-GenericCloud.latest.x86_64.qcow2"
  format  = "qcow2"
  pool    = libvirt_pool.ghrunner.name
}

resource "libvirt_volume" "runner" {
  name            = "ghrunner_${count.index}.qcow2"
  base_volume_id  = libvirt_volume.rocky.id
  count           = var.runner_count
  pool            = libvirt_pool.ghrunner.name
  size            = var.runner_disk_size
  format          = "qcow2"
}

data "template_file" "ci_meta" {
  template = file("${path.module}/cloud_init_-_meta_data.cfg")
}

data "template_file" "ci_network" {
  template = file("${path.module}/cloud_init_-_network_data.cfg")
}

data "template_file" "ci_user" {
  template = file("${path.module}/cloud_init_-_user_data.cfg")
}

resource "libvirt_cloudinit_disk" "cloud_init" {
  name            = "cloud_init.iso"
  network_config  = data.template_file.ci_network.rendered
  meta_data       = data.template_file.ci_meta.rendered
  user_data       = data.template_file.ci_user.rendered
  pool            = libvirt_pool.ghrunner.name
}

# # Create the machine
resource "libvirt_domain" "rocky" {
  name   = format("%s-${count.index}", var.runner_name_prefix)
  memory = var.runner_memory
  vcpu   = var.runner_vcpus

  count = var.runner_count

  cloudinit = libvirt_cloudinit_disk.cloud_init.id

  network_interface {
    network_name    = var.network_name
    hostname        = format("%s-${count.index}", var.runner_name_prefix)
    wait_for_lease  = true
  }

  disk {
    volume_id = libvirt_volume.runner[count.index].id
  }

  graphics {
    type        = "spice"
    listen_type = "address"
    autoport    = true
  }
}
```

variables

```tf
#
# Libvirt: qemu related variables
#
variable "qemu_uri" {
  description = "URI to connect with the qemu-service."
  type        = string
  default     = "qemu:///system"
}

# Libvirt: network related variables
#
variable "network_addresses" {
  description = "List of none or one IPv4 and one or none IPv6 Subnet."
  type        = list(string)
  default     = [ "192.168.175.0/24" ]
}

variable "network_bridge" {
  description = "Name of the network bridge to bind VMs to."
  type        = string
  # FIXME: Remove default value and prompt only for bridge interface when network mode bridge is selected.
  default     = "br0"

  validation {
    condition     = length(var.network_bridge) >= 1
    error_message = "The network_bridge must contains at least one char."
  }
}

variable "network_domain_name" {
  description = "DNS domain name for VM's."
  type        = string
  default     = "ghrunner.local"
}

variable "network_mode" {
  description = "Network mode of the virtual network. Possible values are: \"nat\" (default), \"bridge\"."
  type        = string
  default     = "nat"

  validation {
    condition     = contains(["bridge", "nat"], var.network_mode)
    error_message = "Invalid virtual network mode. Only \"nat\" and \"bridge\" is supported."
  }
}

variable "network_name" {
  description = "Name of the virtual network displayed in libvirt."
  type        = string
  default     = "ghrunner"

  validation {
    condition     = length(var.network_name) >= 1
    error_message = "The network_name must contains at least one char."
  }
}

#
# Libvirt: storage related variables
#
variable "storage_pool_name" {
  description = "Name of the storage pool for the runner VM disks"
  type        = string
  default     = "ghrunner"

  validation {
    condition     = length(var.storage_pool_name) >= 1
    error_message = "The storage_pool_name must contains at least one char."
  }
}

variable "storage_pool_path" {
  description = "Path of the storage pool for the runner VM disks"
  type        = string
  default     = "/var/lib/libvirt/pool/ghrunner"
}

#
# Libvirt: vm related variables
#
variable "runner_name_prefix" {
  description = "Name prefix of the VM's"
  type        = string
  default     = "ghrunner"

  validation {
    condition     = length(var.runner_name_prefix) >= 1
    error_message = "The runner_name_prefix must contains at least one char."
  }
}

variable "runner_count" {
  description = "Amount of runner VMs"
  type        = number
  default     = 5

  validation {
    condition     = var.runner_count >= 1
    error_message = "The runner_count must be greather or equal than 1."
  }
}

variable "runner_disk_size" {
  description = "Size of the virtual hard disk"
  type        = string
  default     = "53687091200"

  validation {
    condition     = var.runner_disk_size >= 53687091200
    error_message = "The runner_disk_size must be greather or equal than 53687091200 bytes."
  }
}

variable "runner_memory" {
  description = "Memory assigned to one VM"
  type        = number
  default     = 4096

  validation {
    condition     = var.runner_memory >= 1
    error_message = "The runner_memory must be greather or equal than 1."
  }
}

variable "runner_vcpus" {
  description = "Amount of vCPU's for one VM"
  type        = number
  default     = 2

  validation {
    condition     = var.runner_vcpus >= 1
    error_message = "The runner_vpu must be greather or equal than 1."
  }
}
```
### Steps to Reproduce Issue

```sh
terraform apply
...
libvirt_pool.ghrunner: Creating...
libvirt_network.gh_runner: Creating...
libvirt_pool.ghrunner: Creation complete after 5s [id=f020231e-c8a2-4747-b5f7-78a2c9b8c49b]
libvirt_cloudinit_disk.cloud_init: Creating...
libvirt_volume.rocky: Creating...
libvirt_network.gh_runner: Creation complete after 5s [id=7eacddf8-2ded-4197-9c2e-47adf7ab220c]
libvirt_cloudinit_disk.cloud_init: Creation complete after 1s [id=/var/lib/libvirt/pool/ghrunner/cloud_init.iso;ba7aabbf-3365-408b-90ea-608467394fc7]
╷
│ Error: error while determining image type for https://download.rockylinux.org/pub/rocky/8/images/Rocky-8-GenericCloud.latest.x86_64.qcow2: can't retrieve partial header of resource to determine file type: https://download.rockylinux.org/pub/rocky/8/images/Rocky-8-GenericCloud.latest.x86_64.qcow2 - 200 OK
│ 
│   with libvirt_volume.rocky,
│   on main.tf line 81, in resource "libvirt_volume" "rocky":
│   81: resource "libvirt_volume" "rocky" {
│ 
╵
```
## Additional information:

```sh 
SELinux: no
Apparmor/Firewall: no
```

scabala:
Another rather easy fix.

Closes #958 

FYI @dmacvicar @volker-raschek @hansingt

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced image processing reliability by refining the validation and error handling of image data. These improvements ensure that images are loaded with complete and accurate data, resulting in a more robust and smoother experience for users.

### DIFF
--- a/libvirt/volume_image.go
+++ b/libvirt/volume_image.go
@@ -137,7 +137,7 @@ func (i *httpImage) IsQCOW2() (bool, error) {
 	}
 	defer response.Body.Close()
 
-	if response.StatusCode != http.StatusPartialContent {
+	if response.StatusCode != http.StatusPartialContent && response.StatusCode != http.StatusOK {
 		return false, fmt.Errorf(
 			"can't retrieve partial header of resource to determine file type: %s - %s",
 			i.url.String(),
--- a/libvirt/volume_image.go
+++ b/libvirt/volume_image.go
@@ -144,16 +144,17 @@ func (i *httpImage) IsQCOW2() (bool, error) {
 			response.Status)
 	}
 
-	header, err := io.ReadAll(response.Body)
+	header := make([]byte, 8)
+	n, err := io.ReadFull(response.Body, header)
 	if err != nil {
 		return false, err
 	}
 
-	if len(header) < 8 {
+	if n < 8 {
 		return false, fmt.Errorf(
 			"can't retrieve read header of resource to determine file type: %s - %d bytes read",
 			i.url.String(),
-			len(header))
+			n)
 	}
 
 	return isQCOW2Header(header)


hansingt:
Thanks for the fix @scabala!

Some thoughts I have about this change and the code here in general:

If we accept HTTP 200 responses here, the response might contain the whole image. This depends on whether the `http` go module supports streaming or not. But I'd guess it does.

If it does, only the first chunks will be received until the network buffer is filled. Thus, this is not that much of waste of network traffic, but should be considered. Maybe, it should be considered / checked, whether the `IsQCOW2` check could be merged with the `Import` method, to download things only once? But this might be a separate issue.

Anyway, right now, the `io.ReadAll()` in [Line 147](https://github.com/dmacvicar/terraform-provider-libvirt/blob/26e70c9432765ed44d28e9db34819ddc07eba05c/libvirt/volume_image.go#L147) causes the whole image to be downloaded. This does not only block until the download is complete, it does also waste (possibly several gigabytes of) memory, or might even fail, if the computer runs out of memory. This depends on the size of the image to be downloaded.

We already know, that we only need the first 8 bytes of the file, thus, it would be better to only read that much. Maybe 
 [`io.ReadFull`](https://pkg.go.dev/io#ReadFull) would be correct here? 

Last, but not least, actually, all HTTP Codes in the 200-299 range are to be considered as successful responses, but that doesn't really matter here, because the other status codes defined in this range arn't important in this case.

Besides this, this change looks good for me! :+1:

dmacvicar:
I understand the ReadFull is harmless if the server answered with `StatusPartialContent`.

I don't get why we should support servers that do not accept ranges? What problem are we trying to solve?

hansingt:
> I understand the ReadFull is harmless if the server answered with `StatusPartialContent`.
> 
> I don't get why we should support servers that do not accept ranges? What problem are we trying to solve?

Hi @dmacvicar,

Please see https://github.com/dmacvicar/terraform-provider-libvirt/issues/958#issuecomment-2415946592:

There are HTTP-Servers, which do not support Range-Requests, like the [flatcar](https://www.flatcar.org) download servers. And according to the [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#range.requests), they don't have to, as it is an optional feature.

In this case, the server is responding with a normal HTTP 200 response whixh contains the whole content.

Thus, to support these servers, it would be better to explicitly only read the first bytes required Form the request, to avoid loading the whole image in memory.

jonkerj:
@dmacvicar did you get a chance to look into this? This bug/feature is blocking IAC-ing Talos deployments using Terraform in a clean way, and it seems Talos servers are behaving according to RFC 9110.

dmacvicar:
If we merge, and the server does not support ranges, we would potentially download a big image twice correct? (one for the format check, one for the actual download).

hansingt:
> If we merge, and the server does not support ranges, we would potentially download a big image twice correct? (one for the format check, one for the actual download).

Potentially yes. But, as I,ve written above, this largely depends on whether wie can utilize a streaming API or not. Normally HTTP Clients use a buffer for receiving contents. If this buffer is filled, it has to be read first before new contents can be downloaded from the Server. Thus, as only the first few bytes are required, the buffer will be filled once during format check, we can read the bytes required and then discard the request without reading further content.

On the other hand: Is the format check required to be done before downloading the image? As I understand, this is just an optimization to avoid unnecessary download of unsupported contents, right? 

If the server does not support ranges and reading only the first few bytes without downloading the whole image is not possible, we could simply delay the format check to after downloading the image.

dmacvicar:
Looking at the code (https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/resource_libvirt_volume.go#L118), I see we only use IsQCOW2() in one place.

We first get the `format`, if given, and then we unconditionally detect the format. If `format` was given, we complain if these don't match. If no `format` is given, we use the auto-detected one.

I think a better fix would be:

- Don't unconditionally auto-detect format, but only do it if `format` is not provide. Do require HTTP Range support for autodetection and document it.
- If `format` is given, don't auto-detect, and if the format do not match what is downloaded, libvirt will error on its own later.

For your usecase, you would have then to provide the right format.

What do you think about this approach?

hansingt:
Why just not do the auto-detection after downloading the image? This way range support isn't required at all.

Yes, this might download something which cannot be used afterwards, but if the user passes a wrong download URL it is broken anyway.

Besides this, the suggested approach would be done for me. Just looks a little overcomplicated to me.

dmacvicar:
I think that would be better. But I am not sure if that would require a major rework of the code paths. Need to take a deeper look.

dmacvicar:
What you suggest would require:

- modify the created volume afterwards, as uploading data to the volume requires creating the volume first (where we already specify format)
- We would need to buffer the header, as we just pipe the source reader directly to the target pool writer

I have some code to do this already, however I am not a fan of the added complexity, as the feature to support remote images is already something we do as just convenience on top of libvirt.

At the moment I am more inclined to:

1. just set it based on the extension, at least for HTTP + only detect if no format is provided
2. completely removing content sniffing. Which also goes more in line with: https://libvirt.org/kbase/secureusage.html#disk-image-format-probing

Will start with 1. for now, and may be go 2. later.

dmacvicar:
Obsoleted by 2de95d1f3fd7efc788ab2d94e2e286e9baa02694

Issue #1017, opened 2023-05-17
when expanding the disk, the disk is recreated


sergey-ostrovskiy:
Hello

terraform version v1.4.5
provider version v0.7.1

I want to increase the disk, but at the same time the disk is recreated and the data from it is deleted.

Initially I use an empty disk and install the OS through pxe

Change disk size in resource libvirt_volume

what am I doing wrong?

Magnitus-:
You are not doing anything wrong. The resource is configured to be recreated if the volume size change: https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/resource_libvirt_volume.go#L40

You have some options open to you to circumvent this:
- If you have an HA setup (etcd cluster, postgres with patroni, opensearch cluster with all indexes having replica factor greater than 1, etc), you can recreate your vms one by one and let them resynchronize with barely noticeable effects on the system (though if you can restore the new vms from a backup so that it is not too far behind, the synchronization will be less resource intensive). I'd backup the data just in case before doing this though (in theory, HA solutions should support this gracefully, in practice, software can have bugs)
- If you don't have an HA setup, then you will need to:
   - Ensure you have an up to date backup
   - Bring destroy your vm and incur some downtime
   - Re-create the vm and restore from your backup
   
Alternatively, you can:
- Change your code with the new disk size
- Change the terraform state to resize the vm's disk
- Manually change the disk size
- Go into your vm to make the necessary adjustments to utilize the increased disk size (ie, increase partition size)
- Either way, I'd backup your vm before doing this in case something goes wrong. This would be a rather error-prone process. Also, you'll have to shutdown your vm temporarily and incur some downtime anyways.

But the last solution would definitely be working against the grain of what terraform and this provider does.

scabala:
Thanks @Magnitus- for detailed workaround.
I think it would be nice feature to allow disk size change but only in one direction - to allow increase. Decreasing would be destructive change nonetheless.

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

This issue report is relevant to the legacy version of the provider, which is now in the v0.8 branch of this repository.

Future development is based on a new provider, which is not compatible with this one, nor does share code.

As the new provider solves many issues with the legacy, and to make development in my free time more enjoyable, I have decided to close all bugs for the legacy provider, and to ask to check if the bug would apply to the new one. This also to encourage trying the new version.

[...]

Issue #952, opened 2022-05-23
[Feature Request] Non destructive volume resize


Hybrid512:
### Linux distribution

``` Ubuntu 20.04/22.04 ```

### Terraform version

```sh
Terraform v1.0.11 on linux_amd64
```

### Provider and libvirt versions

```sh
terraform-provider-libvirt 0.6.14
```
___

## Checklist

- [x] Is your issue/contribution related with enabling some setting/option exposed by libvirt that the plugin does not yet support, or requires changing/extending the provider terraform schema?
  - [ ] Make sure you explain why this option is important to you, why it should be important to everyone. Describe your use-case with detail and provide examples where possible.
  - [ ] If it is a very special case, consider using the _XSLT_ support in the provider to tweak the definition instead of opening an issue
  - [ ] Maintainers do not have expertise in every libvirt setting, so please, describe the feature and how it is used. Link to the appropriate documentation

- [ ] Is it a bug or something that does not work as expected? Please make sure you fill the version information below:

## Description of Issue/Question

Terraform can detect state differences between an existing volume size and an expected new volume size but with the current libvirt provider, volumes are first deleted then re-created with the new size.
This is ok for reproductible and immutable environments but this is annying especially when you have big volumes or data you would like to avoid losing or restoring (repository mirroring, datasets, ...).
Libvirt enable resizing a volume without destrying/re-creating it, I can do it easily with virsh just like this :

```sh
virsh vol-resize <MyVolume> <NewSizeInBytes> --pool <MyVolumePool>
```

Only conditions for that : 
* VM must be stopped
* new volume size can't be smaller than original size (can only expand, not shrink)

### Steps to Reproduce Issue

Any VM with at least a volume

jtackaberry:
It's actually possible to do an online resize with libvirt:

```
virsh blockresize somevm --path /vms/somevm-root.qcow2 --size 10G
```

And this is detected immediately within the guest:

```
[  267.339414] virtio_blk virtio2: [vda] new size: 20971520 512-byte logical blocks (10.7 GB/10.0 GiB)
[  267.339423] vda: detected capacity change from 12582912 to 20971520
```

I'd love to see this possible via terraform-provider-libvirt too.

JSmith-Aura:
I'd love to see this as well, would make life a lot more fun (also thanks for all the work you do on this project)

merlinfvr:
[...]

PR #1122, opened 2024-10-24
Feat: Allow increasing a storage volume size

NamelessOne91:
Related to https://github.com/dmacvicar/terraform-provider-libvirt/issues/952 and https://github.com/dmacvicar/terraform-provider-libvirt/issues/1017

This PR adds support for a non-destructive resize when a storage volume size is increased.
Shrinking a storage volume will still ultimately result in destroying the existing one and recreating it.

It would have been possible to handle both cases, shrinking and expanding, in the `UpdateContext `function, but having a `CustomizeDiff` function allows to correctly differentiate between a resource update  and a destroy + add, notifying it to the user.


scabala:
Thank you for this PR @NamelessOne91!

I'll try to test it in following days and give my feedback.

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #1123, opened 2024-10-27
Allow resizing the source image for a volume

wfdewith:
I've put a restriction of only resizing when the desired size is larger than the original size. Even though libvirt allows shrinking (but only when VIR_STORAGE_VOL_RESIZE_SHRINK is specified), this is not an operation that is readily supported by most VM images.


wfdewith:
I see that #1122 touches something related to this change. If both are merged, we probably need to update some stuff to reuse the same code paths for resizing images.

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

Issue #1124, opened 2024-10-29
Specifying a port in the connection string incorrectly adds the port to the name of the host


memetb:
This issue relates to #1116.

### Linux distribution

``` Debian 6.1.94-1 (2024-06-21) x86_64 GNU/Linux ```

### Terraform version

```
Terraform v1.5.7
on linux_amd64
```

### Provider and libvirt versions

```sh
v0.8.1
```

___

### Setup

As per the comment [here](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/uri/connection_uri.go#L30-L47), the name passed to the virConnectOpen should be stripped of port names.

```
// According to https://libvirt.org/uri.html
// The name passed to the remote virConnectOpen function is formed by removing
// transport, hostname, port number, username and extra parameters from the remote URI
// unless the name option is specified.
```

However, it seems specifying a port name does not correctly do this. This bug is subtle, however adding a test scenario reveals its presence. It also affects ssh-config based settings (specifically #1116) because the server name becomes incorrectly parsed as the string "servername:port" instead of "servername", causing the ssh_config to fail loading specific settings.

This code affects [this](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/config.go#L36) call. However, I do not know how it affects the operation of virConnectOpen itself.

```
	if err := l.ConnectToURI(libvirt.ConnectURI(u.RemoteName())); err != nil {
```

PR #1125, opened 2024-10-29
Connection port parsing

memetb:
This issue is meant to address #1124 

This is a subtle bug because it has practically no impact on the libvirt code base. It only affects libvirt [here](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/config.go#L36):

```golang
	if err := l.ConnectToURI(libvirt.ConnectURI(u.RemoteName())); err != nil {
```

Outside of this location, this bug only affects ssh_config settings by incorrectly passing the target as "hostname:portname" instead of "hostname" [here](https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/libvirt/uri/ssh.go#L135):

```

	// configuration loaded, build tunnel
	sshClient, err := u.dialHost(u.Host, sshcfg, 0)
	if err != nil {
		return nil, err
	}
```

However, when a port override is specified in the ConnectionString, u.Host will contain the hostname.

### DIFF
--- a/libvirt/uri/connection_uri_test.go
+++ b/libvirt/uri/connection_uri_test.go
@@ -12,15 +12,17 @@ func TestURI(t *testing.T) {
 		Driver     string
 		Transport  string
 		RemoteName string
+		HostName   string
+		Port       string
 	}{
-		{"xxx://servername/", "xxx", "tls", "xxx:///"},
-		{"xxx+tls://servername/", "xxx", "tls", "xxx:///"},
-		{"xxx+tls:///", "xxx", "tls", "xxx:///"},
-		{"xxx+tcp://servername/", "xxx", "tcp", "xxx:///"},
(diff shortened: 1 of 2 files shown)


memetb:
The added unittest assertions in this PR reveal the problem here:

```
 === RUN   TestURI
    connection_uri_test.go:34: 
        	Error Trace:	/home/runner/work/terraform-provider-libvirt/terraform-provider-libvirt/libvirt/uri/connection_uri_test.go:34
        	Error:      	Not equal: 
        	            	expected: "hostname"
        	            	actual  : "hostname:2222"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-hostname
        	            	+hostname:2222
        	Test:       	TestURI
```

Issue #1041, opened 2023-10-22
Add support for managing the domain title property


rgl:
Please add support for managing the domain `title` property.

Example [libvirt domain xml](https://libvirt.org/formatdomain.html) `title` property set to `xxx`:

```xml
<domain type="kvm">
  <name>example1-300594f</name>
  <uuid>7e9e1fe8-5c8b-4a87-8504-98b91079c7e4</uuid>
  <title>xxx</title>
  <description>managed by /home/vagrant/Projects/pulumi-typescript-libvirt-debian</description>
...
```

The [libvirt domain xml](https://libvirt.org/formatdomain.html) documents the property as:

> The optional element title provides space for a short description of the domain. The title should not contain any newlines. Since 0.9.10 .

Sathylias:
I second that, I think it's very handy when you have a lot of servers and the naming convention established sometimes just doesn't cut it.

scabala:
Hi,
thanks for the report. This is nice enhancement.

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

This issue report is relevant to the legacy version of the provider, which is now in the v0.8 branch of this repository.

Future development is based on a new provider, which is not compatible with this one, nor does share code.

As the new provider solves many issues with the legacy, and to make development in my free time more enjoyable, I have decided to close all bugs for the legacy provider, and to ask to check if the bug would apply to the new one. This also to encourage trying the new version.

- https://registry.terraform.io/providers/dmacvicar/libvirt/latest

and check the documentation:

- https://github.com/dmacvicar/terraform-provider-libvirt/blob/main/README.md
- https://registry.terraform.io/providers/dmacvicar/libvirt/latest/docs

You are free to reopen the bug for v0.8. It may encourage someone to fix it. My efforts will go into the new provider.

I ask you to check the new provider and re-evaluate this bug. 🙏

PR #1126, opened 2024-10-30
Enhancement: Add domain title property

NamelessOne91:
Related to https://github.com/dmacvicar/terraform-provider-libvirt/issues/1041

Adds basic support for managing a libvirt domain title property.

Possible further improvements:

- [ ]  have a `CustomDiff` function that checks if a new title is valid (read: contains no newline characters) BEFORE  attempting to replace a domain. Right now **when the new title contains newlines** the user can still attempt a replacement, the previous domain is destroyed and the new one fails to be created because the title is invalid.
No bueno.

- [ ] do not force a replacement when the title changes and correctly update it in place. I think libvirt API has no function to update only the title property, so this may require directly updating the domain XML configuration and redefining the domain.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -48,6 +48,11 @@ func resourceLibvirtDomain() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"title": {
+				Type:     schema.TypeString,
+				Optional: true,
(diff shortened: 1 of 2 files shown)


dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #1141, opened 2024-12-27
Make vcpu a block property

### LINKED ISSUE #1140: Enabling vcpu cpuset parameter for heterogeneous cpus
### Linux distribution

``` ARMtix ```

### Terraform version

```sh
terraform -v
```

Terraform v1.10.2
on linux_arm64

Provider: v0.8.1
Libvirt: 10.10.0
___

## Checklist

  - [x] If it is a very special case, consider using the _XSLT_ support in the provider to tweak the definition instead of opening an issue

An attempt to create libvirt domain on aarch64 device with heterogeneous cpu (4xCortex-A76+4xCortex-A55 cores) results into error message:
```
qemu-system-aarch64: Failed to put registers after init: Invalid argument
```
Running bare qemu-system-aarch64 with, e.g. cpu=host, works fine. Also running same configuration on the same system with only 4xCortex-A55 cores works fine too.
As shown in [this issue](https://github.com/Joshua-Riek/ubuntu-rockchip/issues/731#issuecomment-2127360776), in order to solve the problem one needs to either disable 4 cores on the host or force libvirt to only use subset of cpus via xml or cpuset arg (the latter is linked). The provider currently does not support cpuset argument for vcpus.
Unfortunately, i am no expert in libvirt myself, thus cannot give any underlying documentation.

### Setup

```
resource "libvirt_domain" "vm" {
  name      = "test"
  vcpu      = 1
  memory    = 412
  running   = true
  autostart = true

  arch     = "aarch64"
  machine  = "virt"
  type     = "kvm"
  firmware = "/usr/share/edk2/aarch64/QEMU_CODE.fd"

  cpu {
    mode = "host-passthrough"
  }
  disk {
    volume_id = "/armtix-worker/libvirt/pools/armtix/test_root"
    scsi      = true
    wwn       = "615392a40ddc3170"
  }

  console {
    type           = "tcp"
    target_port    = "7"
    target_type    = "serial"
    source_service = 20000
  }
}
```

### Steps to Reproduce Issue

The only thing needed to reproduce the issue is a heterogeneous cpu typically found on aarch64 devices.

___
## Additional information:

Disabled; the same setup works on the same system without heterogeneous cpus (4xCortex-A55)

phkrl:
This pull requests is intended to close #1140
libvirt_domain vcpu field is made into block, where count is old vcpu count and placement and cpusets are also configured. As a result, specifying the latter fields allow to run libvirt domains on aarch64 SoCs with different cores via this provider. Further testing is required, however, to check other platforms still work as intended.
Note: i cannot confirm that acceptance tests are working. On aarch64 they also fail for me in current main branch.
I'll update documentation to this PR once the schema and functionality are confirmed and approved by maintainers.

### DIFF
--- a/libvirt/resource_libvirt_domain.go
+++ b/libvirt/resource_libvirt_domain.go
@@ -525,10 +525,18 @@ func resourceLibvirtDomainCreate(ctx context.Context, d *schema.ResourceData, me
 		Value: uint(d.Get("memory").(int)),
 		Unit:  "MiB",
 	}
-	domainDef.VCPU = &libvirtxml.DomainVCPU{
-		CPUSet: d.Get("vcpu.0.cpuset").(string),
-		Placement: d.Get("vcpu.0.placement").(string),
-		Value: uint(d.Get("vcpu.0.count").(int)),
+	if vcpuCount, ok := d.GetOk("vcpu.0.count"); ok {
+		domainDef.VCPU = &libvirtxml.DomainVCPU{
+			CPUSet: d.Get("vcpu.0.cpuset").(string),
+			Placement: d.Get("vcpu.0.placement").(string),
+			Value: uint(vcpuCount.(int)),
+		}
+	} else {
+		domainDef.VCPU = &libvirtxml.DomainVCPU{
+			CPUSet: d.Get("vcpu.0.cpuset").(string),
+			Placement: d.Get("vcpu.0.placement").(string),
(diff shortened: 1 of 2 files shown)


dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #1143, opened 2025-01-10
fix: get ssh username from ~/.ssh/config, if not specified in uri

pnsafonov:
Fix allows to use

```
provider "libvirt" {
 uri ="qemu+ssh://mn0/system?no_verify=1&sshauth=privkey"
}
```

in addition to:
```
provider "libvirt" {
 uri = "qemu+ssh://p.safonov@mn0/system?no_verify=1&sshauth=privkey"
}
```
Then username is specified in ~/.ssh/config:
```
Host micronode0 mn0
    User p.safonov # read this string
    Port 22
    HostName 10.128.4.38
    IdentityFile /home/pavel/.ssh/server/pp/pp_mn0
    IdentitiesOnly yes
```

### DIFF
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -118,7 +118,7 @@ func (u *ConnectionURI) parseAuthMethods(target string, sshcfg *ssh_config.Confi
 // construct the whole ssh connection, which can consist of multiple hops if using proxy jumps,
 // the ssh configuration file is loaded once and passed along to each host connection.
 func (u *ConnectionURI) dialSSH() (net.Conn, error) {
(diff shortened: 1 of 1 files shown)


dmacvicar:
coderabbitai review

dmacvicar:
I'd like to deprecate the current ssh transport instead of adding support to more config options, in favour of the transport based on the native ssh command.

Is there a way you could help us testing those use cases will work in the new transport instead?

dmacvicar:
ℹ️ ℹ️ ℹ️ ℹ️ ℹ️ ℹ️

and check the documentation:

PR #1154, opened 2025-02-20
Add ability to set mode=private to a macvlan interface. 

farsonic:
As discussed in in open issues this is a pull-request to add mode=private to a MACVLAN interfaces, completing all current available options for this type of interface. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new "private" network interface mode that enables users to configure virtual machines with private networking options.

- **Documentation**
  - Updated documentation to explain the behavior of private networking, including how packets are routed through external bridges and routers under this mode.

### DIFF
--- a/libvirt/domain.go
+++ b/libvirt/domain.go
@@ -755,6 +755,13 @@ func setNetworkInterfaces(d *schema.ResourceData, domainDef *libvirtxml.Domain,
 					Mode: "vepa",
 				},
 			}
+		} else if devI, ok := d.GetOk(prefix + ".private"); ok {
+			netIface.Source = &libvirtxml.DomainInterfaceSource{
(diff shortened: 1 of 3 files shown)


dmacvicar:
@coderabbit review

dmacvicar:
Just for my own context in the future:

The `network_interface` block in our schema is flattened. This means `vepa`, `macvtap`, `passthrough` represent a device, which then gets converted to libvirt `source.direct.dev` in all cases, with a differerent `mode`.

I don't like this, but it is not something we will change in the domain schema (may be a v2).

Adding a `private` one just follows this approach and should be harmless.

dmacvicar:
Thanks for the contribution :pray:

PR #1172, opened 2025-04-06
fix known_hosts custom port

danwats:
If you're **not** using the default SSH port (22), and your known_hosts file only contains the default entry, terraform-provider-libvirt will fail with an error like:

```
Error: failed to connect: failed to connect to remote host 's1': ssh: handshake failed: knownhosts: key is unknown
```

with config entry like

```
Host s1
    HostName <hostname>
    Port 2222
```

### How to replicate:
1) ssh into the server under port 22 and accept known hosts entry
2) create custom ssh port on the server to 2222
3) change terraform/opentofu to use custom port or alias and it should error since known_hosts will be using default 22 port. SSH will not add in port 2222 since it uses port 22.

### Alternative Solution
The other alternative is just to add it manually using ssh-keygen e.g.
```ssh-keyscan -p 2222 <hostname>```

### DIFF
--- a/libvirt/uri/ssh.go
+++ b/libvirt/uri/ssh.go
@@ -118,7 +118,7 @@ func (u *ConnectionURI) parseAuthMethods(target string, sshcfg *ssh_config.Confi
 // construct the whole ssh connection, which can consist of multiple hops if using proxy jumps,
 // the ssh configuration file is loaded once and passed along to each host connection.
 func (u *ConnectionURI) dialSSH() (net.Conn, error) {
-	var sshcfg* ssh_config.Config = nil
+	var sshcfg *ssh_config.Config = nil
 
 	sshConfigFile, err := os.Open(os.ExpandEnv(defaultSSHConfigFile))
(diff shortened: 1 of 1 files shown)


guidot:
> [...] your known_hosts file only contains the default entry [...]

To clarify it should say: "your known_hosts file only contains an entry **that does not specify a port**"
Such an entry is valid for any port, not only the default (22).

Therefore I don't think the above mentioned alternative solution is valid. That would be just working around the bug.

danwats:
> Therefore I don't think the above mentioned alternative solution is valid. That would be just working around the bug.

you are correct, but this issue is bigger than that. I don't believe the knownhosts provided by golang doesn't actually do this, and this might be the actual fix. https://github.com/skeema/knownhosts/ which is already a note in the code.

I think the only work around is this solution

> The other alternative is just to add it manually using ssh-keygen e.g.
> ssh-keyscan -p 2222 <hostname>

I'll close this PR, as again, this is just skating around the issue.

PR #1183, opened 2025-07-11
fix: support domain name lookup when ID is not a UUID

rossigee:
Allow resourceLibvirtDomainRead to handle both UUID and domain name IDs. This fixes crashes when domains are imported by name instead of UUID.

When a domain is imported using terraform import with a domain name, the resource ID is set to the domain name. However, the read function was always trying to parse the ID as a UUID, which caused panics.

This change falls back to DomainLookupByName if UUID parsing fails and maintains backward compatibility.


rossigee:
@dmacvicar - any interest?

PR #1184, opened 2025-07-11
fix: implement libvirt TLS server verification

rossigee:
Add proper TLS handshake verification for libvirt connections over TLS.

When connecting to libvirt over TLS, the server writes a verification byte after the TLS handshake to indicate whether the client certificate and/or IP address verification succeeded. This was not being handled, which could lead to connection issues in some environments.

This change:
- Reads the server verification byte after TLS connection
- Properly handles verification failures with clear error messages  
- Improves robustness of TLS connections to libvirt
- Follows the libvirt TLS protocol as documented in go-libvirt issues

Also improves user detection fallback when /etc/passwd lookup fails by using os.Getuid() directly for container environments.

References: https://github.com/digitalocean/go-libvirt/issues/89#issuecomment-1607300636


rossigee:
@dmacvicar - any interest?

PR #1225, opened 2025-11-21
feat: Undefine on update toggle

atopuzov:
Exposes a provider level configuration to toggle undefining domains on update. The current default behavior is to undefine on update, but this causes issues with some setups (eg. efi vms with nvram).
This change makes it configurable, defaulting to the current behavior for backward compatibility.

Another option what we can do is also modify the Undefine behaviour (with `DomainUndefineFlags(existingDomain, flags)`) and select to undefine `nvram` explicitly (this removes nvram), preferably by exposing a config toggle. But IMHO then the user should just tell terraform to recreate the resource. 

An example change (lowering the amount of ram) where it fails.
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

  # libvirt_domain.builder-win will be updated in-place
  ~ resource "libvirt_domain" "builder-win" {
        id          = 56
      ~ memory      = 4096 -> 2096
        name        = "builder-win"
        # (10 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
libvirt_domain.builder-win: Modifying... [name=builder-win]
╷
│ Error: Domain Undefine Failed
│
│   with libvirt_domain.builder-win,
│   on builder-win.tf line 1, in resource "libvirt_domain" "builder-win":
│    1: resource "libvirt_domain" "builder-win" {
│
│ Failed to undefine existing domain: Requested operation is not valid: cannot undefine domain with nvram
```

With the patch and update to the provider config.
```
provider "libvirt" {
  uri = "qemu:///system"
  undefine_on_update = false
}
```
Log:
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

  # libvirt_domain.builder-win will be updated in-place
  ~ resource "libvirt_domain" "builder-win" {
        id          = 56
      ~ memory      = 4096 -> 2096
        name        = "builder-win"
        # (10 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
libvirt_domain.builder-win: Modifying... [name=builder-win]
libvirt_domain.builder-win: Modifications complete after 1s [name=builder-win]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

Verify the change:
```
virsh  dumpxml builder-win | grep currentMemory
  <currentMemory unit='KiB'>2146304</currentMemory>
```

### DIFF
--- a/internal/libvirt/client.go
+++ b/internal/libvirt/client.go
@@ -14,8 +14,9 @@ import (
 
 // Client wraps the libvirt connection and provides helper methods
 type Client struct {
-	conn *libvirt.Libvirt
-	uri  string
+	conn             *libvirt.Libvirt
+	uri              string
+	UndefineOnUpdate bool
 }
 
 // NewClient creates a new libvirt client from a connection URI
@@ -74,6 +75,10 @@ func (c *Client) Libvirt() *libvirt.Libvirt {
 // URI returns the connection URI
 func (c *Client) URI() string {
 	return c.uri
+}
+
+func (c *Client) IsUndefineOnUpdate() bool {
+	return c.UndefineOnUpdate
 }
 
 // Ping verifies the connection is still alive
--- a/internal/provider/domain_resource.go
+++ b/internal/provider/domain_resource.go
@@ -822,12 +822,14 @@ func (r *DomainResource) Update(ctx context.Context, req resource.UpdateRequest,
 		return
 	}
 
-	if err := r.client.Libvirt().DomainUndefine(existingDomain); err != nil {
-		resp.Diagnostics.AddError(
-			"Domain Undefine Failed",
-			"Failed to undefine existing domain: "+err.Error(),
-		)
-		return
+	if r.client.IsUndefineOnUpdate() {
+		if err := r.client.Libvirt().DomainUndefine(existingDomain); err != nil {
+			resp.Diagnostics.AddError(
+				"Domain Undefine Failed",
+				"Failed to undefine existing domain: "+err.Error(),
+			)
+			return
+		}
 	}
 
 	newDomain, err := r.client.Libvirt().DomainDefineXML(xmlString)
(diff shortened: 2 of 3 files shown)


dmacvicar:
Thanks for the PR.

The flag you describe for nvram is already in main (67f1b0290db016cfba1f72026a0a18eeb403ed29).

If we would allow flags to be passed for different parts of the lifecycle, we already have a block "create", so we would do a similar pattern. I think we can live without it for a while now that the flag is there.

atopuzov:
Hi @dmacvicar 

Thanks for the review.

I think there is a slight misunderstanding regarding the context of the error. I’ve looked at commit 67f1b02, and it appears to address the Delete lifecycle (handling NVRAM removal during terraform destroy).

The Issue:
My error occurs during the Update lifecycle (e.g., scaling RAM up or down on an existing VM). Currently, the provider's internal logic for resourceLibvirtDomainUpdate attempts to temporarily undefine the domain definition to apply changes. Because the VM has NVRAM (EFI), Libvirt rejects this internal undefine call, causing terraform apply to fail entirely for simple updates.

After digging deeper into the libvirt API constants, I found a better solution that addresses the issue without requiring any new configuration or user flags.

Instead of skipping the undefine step (my previous PR), I can modify the Update function to use domain.UndefineFlags(...) with VIR_DOMAIN_UNDEFINE_KEEP_NVRAM.

The issue happens because Undefine() defaults to deleting NVRAM for EFI VMs, which Libvirt blocks for safety. Passing KEEP_NVRAM allows the provider to temporarily remove the domain definition (to apply the update) while preserving the NVRAM file on disk. When the provider immediately re-defines the domain with the new RAM/CPU settings, it re-attaches to the existing NVRAM.

This aligns with the Update lifecycle (persisting state) and removes the need for the undefine_on_update toggle entirely.

 Can I push this change to the existing branch/PR for review, or would you prefer a fresh PR?

PR #1248, opened 2025-12-14
fix: return empty list instead of null when user explicitly sets []

### LINKED ISSUE #1247: Empty lists returned as null causing perpetual drift
When using `libvirt_domain` with an explicitly empty list for `hostdevs` (i.e., `hostdevs = []`), the provider returns `null` instead of `[]` after apply. This causes Terraform/OpenTofu to report "Provider produced inconsistent result after apply" errors.

This is a problem when you want to explicitly declare that a domain has no PCI passthrough devices, or when using dynamic configurations where some domains have hostdevs and others don't.

## Steps to Reproduce

1. Create the following minimal configuration:

```hcl
terraform {
  required_providers {
    libvirt = {
      source  = "dmacvicar/libvirt"
      version = "~> 0.9.0"
    }
  }
}

provider "libvirt" {
  uri = "qemu:///session"
}

resource "libvirt_volume" "test" {
  name     = "test-empty-list-bug.qcow2"
  pool     = "default"
  capacity = 1073741824

  target = {
    format = {
      type = "qcow2"
    }
  }
}

  name   = "test-empty-list-bug"
  memory = 512
  vcpu   = 1
  type   = "kvm"

  os = {
    type      = "hvm"
    type_arch = "x86_64"
  }

  devices = {
    disks = [
      {
        source = {
          volume = {
            pool   = "default"
            volume = libvirt_volume.test.name
          }
        }
        target = {
          dev = "vda"
          bus = "virtio"
        }
      }
    ]

    # Explicitly empty list - triggers the bug
    hostdevs = []
  }
}
```

2. Run `terraform init && terraform apply`

## Expected Behavior

Apply should succeed and `terraform plan` should show "No changes".

## Actual Behavior

Apply fails with:

```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to libvirt_domain.test, provider "provider[\"registry.opentofu.org/dmacvicar/libvirt\"]"
│ produced an unexpected new value: .devices.hostdevs: was cty.ListValEmpty(cty.Object(...)), but now null.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

BohdanTkachenko:
Previously, when a user set an empty list like `hostdevs = []`, the provider returned `null` instead of `[]` after apply, causing "Provider produced inconsistent result after apply" errors.

The fix adds a check: when the plan has an explicitly set (non-null) list but the XML has no elements, return an empty list to preserve user intent. This applies to both nested object lists and simple value lists.

- Import/datasource (no plan): still returns null for absent data
- User set `field = []`: now correctly returns `[]`
- User set `field = null` or omitted: returns null

Fixes #1247

### DIFF
--- a/internal/codegen/templates/convert.go.tmpl
+++ b/internal/codegen/templates/convert.go.tmpl
@@ -230,6 +230,13 @@ func {{ .Name }}ToXML(ctx context.Context, model *{{ .Name }}Model) (*libvirtxml
 				return nil, fmt.Errorf("converting {{ .GoName }} to List: %s", diags.Errors()[0].Summary())
 			}
 			model.{{ .GoName }} = listVal
+		} else if plan != nil && !plan.{{ .GoName }}.IsNull() {
+			// User explicitly set an empty list - return empty list to preserve intent
+			listVal, diags := types.ListValueFrom(ctx, types.ObjectType{AttrTypes: {{ .NestedStruct.Name }}AttributeTypes()}, []{{ .NestedStruct.Name }}Model{})
+			if diags.HasError() {
+				return nil, fmt.Errorf("converting {{ .GoName }} to List: %s", diags.Errors()[0].Summary())
+			}
+			model.{{ .GoName }} = listVal
 		} else {
 			model.{{ .GoName }} = types.ListNull(types.ObjectType{AttrTypes: {{ .NestedStruct.Name }}AttributeTypes()})
 		}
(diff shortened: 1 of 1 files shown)


dmacvicar:
Thanks for the PR. I fixed it in https://github.com/dmacvicar/terraform-provider-libvirt/commit/5b4cf1761bfaa4a73fae3b17035cb6a29da46084 with a proper testcase.

PR #1249, opened 2025-12-14
fix: allow integer fields with value 0 to be returned correctly

### LINKED ISSUE #1246: PCI address fields with value 0 returned as null causing perpetual drift
When using `libvirt_domain` with PCI passthrough devices (`hostdevs` with `subsys_pci`), any address field (domain, bus, slot, function) that has a value of `0` is incorrectly returned as `null` by the provider after apply. This causes Terraform/OpenTofu to report "Provider produced inconsistent result after apply" errors and makes it impossible to manage PCI passthrough devices where any address component is 0.

This is extremely common since most PCI devices have `domain=0`, and many have `slot=0` or `function=0` (e.g., `0000:03:00.0`).

## Steps to Reproduce

1. Create the following minimal configuration:

```hcl
terraform {
  required_providers {
    libvirt = {
      source  = "dmacvicar/libvirt"
      version = "~> 0.9.0"
    }
  }
}

provider "libvirt" {
  uri = "qemu:///system"
}

resource "libvirt_volume" "test" {
  name     = "test-pci-bug.qcow2"
  pool     = "default"
  capacity = 1073741824

  target = {
    format = {
      type = "qcow2"
    }
  }
}

  name   = "test-pci-bug"
  memory = 512
  vcpu   = 1
  type   = "kvm"

  os = {
    type      = "hvm"
    type_arch = "x86_64"
  }

  devices = {
    disks = [
      {
        source = {
          volume = {
            pool   = "default"
            volume = libvirt_volume.test.name
          }
        }
        target = {
          dev = "vda"
          bus = "virtio"
        }
      }
    ]

    # PCI passthrough device - update bus to match a real device on your system
    # The bug occurs when domain, slot, or function equals 0
    hostdevs = [
      {
        managed = true
        subsys_pci = {
          source = {
            address = {
              domain   = 0  # This will be returned as null
              bus      = 3  # Change to match your device
              slot     = 0  # This will be returned as null
              function = 0  # This will be returned as null
            }
          }
        }
      }
    ]
  }
}
```

2. Run `terraform init && terraform apply`

## Expected Behavior

Apply should succeed and `terraform plan` should show "No changes".

## Actual Behavior

Apply fails with:

```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to libvirt_domain.test, provider "provider[\"registry.opentofu.org/dmacvicar/libvirt\"]"
│ produced an unexpected new value: .devices.hostdevs[0].subsys_pci.source.address.function: was cty.NumberIntVal(0), but now null.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to libvirt_domain.test, provider "provider[\"registry.opentofu.org/dmacvicar/libvirt\"]"
│ produced an unexpected new value: .devices.hostdevs[0].subsys_pci.source.address.slot: was cty.NumberIntVal(0), but now null.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to libvirt_domain.test, provider "provider[\"registry.opentofu.org/dmacvicar/libvirt\"]"
│ produced an unexpected new value: .devices.hostdevs[0].subsys_pci.source.address.domain: was cty.NumberIntVal(0), but now null.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

BohdanTkachenko:
Previously, the codegen template incorrectly treated integer values of 0 as null/empty when converting XML to Terraform state. This caused PCI address fields (domain, bus, slot, function) with value 0 to be returned as null, resulting in "Provider produced inconsistent result after apply" errors.

Most PCI devices have domain=0, and many have slot=0 or function=0 (e.g., 0000:03:00.0), making PCI passthrough unusable without workarounds.

The fix removes the `!= 0` checks from convert.go.tmpl:
- For pointer fields: only check `!= nil`
- For non-pointer fields: always set the value

This affects 8 locations in the template covering computed, optional, and flattened value attribute fields.

Fixes #1246: perpetual drift for PCI passthrough devices where any address component equals.

### DIFF
--- a/internal/codegen/templates/convert.go.tmpl
+++ b/internal/codegen/templates/convert.go.tmpl
@@ -87,11 +87,7 @@ func {{ .Name }}ToXML(ctx context.Context, model *{{ .Name }}Model) (*libvirtxml
 				model.{{ $.GoName }}{{ .GoName }} = types.StringNull()
 			}
 			{{- else }}
-			if xml.{{ $.GoName }}.{{ .GoName }} != 0 {
-				model.{{ $.GoName }}{{ .GoName }} = types.Int64Value(int64(xml.{{ $.GoName }}.{{ .GoName }}))
-			} else {
-				model.{{ $.GoName }}{{ .GoName }} = types.Int64Null()
-			}
+			model.{{ $.GoName }}{{ .GoName }} = types.Int64Value(int64(xml.{{ $.GoName }}.{{ .GoName }}))
 			{{- end }}
 			{{- end }}
 		} else {
@@ -122,11 +118,7 @@ func {{ .Name }}ToXML(ctx context.Context, model *{{ .Name }}Model) (*libvirtxml
 			model.{{ $.GoName }}{{ .GoName }} = types.StringNull()
 		}
 		{{- else }}
-		if xml.{{ $.GoName }}.{{ .GoName }} != 0 {
-			model.{{ $.GoName }}{{ .GoName }} = types.Int64Value(int64(xml.{{ $.GoName }}.{{ .GoName }}))
-		} else {
-			model.{{ $.GoName }}{{ .GoName }} = types.Int64Null()
-		}
+		model.{{ $.GoName }}{{ .GoName }} = types.Int64Value(int64(xml.{{ $.GoName }}.{{ .GoName }}))
 		{{- end }}
 		{{- end }}
 		{{- end }}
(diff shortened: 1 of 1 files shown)


dmacvicar:
Thanks, I fixed it in [b26fb88](https://github.com/dmacvicar/terraform-provider-libvirt/commit/b26fb88b6389bd03b0af630420773631c42c8274), but only for the pointer case. It is not clear to me if we should set the zero values for optional non-pointer values.

PR #1261, opened 2026-01-07
Add example for dynamic disks and enable the qemu agent

cyberworm54:


cyberworm54:
Just adding some examples on how to enable the qemu-guest-agent as well as dynamic disks. I thought it would be useful for anyone out there looking for the same.

PR #1273, opened 2026-02-10
fix: domain.destroy.graceful/timeout values not working

### LINKED ISSUE #1060: Graceful shutdown of domain resource when recreate is required

nicholas-rees:
Fixes #1060

I was looking into why the domain.destroy.graceful option didn't seem to have any effect, and decided I'd play around with the code.

Didn't seem like a difficult fix so I just did it. After looking at your issues, it seems like other people were requesting this as well.

If there are any changes you would like please let me know. I think this is a very useful feature especially when these values are exposed in the schema of the provider.

### DIFF
--- a/internal/provider/domain_resource.go
+++ b/internal/provider/domain_resource.go
@@ -950,6 +950,14 @@ func (r *DomainResource) Delete(ctx context.Context, req resource.DeleteRequest,
 	if resp.Diagnostics.HasError() {
 		return
 	}
+	destroyState := DomainDestroyModel{
+    Graceful: types.BoolValue(false),
+    Timeout:  types.Int64Value(0),
+	}
+	resp.Diagnostics.Append(state.Destroy.As(ctx, &destroyState, basetypes.ObjectAsOptions{})...)
+	if destroyState.Timeout.ValueInt64() == 0 {
+		destroyState.Timeout = types.Int64Value(30)
+	}
 
 	// Look up the domain
 	domain, err := r.client.LookupDomainByUUID(state.UUID.ValueString())
@@ -970,13 +978,37 @@ func (r *DomainResource) Delete(ctx context.Context, req resource.DeleteRequest,
 
 	// DomainState values: 0=nostate, 1=running, 2=blocked, 3=paused, 4=shutdown, 5=shutoff, 6=crashed, 7=pmsuspended
 	if uint32(domainState) == uint32(golibvirt.DomainRunning) {
-		err = r.client.Libvirt().DomainDestroy(domain)
-		if err != nil {
-			resp.Diagnostics.AddError(
-				"Failed to Destroy Domain",
-				"Failed to stop running domain: "+err.Error(),
-			)
-			return
+		if destroyState.Graceful.ValueBool() {
+		  err = r.client.Libvirt().DomainShutdown(domain)
+			if err != nil {
+				resp.Diagnostics.AddError(
+					"Failed to Shutdown Domain",
+					"Failed to stop running domain: "+err.Error(),
+				)
+				return
+			}
+
+			err = waitForDomainState(
+				r.client,
+				domain,
+				uint32(golibvirt.DomainShutoff),
+				time.Duration(destroyState.Timeout.ValueInt64())*time.Second)
+			if err != nil {
+				resp.Diagnostics.AddError(
+					"Timeout waiting for Domain to shutdown: ",
+					"Timeout exceeded waiting for Domain to shutdown: "+domain.Name,
+				)
+				return
+			}
+		} else {
+			err = r.client.Libvirt().DomainDestroy(domain)
+			if err != nil {
+				resp.Diagnostics.AddError(
+					"Failed to Destroy Domain",
+					"Failed to stop running domain: "+err.Error(),
+				)
+				return
+			}
 		}
 	}
 


nicholas-rees:
I might mention, it looks like there is a linter error on main that is the same, but I'm not sure if I have broken some testing, but willing to fixing the testing if it is determined that I've broken something.

dmacvicar:
Thanks. This is a good catch.

The intention with these `create` and `destroy` blocks was to map them to the flags that `Create` and `Destroy`receive in libvirt. The intention was not to use `Shutdown`on the libvirt side but the [VIR_DOMAIN_DESTROY_GRACEFUL](https://libvirt.org/html/libvirt-libvirt-domain.html#VIR_DOMAIN_DESTROY_GRACEFUL) flag. From the docs:

>VIR_DOMAIN_DESTROY_GRACEFUL in the flags will prevent the forceful termination of the guest, and [virDomainDestroyFlags](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainDestroyFlags) will instead return an error if the guest doesn't terminate by the end of the timeout; at that time, the management application can decide if calling again without

So libvirt always try graceful first, but if after the timeout the domain has not exited, it will forcefully `SIGKILL`it. `VIR_DOMAIN_DESTROY_GRACEFUL` prevents this behavior, so Destroy will fail.

I think I missed the timeout is not configurable from the API, but I understand from `/etc/sysconfig/libvirt-guests` (`SHUTDOWN_TIMEOUT`) , therefore I think the right thing to do is to remove it from the schema.

dmacvicar:
This would be my proposal:

https://github.com/dmacvicar/terraform-provider-libvirt/pull/1280

nicholas-rees:
I think this makes more sense than what I was doing from an api standpoint. I was unaware of the graceful flag, I should have paid a bit more attention to the man pages. I just saw that the flags did not appear to be used at all, and made some assumptions.

However, I'm a tad bit worried about data corruption in the event that there are additional disks attached to a domain, but I'm at a loss how that would be done without extending the api in an inconsistent manner, so I think your proposal, unsurprisingly, makes more sense.

dmacvicar:
btw: I am not against adding a shutdown request before the destroy call. The problem is that I don't think timeout is a good way of making sure it worked, and I don't want to add too much logic on top.

An alternative would be to add to destroy a shutdown flag, and if true, send the shutdown call and wait until the domain is shutoff. No timeout configuration, but eventually if the domain does not shutoff, it would trigger the destroy operation timeout, preventing the domain to be destroyed then. I'd need to research this solution as it makes everything more clunky. It still maps 1:1 with libvirt functionality.

nicholas-rees:
I would be willing to contribute to this feature above mentioned if we could clearly define the issues. However, I'm not trying to dump a load of maintenance onto you.

Would you like me to open an issue about this? While I understand how this would still map 1:1 with libvirt functionality, I am not sure about whether the terafrom schema should map to functionality, or if it should instead map the the actual libvirt xml schema of each resource. i.e. I think deleting a domain from a .hcl file wouldbe like "undefining it" but how exactly do you "only" shutdown or "destroy" a vm? Just like flags like you're saying? It sounds like you have a specific way this would like to happen that is probably much more simple than I'm making it out to be. I guess for this, we're only concerned about whether or not the vm is only shut down or destroyed before being undefined, but I'm definitely not in the circle of understanding the goals of the project.

Would definitely like to help, but only if I could get on get a good understanding of how you want the api to match to libvirt. Either way, I would really like this as a feature to be sure that applications running in a vm are being safely shutdown to avoid potential problems.

dmacvicar:
@nicholas-rees I’m planning to evolve destroy behavior a bit while keeping your intent:

  1. Domain delete/undefine behavior remains unchanged: domains are destroyed and undefined during Terraform destroy.
  2. We’ll add an experimental destroy.shutdown mode:
      - if set, provider will call DomainShutdown and wait for shutoff up to a configured timeout.
      - if timeout is reached, Terraform destroy fails (no force fallback).
  3. destroy.graceful will remain available, but will also be marked experimental while we stabilize this area.

Example shutdown-with-timeout behavior (planned):

```
resource "libvirt_domain" "vm" {
    # ...
    destroy = {
      shutdown = {
        timeout = 30
      }
    }
  }
```

I want to keep this experimental for now because shutdown itself has many possible flags/methods (for example ACPI-related behavior), and I expect follow-up requests to make those configurable too. I'd prefer to keep shutdown behavior simple rather than adding more abstraction knobs.

I’d really appreciate if you can test this from main (or in the next release) once it lands, since we want feedback while this lifecycle behavior is still experimental.

nicholas-rees:
I will definitely be using this feature even if it is experimental. having to deal with api changes is better than not having the feature at all.

Thank you for supporting this feature. I will be on the lookout for the next release.

Issue #1242, opened 2025-12-07
vnc - unexpected new value


karamelichkin:
vm description: 
```
resource "libvirt_domain" "example" {
  provider = libvirt.kvm3
  
  name        = "example-vm"
  memory      = 1024
  memory_unit = "MiB"
  vcpu        = 1
  
  type = "kvm"

  os = {
    type         = "hvm"
    type_arch    = "x86_64"
    type_machine = "q35"
    boot = {
      devices = ["hd"]
    }
  }

  features = {
    acpi = true
  }

  devices = {
      graphics = [
    {
      type = "vnc"
      vnc = {
        autoport = true
        listen = "0.0.0.0"
        passwd = "password"
      }
    }
    ]

    disks = [
      {
        type = "volume"
        source = {
          volume = {
            pool   = libvirt_pool.st0_pool_kvm3.name
            volume = libvirt_volume.st0_volume_kvm3.name
          }
        }
        target = {
          dev = "vda"
          bus = "virtio"
        }
        driver = {
          name = "qemu"
          type = "raw"
        }
      }
    ]
    
    interfaces = [
      {
        type = "bridge"
        model = {
          type = "virtio"
        }
        source = {
          bridge = {
            bridge = "br0" 
          }
        }
        mac = {
          address = "52:54:00:12:34:56"
        }
      }
    ]

    consoles = [
      {
        type        = "pty"
        target_port = "0"
        target_type = "serial"
      }
    ]
  }
  running = true
}
```

vm runs successfully and vnc works fine with password i set

 but terraform output

```
libvirt_domain.example: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to libvirt_domain.example, provider "provider[\"registry.terraform.io/dmacvicar/libvirt\"].kvm3" produced an unexpected new value:
│ .devices.graphics[0].vnc.passwd: was cty.StringVal("password"), but now null.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

sampathP:
Hi @karamelichkin ,
Thanks for opening this issue.
I'm hitting the exact same error pattern right now with `libvirt_domain` network interfaces in provider version 0.9.1:
I get the following errors after terrafrom apply. Regardless of the error,
**VM works fine and both networks are connected properly.**
```
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to libvirt_domain.test_vm, provider
[...]

PR #1279, opened 2026-02-22
bug: add secure flag for GetXMLDesc

SkinGad:
When using private data, such as a VNC password, etc., it is recorded and then requested from the server, but due to the lack of the DomainXMLSecure flag (1), null is returned instead. Without the flag (0), passwords are actually set, but the provider returns an error when checking. Issue #1242

### DIFF
--- a/internal/provider/data_source_node_device_info.go
+++ b/internal/provider/data_source_node_device_info.go
@@ -210,7 +210,7 @@ func (d *NodeDeviceInfoDataSource) Read(ctx context.Context, req datasource.Read
 	deviceName := data.Name.ValueString()
 
 	// Get device XML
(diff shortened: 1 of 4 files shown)


dmacvicar:
Thanks for flagging this. :pray: 

I have pushed a fix in 0f47396bd058ece91ff0a1ef894f2cb8a563cfc0 which focuses on the domain case, and adds a test for it.

Your fix does not look correct to me. In addition to using the literal `1` instead of the `DomainXMLSecure` constant for domain, it sends 1 to other methods than domain e.g. `NetworkGetXMLDesc`, where 1 means `VIR_NETWORK_XML_INACTIVE`.

PR #1286, opened 2026-02-27
fix: Add internal/generated/imports.go for vendoring of generated sources

jeffmahoney:
The move to using generated schemas meant that 'go vendor' didn't see the imports in the generated code.  This commit adds an otherwise empty imports.go that contains the imports so offline builds can continue to work.  It's hidden behind a //go:build imports so there's no impact at build or runtime.

### DIFF
--- a/internal/generated/imports.go
+++ b/internal/generated/imports.go
@@ -1,0 +1,16 @@
+// go:build imports
+package generated
+
+// This is only here for vendoring
+
+import (
+	_ "github.com/hashicorp/terraform-plugin-framework/attr"
+	_ "github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	_ "github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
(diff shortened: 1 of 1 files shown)


dmacvicar:
Hi Jeff. I imagine this is in the context of packaging.

I don't like much the idea of having to maintain a file with imports only for this reason.

Have you thought avout vendoring internal/codegen, runing the generator and then vendoring again?

jeffmahoney:
I have but OBS doesn’t support automating that kind of workflow. I went this route because I got the impression that the list of imports wouldn’t change much over time and wouldn’t impose a meaningful maintenance burden. IMO, if it’s not maintained by the project, it won’t break regular builds. If changes are made that break OBS builds, whomever notices can submit a new PR.

dmacvicar:
@jeffmahoney can't you use OBS services for this? 

https://github.com/openSUSE/obs-service-go_modules
https://en.opensuse.org/openSUSE:Packaging_Go

My memory is fragile, but I remember they had a way to vendor dependencies and generate sources, just like packages maintained with git references create the sources ahead of the build.

PR #1331, opened 2026-05-10
Force network replacement on any attr change except autostart.

sgpinkus:
Currently network update only handles `autostart` correctly. A few attributes are apparently [live updatable](https://wiki.libvirt.org/Networking.html#applying-modifications-to-the-network) with `virsh net-update <network> <command> <section> <xml>` (or whatever the API equivalent of that is) but you have to supply a `<command> <section> <xml>` which would be tedious and messy AFAICT. So this change just marks everything except autostart as `RequiresReplace`.


dmacvicar:
Thanks for the contribution

Issue #1277, opened 2026-02-17
cannot import volumne


michalgrz:
I tried to migrate form 0.8.3 branch to 0.9.2 and I stuck on importing volumes. I narrowed the problem down to procedure of import resource in terraform. I created the following volume:

```
resource "libvirt_volume" "test_temp" {
  name = "test-format.qcow2"
  pool = "test"
  type     = "file"
  target = {
    format = {
      type="qcow2"
    }
  }
  capacity = 1024 * 1024 * 10 # 10MB
}
```
then I deleted the volume from terraform.tfstate and I tried to import it. I tried the following commands:

```
terraform import libvirt_volume.test_temp ad66d04b-c022-4be8-9218-7df7f43386f8/test-format.qcow2
terraform import libvirt_volume.test_temp test//media/kvm/test/test-format.qcow2
terraform import libvirt_volume.test_temp /media/kvm/test/test-format.qcow2
```

ad66d04b-c022-4be8-9218-7df7f43386f8 is id of pool.
None of these command succeed. There was allays the same error: "Storage volume not found: no storage vol with matching key".
When I manually edited terraform state and inserted volume:

   {
      "mode": "managed",
      "type": "libvirt_volume",
      "name": "test_temp",
      "provider": "provider[\"registry.terraform.io/dmacvicar/libvirt\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "id": "/media/kvm/test/test-format.qcow2",
            "key": "/media/kvm/test/test-format.qcow2",
            "name": "test-format.qcow2",
            "path": "/media/kvm/test/test-format.qcow2",
            "size": 10485760,
            "pool": "test",
            "target": {
              "format": {
                "type": "qcow2"
              },
              "path": "/media/kvm/test/test-format.qcow2"
            },
            "type": "file"
          },
          "dependencies": [
            "libvirt_pool.test"
          ]
        }
      ]
    }  

then terraform plan accepted the volume. Is there a better and easier way to import existing volume? during the import I skip some of the fields that terraform originally created for this volume.

mintos5:
I was struggling with this same problem, looks like a problem of importing of the 0.9 version. If I understood the code properly, when you are importing the volume user specify **only ID** of the object to import.  https://github.com/dmacvicar/terraform-provider-libvirt/blob/1c5c001600ba2b787970aff286f2fbfef7376060/internal/provider/volume_resource.go#L490-L492

[...]

PR #1334, opened 2026-05-18
fix: support importing volume by key

rst0070:
Hi, Thanks a lot for maintaining this project. 

I've been trying to import volume with this provider like `terraform libvirt_volume.name <key>`, and it kept failing. 
I could find same case #1277 here too. 

The reason was (`internal/provider/volume_resource.go`): `Read` looks the volume via `key` while `ImportState` doesn't populate it.

Happy to adjust the approach if you prefer other way. Thanks


dmacvicar:
Thanks for the contribution.

PR #1342, opened 2026-06-03
feat: make storage volume allocation user-overridable

bakhtin:
Previously allocation was a computed-only field with PreserveUserIntent disabled, so it could not be set from configuration and the unset value was always read back from libvirt. There was no way to influence whether a volume was sparsely or fully allocated.

Mirror capacity's treatment: keep allocation Computed but expose an Optional+Computed schema override and leave PreserveUserIntent enabled. When unset the volume is sparsely allocated exactly as before; when set, the configured value is authoritative and emitted to libvirt. Setting allocation equal to capacity yields a fully-allocated, non-sparse volume.

### DIFF
--- a/docs/resources/volume.md
+++ b/docs/resources/volume.md
@@ -98,6 +98,7 @@ resource "libvirt_volume" "from_local" {
 
 ### Optional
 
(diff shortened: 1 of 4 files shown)


bakhtin:
Fixed in https://github.com/dmacvicar/terraform-provider-libvirt/commit/1f59c15765f28aac63ebaa19ace9c364e47545af

PR #1361, opened 2026-07-07
fix(network): read persistent config for active networks to avoid inconsistent result

### LINKED ISSUE #1356: bug(network): "Provider produced inconsistent result" after apply on DNS host update for active networks
## Environment

- **Provider Version**: v0.9.7
- **OS**: Fedora 44 on x64
- **Virsh Version**: 12.0.0
- **Trigger**: adding a new DNS host entry to an existing, active `libvirt_network`

## Behaviour

1. When a new dns.host entry is added to an already-active `libvirt_network`, `terraform apply` completes the write but immediately reports:

    ```text
    Provider produced inconsistent result after apply
    libvirt_network.<name>: .dns.host: element N has vanished.
    ```

    The entry is correctly present in the persistent (on-disk) configuration that confirmed via `sudo virsh net-dumpxml --inactive <name>`, but absent in the active (in-memory) configuration at the time of read-back. Running `-parallelism=1` does not resolve the issue, which rules out a possibly race condition between concurrent applies.

2. The workaround is to manually sync the active configuration after apply:

    ```bash
    sudo virsh net-update <name> add dns-host --xml '<host ip="...">...</host>' --live
    terraform apply -refresh-only
    ```

### Root Cause as far as I can tell

1. `virNetworkDefineXML()` writes to persistent config only. For an active network, the active in-memory config is not updated until the network is restarted.
2. In `readNetwork()` in `internal/provider/network_resource.go:367`:

    ```go
    xmlDoc, err := r.client.Libvirt().NetworkGetXMLDesc(net, 0)
    ```

    `flags=0` retrieves the active config. Since `DefineXML` only updated the persistent config, the newly added `dns.host` entry is not visible in the read-back, causing Terraform to report the inconsistency. This function seems share across the Create, Update, and Read paths.

AruneshDwivedi:
When updating an active libvirt network (e.g., adding DNS host entries), the provider reads the active config (flags=0) but virNetworkDefineXML only updates the persistent config. This causes 'Provider produced inconsistent result' errors because the read-back doesn't see the changes.

Fix: Use flag 1 (VIR_NETWORK_XML_INACTIVE) to read the persistent configuration, which reflects the changes immediately after Apply.

Used AI to help identify and draft this fix — reviewed and tested before submitting.

Fixes #1356

### DIFF
--- a/internal/provider/network_resource.go
+++ b/internal/provider/network_resource.go
@@ -363,8 +363,11 @@ func (r *NetworkResource) ImportState(ctx context.Context, req resource.ImportSt
 
 // readNetwork reads network state from libvirt and populates the model
 func (r *NetworkResource) readNetwork(ctx context.Context, model *NetworkResourceModel, net golibvirt.Network, plan *generated.NetworkModel) error {
-	// Get network XML
-	xmlDoc, err := r.client.Libvirt().NetworkGetXMLDesc(net, 0)
+	// Get network XML - use flag 1 (VIR_NETWORK_XML_INACTIVE) to read the persistent
+	// configuration instead of the active configuration. This ensures that updates
+	// to active networks (like DNS host entries) are visible immediately after
+	// Apply, since virNetworkDefineXML only updates the persistent config.
+	xmlDoc, err := r.client.Libvirt().NetworkGetXMLDesc(net, 1)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


csning1998:
Changing flag to `VIR_NETWORK_XML_INACTIVE` masks the state inconsistency instead of resolving the issue, as Terraform will **no longer** reflect the actual active running config.

